PR #397, opened 2014-09-20
adding cordova support to apisocket.py

puneetgkaur:
Added a function named cordova to the ActivityAPI class of the apisocket.py. This would enable the messages from cordova to communicate with the native part lying in the cordova folder under sugar-toolkit.


SAMdroid-apps:
@sugarlabs/owners, @puneetgkaur, should we close this?

godiard:
Should be better have a consolidated patch

puneetgkaur:
Yes gonzalo, would do that

PR #421, opened 2014-12-08

i5o:
This is a possible fix of: SL#3672

[Fix some typos of https://github.com/sugarlabs/sugar/pull/420]


samdroid-apps:
@i5o, please make the `set_busy_cursor` method private as per @godiard's comments and then we can merge the patch!

PR #422, opened 2014-12-09
Volume up notifications

samdroid-apps:
**This is a feature and is for after the freeze.**

Show a little notification when you press the volume keys to change the volume.

### DIFF
--- a/src/jarabe/frame/notification.py
+++ b/src/jarabe/frame/notification.py
@@ -249,7 +249,7 @@ class NotificationIcon(Gtk.EventBox):
 
     _PULSE_TIMEOUT = 3
 
-    def __init__(self, **kwargs):
+    def __init__(self, pulse=True, **kwargs):
         self._icon = NotificationPulsingIcon()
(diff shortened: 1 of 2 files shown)


samdroid-apps:
Ok, this has way missed the freeze, but I was browsing through the designs on the wiki and came through some interesting stuff.


>  The new arrangement of icons within the Frame associates each edge with a particular corner of the screen. This goes hand in hand with the notification system, which takes advantage of the corners of the screen to provide meaningful yet unobtrusive alerts. These may occur, for instance, when a new invitation arrives, when people join or leave the current activity, when the battery is low, or when clippings are created.

Maybe that should inspire the design for this feature if it ever becomes a thing?

BTW, full wiki page:  http://wiki.sugarlabs.org/go/Design_Team/Designs/Frame

godiard:
Yes, something to the next cycle.

samdroid-apps:
We could peruse this for the 108 cycle, but I don't think it is very high priority.  I won't be perusing it myself and will therefore close the pull request.

PR #443, opened 2014-12-13
Adding control over the settings window from the keyboard.

cristian99garcia:
First I define keyval each key that I will use. When the window detects that a key is pressed, fixed than a movement key, escape or enter, if so, begin the movement actions, otherwise, continues the search operation that was previously.
To perform the movement, first saves the coordinates of each icon depressible in a dictionary to revise motion when a key is pressed. When a movement key is pressed, it first sets which icon is selected and makes the underlying transactions to make the next move.
When selecting a new icon, this is painted in the colors of the character XO, and you have to deselect icon is painted in black and white.
If the key is pressed enter, performs the same action as when you click on an icon (this does not work if the search does not match the selected icon).
When the button is pressed scape, should be in a section returns to the home screen, otherwise closes the window.

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -22,9 +22,11 @@ from gi.repository import GObject
 from gi.repository import Gtk
 from gi.repository import Gdk
 
+from sugar3 import profile
 from sugar3.graphics.icon import Icon
 from sugar3.graphics import style
 from sugar3.graphics.alert import Alert
+from sugar3.graphics.xocolor import XoColor
 
 from jarabe.model.session import get_session_manager
 from jarabe.controlpanel.toolbar import MainToolbar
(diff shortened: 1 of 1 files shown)


src/jarabe/controlpanel/gui.py
```diff
@@ -206,15 +283,36 @@ def _show_main_view(self):
...
-        if not entry.has_focus():
-            entry.grab_focus()
-        return False
+        if key not in [KEY_RIGHT, KEY_UP, KEY_LEFT, KEY_DOWN, KEY_INTRO,
+                       KEY_SCAPE]:
+            entry = self._main_toolbar.get_entry()
+            if not entry.has_focus():
+                entry.grab_focus()
+            return False
+
```
SAMdroid-apps: Why the empty line?

SAMdroid-apps: Usually we wouldn't have an empty line before an `else`

godiard:
In my opinion, if we want make the control panel accessible by keyboard, we should use a Gtk.IconView instead of a Gtk.Table to display the different icons. This patch show how complex is implement all the keyboard management by hand. This is a big change and probably would be better merge on 0.106

samdroid-apps:
This one will miss 0.106

samdroid-apps:
Well, unless somebody heroically patches it... but I will remove the milestone

samdroid-apps:
I will close this.  We need changes for this patch

PR #446, opened 2014-12-13
Don't save jabber without user confirmation.

i5o:
Right now the jabber server entry autosaves when the user presses enter.
With this patch, that no longer happens. The user must click the accept
button in the control panel in order to save a new jabber server entry.

The timeout has been removed because we need now reset the server immediately
when the accept button is pressed.

The change-signal handler has been removed because we only want to change the
server name when the user clicks on accept button.

An apply flag has been added so if the user clicks on accept, we know
whether or not the server should be updated.

### DIFF
--- a/extensions/cpsection/network/view.py
+++ b/extensions/cpsection/network/view.py
@@ -17,7 +17,6 @@
 
 from gi.repository import Gtk
 from gi.repository import Gdk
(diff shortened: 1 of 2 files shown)


SAMdroid-apps:
Merged!

PR #465, opened 2014-12-31
Use SugarExt.clipboard_set_with_data

tchx84:
As described in SL#4307, Gtk.Clipboard.set_with_data is not
introspectable, use SugarExt.clipboard_set_with_data to access
it. Use Gtk.Clipboard.set_with_data if available.

Fixes SL#4307.

Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/182 before merge.

### DIFF
--- a/src/jarabe/frame/clipboardicon.py
+++ b/src/jarabe/frame/clipboardicon.py
@@ -19,6 +19,7 @@ import logging
 
 from gi.repository import Gtk
 from gi.repository import Gdk
+from gi.repository import SugarExt
 
 from sugar3.graphics.radiotoolbutton import RadioToolButton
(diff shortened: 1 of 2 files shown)


godiard:
@tchx84 , the problem is, wh this fix we can crash sugar. I think is better don't have copy/paste than the risk of a complete crash

tchx84:
@godiard have you found another scenario where it crashes? Can you describe how to reproduce it?
 
Otherwise I think we should include it. Because this new PR includes a version that does not crash Sugar anymore.

tchx84:
The toolkit part is now in https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/189

godiard:
merged

PR #466, opened 2015-01-04
Background journal

i5o:
Add "Backgrounds" option in Copy to journal volume menu
First, this copy the image to Document folders, and then set it as background.


godiard:
I don't think add more options to the palette is a good idea. Is really too crowded now, and this is not a option you use usually. I would prefer add the possibility of get a image from the Journal in the control panel background section

walterbender:
I think that the submenus are not so crowded and it has proven the case that the background mechanism is a bit too obscure from the control panel widget.

godiard:
@walterbender, ok, no problem with disagree. Anyway would be good discuss this change when we start with 0.106

i5o:
I think it's time to discuss this :)

godiard:
@iso, really? you wait until we freeze features?

i5o:
May 18 2015 	Sucrose 0.105.1 unstable tarballs & release 	Features freezed 
https://wiki.sugarlabs.org/go/0.106/Roadmap

AFAIK, we have still have 2 days.

godiard:
@i5o, exactly....

samdroid-apps:
Moving this one to the 0.108 milestone.

@i5o, let's aim for a discussion at the **start** of the cycle if you want to merge it.

i5o:
Ok now, yes. 

I would like to start a discussion for include this in sugar 0.108

quozl:
I'd like this in 0.108.

Code reviewed.  Not tested.

samdroid-apps:
Reviewed, not tested.  Please see my comments.

Also, why do we copy it to the documents folder?  We should copy it to a folder (somewhere in the .sugar folder) where the user is less likely to delete it from.

quozl:
The "use this image as background" is a common use case, and making it more discoverable in this way lowers the barrier to entry for users.  It is a method of a journal object, and the journal lists the objects.  Hiding the feature in the control panel is less intuitive.

This also suggests we should review the uses of the object picker to ensure they are all included in the journal object context menu.

i5o:
Thanks for the comments guys.

@godiard I think we can include both, lets think about "new" users.. A menu option is more easy to figure out (or that is what I think), I don't exactly remember, but I think there was a "set as background" option in gnome too.

@samdroid-apps I will create two commits then. I choose to copy it to Documents folders to be able to see the background in the panel control.
But maybe is not necessary? Or maybe as u said, create a new folder for backgrounds.

Thanks @quozl  for review.

I will make a few changes and submit a new pr, for see what you guys thinks.
Thanks again

godiard:
@ii5o, please discuss the feature on the list.

PR #469, opened 2015-01-07
Show a wait cursor while the journal is updating - Fixes #4330

godiard:


SAMdroid-apps:
testing,,.,

SAMdroid-apps:
How does this fix the bug?  I think we should merge the patch, but I'm not sure we should close the bug.

Ticket Info:

> #4330 closed defect (fixed)
> # Journal: start editing unwanted entry
> Reported by: 	erikos 	Owned by: 	erikos
> Priority: 	High 	Milestone: 	Unspecified
> Component: 	Sugar 	Version: 	0.98.x
> Severity: 	Major 	Keywords: 	
> Cc: 		Distribution/OS: 	OLPC
> Bug Status: 	Unconfirmed 		
> Description
> 
> Steps to reproduce:
>
>    open an Activity (e.g. Log) 
>
>    leave it open and now open a second one (e.g. Memorize) 
>
>    close the second one 
>
>    go to the Journal 
>
> ---> both will have the correct order
>
>    wait one or two minutes 
>
>    now resume the first activity 
>
>    then go back to the Journal and directly try to go to the detail view of the second activity (Memorize) 
>
> ---> the view is updated and since the first activity (Log) has been accessed just now it will be updated and moved to the top of the Journal but already when the Journal is shown, so when you tap quickly you likely will go to the detail view of Log instead of Memorize

godiard:
Maybe do not fixes the bug, but makes more evident to the user the Journal is updating. The user still can click in a entry where is changing.

SAMdroid-apps:
Merged!

PR #471, opened 2015-01-08
Hide the Speech TrayIcon if GST or the espeak module isn't installed

ignaciouy:
This patch disable speech trayicon, in case gst or espeak module
isn't installed


godiard:
merged

PR #472, opened 2015-01-11
Get audio and video mimes from GStreamer during intro

SAMdroid-apps:
Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/194


SAMdroid-apps:
In reply to your ticket comment @godiard, I will send a new patch.

PR #473, opened 2015-01-15
Get audio and video mimes from GStreamer during first start

samdroid-apps:
Replaces #472

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -358,6 +358,11 @@ def _start_intro(start_on_age_page=False):
     window.show()
 
 
+def _start_first_run_tasks():
+    # Generate audio and video mimes from GStreamer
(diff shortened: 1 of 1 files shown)


quozl:
Has a bug label, but can't find a bug in trac, what bug number is it?

godiard:
@quozl  http://bugs.sugarlabs.org/ticket/3826

quozl:
Thanks.  Actually looks like a feature that happens to have a bug number.  ;-)

samdroid-apps:
@quozl Your right, this is a feature.

I'm not 100% sure this is a very necessary feature looking back on things.  To be honest, 5sec extra 1st startup time wait seems like a lot to have the right mime type icons - especially when there is no way for activities to use this data.

I think we can close this patch.

PR #474, opened 2015-01-19
autopep8, flake8 to sugar code

ignaciouy:
This fixes the flake8 errors in 'osbuild check sugar'
Pre-patch: (check-modules.log) http://fpaste.org/171272/
After-patch (check-modules.log) http://fpaste.org/171273/

NOTE: As you can see in http://fpaste.org/171273/ the testscases
are broken. We should fix it asap.

### DIFF
--- a/bin/sugar-launch
+++ b/bin/sugar-launch
@@ -50,6 +50,7 @@ cmd_args = activityfactory.get_command(activity)
 if args.uri:
     cmd_args.extend(["--uri", args.uri])
 
+
 def _which(exec_file):
     if 'PATH' in os.environ:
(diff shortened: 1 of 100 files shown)


godiard:
As in https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/197, I would prefer separate the changes, and merge only the changes neeeded to fix the tests.

tchx84:
As I mentioned in the toolkit patch, it is too late for these changes at this stage of the release. Lets just focus on fixing specific bugs and let do a major cleanup at the very beginning of the next release cycle.

PR #475, opened 2015-01-19
Flush the busy cursor on homeview load

SAMdroid-apps:
Fixes the issue reported by @quozl, in relation to commit 6c8141ba.


godiard:
Merged, thanks @SAMdroid-apps and @quozl

PR #476, opened 2015-01-20
Select autocompleted activity search text

SAMdroid-apps:
This means that slow typers who are not looking at the screen will
will overwrite the autocomplete, instead of adding to it.

Thanks to @quozl for reporting and suggesting the solution.

### DIFF
--- a/src/jarabe/desktop/homebox.py
+++ b/src/jarabe/desktop/homebox.py
@@ -99,10 +99,13 @@ class HomeBox(Gtk.VBox):
                 toolbar.search_entry._icon_selected = [activity]
                 break
 
-        if len(toolbar.search_entry._icon_selected) == 1:
+        # Don't change the selection if the entry has been autocompleted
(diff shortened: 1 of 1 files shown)


tchx84:
Hello @SAMdroid-apps , can you describe some use cases so I can try?

SAMdroid-apps:
Try becoming a find an peck typer for a minute.  So, type in the activity name slowly with 1 finger and looking at the keyboard.  With the old version the user would end up with something like "Browsewse" in the search bar.

quozl:
Yes, that's a good example.  A tester should test with independent use cases as well, that's what I did to detect the problem.  Perhaps tchx84 hasn't been able to review the patch quickly because the commit doesn't give an example of the problem.  Hopefully that isn't needed.

godiard:
Tested and is a good fix, specially thinking in our younger users. Merged.

PR #479, opened 2015-03-04
Modified typo in sugar

exploreshaifali:
Changed "canceled" --> "cancelled".
File affected sugar/src/jarabe/frame/activitiestray.py

Fixes #4823


godiard:
Note that this change break translations and we are already in String freeze. (http://wiki.sugarlabs.org/go/0.104/Roadmap).
I think we need ask cjl or localization list if is ok include this change, and if not, revert.

tchx84:
I will revert this change for the release on Friday, we can re-include it after that.

PR #481, opened 2015-03-16
fix for #2974

quozl:
Fix for spurious "failed to start" message, solved while testing repeated Write-96 resume from journal on Fedora 18.

http://bugs.sugarlabs.org/ticket/2974

### DIFF
--- a/src/jarabe/model/neighborhood.py
+++ b/src/jarabe/model/neighborhood.py
@@ -417,7 +417,7 @@ class _Account(GObject.GObject):
             error_handler=self.__set_current_activity_error_cb)
 
     def __set_current_activity_cb(self):
-        logging.warning('_Account.__set_current_activity_cb')
+        logging.debug('_Account.__set_current_activity_cb')
 
     def __set_current_activity_error_cb(self, error):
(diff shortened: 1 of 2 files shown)


godiard:
Thanks for working on this issue.
I think we should include it. 
Can you fix this simple pep8 error, and send  a new pr?
```
pep8 src/jarabe/model/shell.py
src/jarabe/model/shell.py:742:37: E128 continuation line under-indented for visual indent
```

tchx84:
+1, new PR with pep8 fixes and goes in :)

quozl:
Okay, thanks, expect to do that within four weeks.

godiard:
@quozl, I can that for you if you are busy. Just don't want steal the credit.

quozl:
Well, since the PEP8 detection was you, and not me, you wouldn't be stealing any credit.  ;-)  If I was a maintainer, I'd merge the patch and immediately commit a PEP8 patch referring to the PEP8 regression introduced in the previous patch ...

godiard:
Ok. Pushed and added another patch with the pep8 fix

PR #485, opened 2015-04-16
Allow users to click disabled control panel icons, fixes #4833

SAMdroid-apps:
**Requires https://github.com/sugarlabs/sugar-artwork/pull/49**

This commit styles the disabled icons with CSS instead of using the
insensitive style, allowing users to click disabled icons.  This is
more consistent with the design in the home view as well as more
accessible.

Ticket URL:  <http://bugs.sugarlabs.org/ticket/4833>


SAMdroid-apps:
Hum, so close this @godiard?

godiard:
Yes. Let's go with the other way, disable icons in the home when are filtered out.

PR #486, opened 2015-04-16
Update copyright date to 2015 in my settings

SAMdroid-apps:
We have made changes in 2015, so I think we can change it :smile:


godiard:
merged

PR #487, opened 2015-04-17

SAMdroid-apps:
makes the bundleregistry keep a list of alternative/outdated
activities stored on the system.  When uninstalling, this list is
used to find an alternative copy.


tchx84:
Nice, I will give it a try!

SAMdroid-apps:
I will fix that.

PR #488, opened 2015-04-18

SAMdroid-apps:
Replaces #487 

### DIFF
--- a/src/jarabe/model/bundleregistry.py
+++ b/src/jarabe/model/bundleregistry.py
@@ -548,6 +548,39 @@ class BundleRegistry(GObject.GObject):
         bundle.uninstall(force, delete_profile)
         self.remove_bundle(install_path)
 
+        alt_bundles = self.get_system_bundles(act.get_bundle_id())
+        if alt_bundles:
+            alt_bundles.sort(
+                key=lambda b: NormalizedVersion(b.get_activity_version()))
+            alt_bundles.reverse()
+            new_bundle = alt_bundles[0]
+            self.add_bundle(new_bundle.get_path())
+
+    def get_system_bundles(self, bundle_id):
+        """
+        Searches for system bundles (eg. those in /usr/share/sugar/activities)
+        with a given bundle id.
+
+        Prams:
+            * bundle_id (string):  the bundle id to look for
+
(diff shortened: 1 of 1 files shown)


quozl:
Reviewed. You are also handling the possibility of multiple versions of the same bundle. Well done. Did not test.

tchx84:
Hello @SAMdroid-apps,

This looks way better, specially since it can cover more cases than the previous approach. There seem to be a few problems though, because is not working:

```
  File "/home/tch/Devel/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/desktop/activitieslist.py", line 565, in __erase_confirmation_dialog_response_cb
    registry.uninstall(bundle, delete_profile=True)
  File "/home/tch/Devel/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/model/bundleregistry.py", line 551, in uninstall
    alt_bundles = self.get_system_bundles(act.get_bundle_id())
  File "/home/tch/Devel/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/model/bundleregistry.py", line 574, in get_system_bundles
    for activity_dir in os.listdir(root):
OSError: [Errno 2] No such file or directory: '/usr/share/sugar/activities'
```

quozl:
No, don't check for existence first, that introduces a race; instead check for an exception and handle it.

tchx84:
@quozl can you elaborate on how checking for a path existence before doing operations on it can creates a race condition?

godiard:
@tchx84, if the path is removed just after you do the check :)

quozl:
@tchx84, as @godiard says.  in this scenario it increases the possible error paths.  result is harmless.  so it is a comment on code design that is not critical.  it also increases the processing to handle a scenario that is rare.  more efficient to do the operation and handle the failure than to do an extra operation to check for possible failure first.

tchx84:
Alright, lets just make sure that whatever goes to the log when it fails is in DEBUG mode, because at least one of the directories will be missing, resulting in unnecessary logs entries. ie., just check the error I just showed while running code in sugar-build, it means that the path return by GLib.get_system_data_dirs() doesn't necessarily exist, thus os.listdir() will fail often.

PR #489, opened 2015-04-21

SAMdroid-apps:
Replaces #488


quozl:
Reviewed.

tchx84:
Hey @SAMdroid-apps please check that last one comment I left, after that I think is ready to go in.

PR #490, opened 2015-04-22

samdroid-apps:
displaying a /usr/share copy of the same activity. This patch

Ticket URL: http://bugs.sugarlabs.org/ticket/4836


SAMdroid-apps:
Replaces #490

SAMdroid-apps:
Replaces #489

quozl:
Reviewed.

tchx84:
Merged!

PR #491, opened 2015-04-23
Bundle launcher

samdroid-apps:
## Add BundleLauncher

The doc string for `jarabe.model.bundlelauncher.launch_bundle`
provides a good explanation of the api.


samdroid-apps:
Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/209

godiard:
@samdroid-apps, idem here, please move all the pep8 fixes unrelated to the feature to another pr. We can merge easily that before start to make changes

PR #492, opened 2015-04-23
Add BundleLauncher

samdroid-apps:

The doc string for `jarabe.model.bundlelauncher.launch_bundle`
provides a good explanation of the api.

PEP8 Compliant

Replaces #491


samdroid-apps:
Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/210

PR #493, opened 2015-04-24
Socialhelp try2

samdroid-apps:
Using the design discussed on the mailing list.

screenshot from 2015-04-25 09 45 30

Requires https://github.com/sugarlabs/sugar-artwork/pull/46


godiard:
Tested and works ok. Added a few questions, but is in a very good shape.

PR #495, opened 2015-05-04
Revert nm introspection

godiard:
This is a refactoring of the patch provided by James Cameron [1]
to use it in Sugar upstream.
Is compatible with XO or other devices.

[1] http://lists.sugarlabs.org/archive/sugar-devel/2015-April/049970.html


quozl:
Reviewed, thanks.

PR #496, opened 2015-05-04
Revert nm introspection try2

godiard:
Replaces https://github.com/sugarlabs/sugar/pull/495

This is a refactoring of the patch provided by James Cameron [1]
to use it in Sugar upstream.
Is compatible with XO or other devices.

[1] http://lists.sugarlabs.org/archive/sugar-devel/2015-April/049970.html

### DIFF
--- a/extensions/cpsection/aboutcomputer/model.py
+++ b/extensions/cpsection/aboutcomputer/model.py
@@ -23,10 +23,9 @@ import errno
 import time
 
 from gi.repository import Gio
-from gi.repository import NMClient
-from gi.repository import NetworkManager
 
 from jarabe import config
+from jarabe.model.network import get_wireless_interfaces
 
 
 _OFW_TREE = '/ofw'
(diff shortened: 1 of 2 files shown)


tchx84:
Tested and that comment is all I noticed wrong, please merge it after the change :)

quozl:
@tchx84, only a build number is in parenthesis, and that's an artefact of the builder, not Sugar.

@godiard, I've not tested this patch yet on commodity hardware, but I think some of it is so specific to the XO that it might need to be protected by a conditional checking for XO first.  Users of commodity hardware would not need to know firmware versions.

quozl:
Reviewing 13.2.3 for XO-4 which has this patch applied by builder, there are three sections;

* identity,
* software, and;
* copyright and license.

We have hardware details spread across the ''Identity'' and ''Software'' sections.  Perhaps this is inconsistent?
screenshot of _home_

tchx84:
This is how it looks like:

untitled

quozl:
@tchx84, thanks, very strange, and all caused by decisions outside this code?

Here is screenshot before patch on commodity hardware running Ubuntu.  Note extra gap after *Build*; extra newline character?  Note missing *Serial Number* and *Wireless Firmware*, so we can avoid showing it.

screenshot of _home_

Perhaps we should change *Identity* to *Hardware*, and move *Firmware* sections to it?

quozl:
Here is screenshot after patch on commodity hardware running Ubuntu.  Same extra gap after *Build*. Same newline before wireless driver as @tchx84 can see.

screenshot of _home__1

godiard:
@quozl, what are your proposals?

quozl:
@godiard, sorry, my code review training in ISO9001 says only make observations, not to make proposals.  You may have to remind me to suggest proposals.

Observations: with some values multi-line, and some other values single-line, seems inconsistent.  There are many other inconsistencies in this UI, which we can ignore for the moment.

Proposals: please see https://github.com/quozl/sugar/tree/revert_nm_introspection-try3-quozl-1 for three small patches on top of this pull request:
* avoid blank space below build,
* detect XO-1 wireless hardware, and;
* avoid split of wireless firmware words.

Tested on XO-1 with Fedora 18, and commodity hardware with Ubuntu Trusty.

godiard:
@quozl, sorry if sound rude, I just was confused by the comments and screenshots, due that some of they were not related to the changes in the pr. I agree with your changes, I will push all.

PR #497, opened 2015-05-05
Add accelerators for control panel open and close

quozl:
There is no accelerator for opening or closing control panel.  This hinders use, requiring mouse or touchscreen.  See #4660.

View Source has *Alt+Shift+V* for open, and *Escape* for close.  Activities have *Ctrl+q* for close.  Wireless Key Required dialog has *Escape* for close.

Patch adds accelerators:

1. *Alt+Shift+M* to open control panel, valid in any Sugar screen or activity, and;
1. *Escape*, *Ctrl+q* or *Ctrl+Shift+q* to close control panel or sections.

Close accelerator is added in the key event callback, rather than using ToolButton accelerators, because they do not work in the control panel.

Test case: use *Alt+Shift+M* to open control panel, verify *q* alone, and *Shift+Q* do insert into search box, select a section, verify *Ctrl+q*, *Ctrl+Shift+Q* and *Escape* closes section, and closes control panel.

References:
* [HotKeys](http://wiki.sugarlabs.org/go/Hotkeys) page on Wiki,
* [Ticket #4660](http://bugs.sugarlabs.org/ticket/4660) (keyboard navigation of Sugar), and;
* [Pull Request #443](https://github.com/sugarlabs/sugar/pull/443) (keyboard navigation of control panel).

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -207,6 +207,16 @@ class ControlPanel(Gtk.Window):
         self.grab_focus()
 
     def __key_press_event_cb(self, window, event):
+        if (event.keyval == Gdk.KEY_Escape) or \
+           ((event.keyval == Gdk.KEY_q or event.keyval == Gdk.KEY_Q) and \
+           (event.state & Gdk.ModifierType.CONTROL_MASK)):
+            if self._toolbar == self._main_toolbar:
+                self.__stop_clicked_cb(None)
+                self.destroy()
+            else:
+                self.__cancel_clicked_cb(None)
+            return True
+
         # if the user clicked out of the window - fix SL #3188
         if not self.is_active():
             self.present()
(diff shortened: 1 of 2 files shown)


samdroid-apps:
Reviewed.

1.  This patch allows you to open many copies of the control panel.  Pressing `Alt-Shift-M` should not open another panel if there is currently one open.  Other shortcuts (view source tested) already deal for this edge case

2.  `Esc` is good and conststant, however other dialog (view source tested) do not use `Ctrl-Q`.  Maybe only add `Esc`?

This is a very good patch otherwise!

Also, another easy target for the control panel:  make `enter` do something on the grid view!

godiard:
Just a comment: In src/jarabe/model/shell.py, we have a array self._shell_windows,
and two methods push_shell_window() and pop_shell_window() to track the windows opened,
and close them when needed.
Probably we should add a function has_shell_window() then we can control if a window is opened before opening one more.

samdroid-apps:
Or maybe only have 1 global instance of the control panel?  Then you can only show and hide the window - and that is all you need!

PR #494, opened 2015-05-01
Socialhelp popup try3

samdroid-apps:
Replaces #493 

Changes:

* Use old style formatting for translated strings
* Use new style "apply" button system in the settings

PR #498, opened 2015-05-05
Socialhelp popup try4

samdroid-apps:
Replaces #494 

Requires https://github.com/sugarlabs/sugar-artwork/pull/50

Changes:

* Add loading/flashing icon
* Deal with edge case for users with no server configured
* Bug fixes for @godiard's comments


godiard:
Pushed with two pyflakes fixes.

PR #499, opened 2015-05-06
Add accelerator for control panel close

quozl:
There is no accelerator for closing control panel.  This hinders use, requiring mouse or touchscreen.  See #4660.

View Source has Escape for close.  Wireless Key Required dialog has Escape for close.

Patch adds an accelerator Escape, to close control panel or sections.

Accelerator is added in the key event callback, rather than using ToolButton accelerators, because they do not work in the control panel.

Test case: open control panel, select a section, verify "Escape" closes section, and closes control panel.

References:

http://wiki.sugarlabs.org/go/Hotkeys
http://bugs.sugarlabs.org/ticket/4660


quozl:
pep

PR #505, opened 2015-05-10
Use tick based animation for the desktop and frame

samdroid-apps:
Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/213

### DIFF
--- a/src/jarabe/desktop/transitionbox.py
+++ b/src/jarabe/desktop/transitionbox.py
@@ -55,7 +55,7 @@ class TransitionBox(ViewContainer):
                                 pixel_size=style.XLARGE_ICON_SIZE & ~1)
         ViewContainer.__init__(self, layout, self._owner_icon)
 
-        self._animator = animator.Animator(0.3)
+        self._animator = animator.Animator(0.3, widget=self)
         self._animator.connect('completed', self._animation_completed_cb)
 
     def _animation_completed_cb(self, anim):
--- a/src/jarabe/frame/frame.py
+++ b/src/jarabe/frame/frame.py
@@ -114,7 +114,7 @@ class Frame(object):
             self._animator.stop()
 
         palettegroup.popdown_all()
-        self._animator = animator.Animator(0.5)
+        self._animator = animator.Animator(0.5, widget=self._top_panel)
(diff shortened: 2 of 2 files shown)


samdroid-apps:
Could we merge this?

godiard:
I think this is a interesting change, but we need test in XO-1 and the other models, and with different gtk versions. I don't see anything wrong in the code, but want do a good test. The changes are small, then (probably) we can merge after freeze.

samdroid-apps:
We have passed the String/UI/API freeze, but would we be able to merge this or do we need to push this back to 0.108?

godiard:
No without testing and performance comparison in all our common hardware (XO-1. XO-1.5, XO-1.75, XO-4). Maybe you can coordinate with Ignacio.

samdroid-apps:
Ok, let's aim to 108 for this patch!

quozl:
Need a test plan, see sugarlabs/sugar-toolkit-gtk3#213 for my comment there.

samdroid-apps:
Feature page:  http://wiki.sugarlabs.org/go/Features/Tick_based_animation

quozl:
Thanks.  Tested on commodity hardware with GTK+ 3.10.  The transitionbox animation was much smoother.  The frame animation was slightly different but was still stepped.  Measuring CPU cycles during test; the frame show operation is CPU bound.  It was surprising how much CPU time was needed for such a straightforward animation, and I wonder if there is some other underlying problem in the existing code.

samdroid-apps:
@quozl Thanks for testing.  I attempted to reproduce the issue with the frame animation using all of the cpu, however it had no noticeable impact on cpu usage (Fedora 23 Rawhide on Core2Duo laptop).

My 2c is that there have probably been some window manager fixes since your release.  For example, I noticed that the grey boxes usually left while the frame is animating had gone away.  But IDK :smile:

quozl:
My test was with Ubuntu 14.04.2 amd64.  Window manager should not be involved, as these windows are undecorated.  The CPU usage was by the shell process.  I used top(1) with 0.1 second update interval.  Yes, it would probably be very fast on a core2duo, wish they were cheap enough to use in a wide deployment.  :grinning:

PR #502, opened 2015-05-06
Add BundleLauncher

samdroid-apps:
Replaces #501 

Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/211

Changes:

* Alphabetize makefile
* Remove unneeded bundlelauncher code

The BundleLauncher is a dbus service served from within the sugar
shell process.  The BundleLauncher service lets activities launch
other activities, with a given bundle_id, mime_type or object_id.

PR #506, opened 2015-05-12
Add BundleLauncher

samdroid-apps:
Replaces #502 

Requires sugarlabs/sugar-toolkit-gtk3#211

Changes:

1.  Move null check to `get_bundle`
2.  Check `metadata['mime_type']` exists
3.  Replace `with_parents=True` with a function on journal.misc


godiard:
merged

PR #510, opened 2015-05-14
Brightness controls try8

tchx84:
This PR is for adding brightness control capabilities to Sugar. It includes a generic mechanism to interact with backlight devices, a jarabe model that acts as an abstraction for the sugar shell, keyboard control keys and a device frame UI.

PR #511, opened 2015-05-14
Brightness controls try9

tchx84:
This PR is for adding brightness control capabilities to Sugar. It includes a generic mechanism to interact with backlight devices, a jarabe model that acts as an abstraction for the sugar shell, keyboard control keys and a device frame UI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ data/sugar.gtkrc
 data/sugar.xml
 data/sugar-xo.gtkrc
 data/sugar-emulator.desktop
+data/org.sugar.brightness.policy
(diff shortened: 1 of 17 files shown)


tchx84:
This PR replaces https://github.com/sugarlabs/sugar/pull/510, making Brightness model constant _private.

tchx84:
Github meeses with the commits order, to see the original order please take a look at https://github.com/tchx84/sugar/commits/brightness-controls-try9

tchx84:
This also needs https://github.com/sugarlabs/sugar-artwork/pull/51

samdroid-apps:
Since this patch adds an almost required polkit policy, could we add that to the makefile or the docs or something?

PR #513, opened 2015-05-15
Use tick based animation for the pulsing icon

samdroid-apps:
This uses the new gtk tick callbacks for smoother pulsing icon
animations.  It is still backwards compatible for environments
without Gtk 3.8, which adds the callbacks.

The animations are so smooth now.  They almost look like completely different animations!  Video:  https://vid.me/HEJ9


samdroid-apps:
I think this significantly slows activity startup times (no tests - just observations on XO4).  I will investigate.

godiard:
@samdroid-apps, thanks.

PR #514, opened 2015-05-16
Correctly apply valignment in journal treeview, fixes #4852

samdroid-apps:
This commit fixes the sroll state resuming in the journal code.  The
state was being restored before the tree view was heigh enough for
the Gtk.ScrolledWindow to scroll to the correct position.  This patch
changes the scroll state only after the treeview has been drawn.

Ticket Link:  http://bugs.sugarlabs.org/ticket/4852


samdroid-apps:
Replaced by #519

PR #515, opened 2015-05-17
Don't change star color on hovering - Fixes #3147

godiard:
Change the star color on hovering cnfuses to the user,
because think the favorite state has changed.
This can be seen in the activities list view, in the Journal listview,
and in the Journal detail view.


samdroid-apps:
Looks good!  We already have the hover states so changing the star color is not needed.

PR #516, opened 2015-05-20
Fix trailing whitespace on firmware version

quozl:
On most XO laptops the firmware version doesn't have the engineer's build number appended, and this results in "Q7C05   with 0.5.01" shown.  Oh the irony.

Strip the trailing whitespace before using the version number.


godiard:
Thanks, merged

PR #517, opened 2015-05-22
Disable overlay scrolling, fixes #4854

samdroid-apps:
This commit disables overlay scrollig in Sugar.  Although this should
be set in the sugar theme, that is not possible.

Setting the environment variable disables the overlay scrolling for
both Sugar Shell and all sub processes (eg. activities).

Ticket URL:  http://bugs.sugarlabs.org/ticket/4854

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -34,6 +34,9 @@ import shutil
 reload(sys)
 sys.setdefaultencoding('utf-8')
 
+# Disable overlay scrolling - before GTK is loaded
+os.environ['GTK_OVERLAY_SCROLLING'] = '0'
+
 import gettext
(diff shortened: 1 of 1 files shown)


quozl:
Typo in commit.  s/scrollig/  Reviewed.  Not tested.  Comment added to ticket:
> You say it Fedora 22 which introduced the feature, but was it the Gtk+ version which did it rather than Fedora 22?  Can you confirm your patch has no effect on systems with earlier versions of Gtk+?

If you amend your commit, please add something to explain that this change has no effect on earlier versions of whatever.

godiard:
@samdroid-apps, Please add  LIBOVERLAY_SCROLLBAR=0 too, to make it work in ubuntu.

PR #518, opened 2015-05-24
Add escape key to close help

quozl:
For consistency with other modal dialogs such as view source and control panel, add a handler to translate escape key into stop button.

Tested on XO-4 with 14.1.0 and Sugar 0.105.1, using this method:
* change file on target, restart display manager:
```
rsync src/jarabe/view/viewhelp.py \
    target:/usr/lib/python2.7/site-packages/jarabe/view/viewhelp.py
ssh target service olpc-dm restart
```
* start Implode activity, use Alt-Shift-H to display help window,
* press Escape, and window does close,
* check for interesting logs.


godiard:
Thanks, merged

PR #519, opened 2015-05-27
Fix favorite view scroll to top

samdroid-apps:
Replaces #514

This patch changes the way favorite changes are done in the journal.
Now, the user clicks the icon, which then adds a patch to the list
view and redraws that square - this removes the whole redo the model
dance that happened before.

1/2 of a fix to #4852

### DIFF
--- a/src/jarabe/journal/listmodel.py
+++ b/src/jarabe/journal/listmodel.py
@@ -92,8 +92,15 @@ class ListModel(GObject.GObject, Gtk.TreeModel, Gtk.TreeDragSource):
         # avoid hitting D-Bus and disk.
         self.view_is_resizing = False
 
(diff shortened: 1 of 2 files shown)


samdroid-apps:
If this looks OK, I'll make a similar patch for the title modification.

godiard:
@samdroid-apps, What a nice hack! :)
I continued working, based in your work, and sent https://github.com/sugarlabs/sugar/pull/520
The most important changes are explained in the header.
I have removed the use of tree_view.convert_tree_to_widget_coords because when the treeview is not in the first page (is scrolled down) return wrong "y" values.
Please test deeply. I have observed at times activities start after click in the title field, and that should not happen. This is a big change, we need be sure we don't break anything more.

samdroid-apps:
Replaces by #520, #524

PR #521, opened 2015-05-28
Disable overlay scrolling, fixes #4854

samdroid-apps:
This commit disables overlay scrolling in Sugar.  Although this should
be set in the sugar theme, that is not possible.

Setting the environment variable disables the overlay scrolling for
both Sugar Shell and all sub processes (eg. activities).

By setting an environment variable, this is ignored by Gtk versions
less than 3.16.

Ticket URL:  http://bugs.sugarlabs.org/ticket/4854

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -34,6 +34,10 @@ import shutil
 reload(sys)
 sys.setdefaultencoding('utf-8')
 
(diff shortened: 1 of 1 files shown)


quozl:
Reviewed.  Not tested.

godiard:
merged

PR #522, opened 2015-05-29
request for discussion: brightness controls: fedora 18

quozl:
On Fedora 18, XO-1.5 with 13.2.4 and Sugar 0.105.1 as upgrade, the display icon is missing from the frame and shell.log shows
```
AttributeError: type object 'GFileMonitorFlags'
    has no attribute 'WATCH_HARD_LINKS'
```

Using 0 for flag does not fail with an error, the display icon is present, can be used to change brightness, but changes to brightness made externally (using echo, or powerd, or olpc-kbdshim) do not move the slider.

inotifywatch does show events for changes to brightness, so kernel support is present, missing support is probably because of Gio.

This patch only suppresses the error.

@tchx84, others, can you suggest a fix?

### DIFF
--- a/src/jarabe/model/brightness.py
+++ b/src/jarabe/model/brightness.py
@@ -70,8 +70,12 @@ class Brightness(GObject.GObject):
         if not self.get_path():
             return
 
+        try:
+            flags = Gio.FileMonitorFlags.WATCH_HARD_LINKS
+        except:
+            flags = 0
         self._monitor = Gio.File.new_for_path(self.get_path()) \
-            .monitor_file(Gio.FileMonitorFlags.WATCH_HARD_LINKS, None)
+            .monitor_file(flags, None)
         self._monitor.set_rate_limit(self._MONITOR_RATE)
         self._monitor_changed_hid = \
             self._monitor.connect('changed', self.__monitor_changed_cb)


tchx84:
hmm, will take a look in the morning..  thanks for reporting it.

tchx84:
The problems is that _FILE_MONITOR_WATCH_HARD_LINKS was introduced in glib2 since 2.36 [1], and fedora 18 includes only 2.34 [2]. This means the Monitor will only detect changes to the path we specified when creating the monitor.

Upgrading glib2 for fedora 18 would be a bit too drastic for this problem. What we could do is to try to workaround the issue for XOs, taking advantage of the fact the device is writeable and IF is possible to make assumptions about which device is being used by other olpc modules.

Will make some tests and reply.

Refs:
1. https://developer.gnome.org/gio/stable/GFile.html#G-FILE-MONITOR-WATCH-HARD-LINKS:CAPS
2. http://koji.fedoraproject.org/koji/buildinfo?buildID=379212

tchx84:
The problems were that: (a) I was watching the whole directory and not the brightness file specifically, and (b) WATCH_HARD_LINKS is not necessary at all, I misunderstood its purpose.

@james, @godiard Just sent the proper fix in https://github.com/sugarlabs/sugar/pull/523

PR #523, opened 2015-05-30
Fix brightness monitor flags and file

tchx84:
Watch the specific brightness file instead of watching the whole
device directory and drop the use of WATCH_HARD_LINKS because is
not necessary.

This patch fixes the brightness monitor in Fedora 18.

### DIFF
--- a/src/jarabe/model/brightness.py
+++ b/src/jarabe/model/brightness.py
@@ -70,8 +70,9 @@ class Brightness(GObject.GObject):
         if not self.get_path():
             return
 
-        self._monitor = Gio.File.new_for_path(self.get_path()) \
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Looks good to me, the icon still updates when I use my keyboard shortcut.

Merged.

BTW:  Can we document the polkit policy somewhere?  Could we make it part of the makefile or document it on developer.slo?  The "sugar-brightness-helper needs root" popup is really annoying.

quozl:
@samdroid-apps, that popup does not occur when Sugar is run properly as a desktop.  The polkit policy is mentioned in Makefile.am, so make sure you've run autogen.sh.

PR #524, opened 2015-05-31
 Change model query in place for Journal ListView

samdroid-apps:
**Note**: this is based off #520 because it is a good performance hack and I am too lazy to git checkout back to master.  Sorry

Instead of replacing the ListModel, we now just change the result set
query.

This means that the buggy scroll position setting code is no longer
needed.

Steps to test:

* Go to the Journal
* Open an item in the details view
* Change title or star status
* Click back
* Notice how it has updated and the scroll position preserved

### DIFF
--- a/src/jarabe/journal/listmodel.py
+++ b/src/jarabe/journal/listmodel.py
@@ -99,6 +99,12 @@ class ListModel(GObject.GObject, Gtk.TreeModel, Gtk.TreeDragSource):
         self._result_set.ready.connect(self.__result_set_ready_cb)
         self._result_set.progress.connect(self.__result_set_progress_cb)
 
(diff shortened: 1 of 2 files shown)


samdroid-apps:
Could we review this?  It would be land one of the fixes for 106.

godiard:
@samdroid-apps, I am in doubt about merge this and #520 at this stage of the cycle,
without a serious test and be sure we don't introduce collateral  issues.

samdroid-apps:
It's a pretty bad bug to ship with though.

samdroid-apps:
Separated it from the other patch in https://github.com/sugarlabs/sugar/pull/542

PR #525, opened 2015-06-03
Fix Journal ListView DnD, fixes #4853

samdroid-apps:

Ticket URL  <http://bugs.sugarlabs.org/ticket/4853>


samdroid-apps:
Replaced by #526

PR #527, opened 2015-06-03
Restore brightness level earlier in startup

tchx84:
Load the brightness module earlier in the shell startup process
in order to restore the brightness level before any other screen
shows up, ie., the intro screen.


tchx84:
@quozl

quozl:
Thanks.  Reviewed.  Tested.  It works.  Also confirmed with inotifywatch that there is only one write to brightness in sysfs.

PR #528, opened 2015-06-12
Hide power control panel if powerd missing

quozl:
On commodity hardware the power control panel does nothing useful.

With this patch, if powerd is not present, the CLASS global is not set, causing the panel to be omitted from the control panel.

The message about missing CLASS is demoted from error to debug.

Tested on Ubuntu 15.04 Vivid with Sugar 0.105.2 (without powerd) and Fedora 18 with Sugar 0.104.1 (with powerd).


tchx84:
Thanks.

PR #529, opened 2015-06-17
Fix favorite selection in multiple home views

tchx84:
Use different Gtk.TreeViewColumn for each favorite column,
otherwise only the fist one can be clicked and prevents
populating the other home views.

Slightly cleanup that code.


tchx84:
@godiard  can you review? This is what German reported the other day.

PR #530, opened 2015-06-18
Fix logic hole in BundleRegistry.add_bundle

tchx84:
BundleRegistry uses Gio.FileMonitor to track changes in activity
installation directories. Any changes in these directories will be
detected and processed by the BundleRegistry.

When processed, the registry will simply assume it is a new bundle
and it will use "bundle_from_dir" helper to load it. The helper
has 3 potential results, (a) it returns a bundle, (b) it throwns
an exception if something went wrong while loading the bundle, or
(c) returns None if the directory is not a bundle.

BundleRegistry.add_bundle did not covered the case (c), therefore
it was failing when non-bundle directories were creates. E.g, when
package managers (such as dpkg) creates temporary directories.

This patch add an extra check to cover the missing case (c) from
when the "bundle_from_dir" helper.

### DIFF
--- a/src/jarabe/model/bundleregistry.py
+++ b/src/jarabe/model/bundleregistry.py
@@ -287,6 +287,11 @@ class BundleRegistry(GObject.GObject):
             logging.exception('Error loading bundle %r', bundle_path)
             return None
 
+        # None is a valid return value from bundle_from_dir helper.
+        if bundle is None:
+            logging.error('No bundle in %r', bundle_path)
+            return None
+
         bundle_id = bundle.get_bundle_id()
         logging.debug('STARTUP: Adding bundle %s', bundle_id)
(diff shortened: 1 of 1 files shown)


tchx84:
@godiard @quozl Please give it a try.

I tried other approaches, such as filtering path on the monitor callback, but is not really possible since content bundles does not have an extension (too late for changing that now). Another approach would be to extend the helper with an specific Exception for this case but is does not worth it to break the API because the helper is only used here.

tchx84:
@godiard if we go for throwing an exception we will need to also add an appropriate exception to the other helper (which cannot use the same exception because are different, maybe even a new Exception), and then make sure we listen to those exceptions where the helpers are being used, making this a bigger change. And arguably breaking the API.

I think we should just use the current helper semantic properly to fix this issue.

PR #534, opened 2015-06-22
translations: add custom entry for Guarani

tchx84:
A group of kids from the OLPC project in Caacupe have translated both
the sugar shell and the sugar gtk3 toolkit to Guarani. They are also in
the process of translating sugar activities [1].

Therefore, it would be a big encouragment for them if we could enable
these translations in the control panel.

This patch adds a custom entry for Guarani as is not yet included in
Fedora [2].

[1] http://translate.sugarlabs.org/gn/
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1207485

### DIFF
--- a/extensions/cpsection/language/model.py
+++ b/extensions/cpsection/language/model.py
@@ -62,6 +62,7 @@ def read_all_languages():
     # FIXME: This is a temporary workaround for locales that are essential to
     # OLPC, but are not in Glibc yet.
     locales.append(('Dari', 'Afghanistan', 'fa_AF.utf8'))
+    locales.append(('Guarani', 'Paraguay', 'gn.utf8'))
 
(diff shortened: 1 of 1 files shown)


tchx84:
This PR requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/230

tchx84:
Merged with @godiard approval.

icarito:
Hi,
I don't think this is the way to go. We've been handling adding missing locales at olpc-os-builder time - this really should be handled at image creation time, not within Sugar.
As a side effect - Sugar 0.104 currently wrongly displays Dari language in a clean Debian system installation, overriding system locale preferences.

quozl:
For reference, discussion has moved to #550.

PR #535, opened 2015-06-24
Simplify logic to apply jabber server config changes - Fixes #4865

godiard:
There multiples fixes here:
* On apply_jabber method, don't compare with the model._get_jabber
function, but with the result of executing the function.
* jabber set/get method don't throw ReadError exceptions,
then do not have sense cathit here, or have a jabber_alert.
Remove all the logic. Probably copied from __radio_toggled_cb,
but not needed here.
* Make apply_jabber private, is a internal implementation.


tchx84:
Tested, Merged.

PR #536, opened 2015-06-25
Compare string with a string, no with a function.

i5o:


i5o:
ups, nvm my fault.
Just saw https://github.com/sugarlabs/sugar/pull/535

quozl:
oops, the ticket didn't have a link to #535

PR #537, opened 2015-06-25
Help window: notify window closed o the shell on hide - Fixes #4866

godiard:
Instead of notify the window was closed in the stop button click event
do it when the window is hidden, because the window can be closed
when the activity associated is closed, and in that case the shell
was confused. More information in the ticket.


tchx84:
Tested, Merged.

PR #538, opened 2015-06-25
Fix journal translations try1

tchx84:
Some strings in the Journal are never translated, even when there are proper translations in the .PO.

To test:
1. Go to CP and switch language to Spanish (restart shell).
2. Once in Sugar again, go to Journal and click on Sort tool button.

Without this patch, only english strings can be seen. With the patch, the strings are properly translated.

EXTRA: Fixed a pep8 warning in that file, in a separate commit.

### DIFF
--- a/src/jarabe/journal/journaltoolbox.py
+++ b/src/jarabe/journal/journaltoolbox.py
@@ -25,7 +25,6 @@ import time
 from gi.repository import GObject
 from gi.repository import Gtk
 from gi.repository import Gdk
-from gi.repository import Pango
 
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Do we still need this now we have https://github.com/sugarlabs/sugar/pull/539?

godiard:
@samdroid-apps, this was merged before found the other problem and the more general solution iimplemented on #539

samdroid-apps:
Should we revert it then?

godiard:
Does not hurt, right?

PR #539, opened 2015-06-25
Earlier gettext initialization to asure translations are loaded

godiard:
One example solved by this patch are the generic mime type names
displayed in the 'What' filter in the Journal.
These names are defined in mime.py in the toolkit, and loaded at import time.


samdroid-apps:
Merged!

PR #540, opened 2015-06-25
Fix Journal ListView DnD, fixes #4853

godiard:

This patch is based in work from Sam Parkinson <sam.parkinson3@gmail.com>
and was tested on F18 and F22.

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -673,6 +673,7 @@ class ListView(BaseListView):
         self._is_dragging = False
 
         self.tree_view.connect('drag-begin', self.__drag_begin_cb)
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Looks good code wise!!!

I have not tested it.  I need to fix the speed of
sugar_runner_find_free_display to be less than 1 minute :)

godiard:
@samdroid-apps  what is sugar_runner_find_free_display? :)

samdroid-apps:
@godiard oh, part of sugar runner.  Just seemed to be a rawhide bug... testing your patch...

samdroid-apps:
Looks good, merging.

Interestingly we seem to have lost the "you can't copy an object that doesn't have a journal entry" alert, but that's probably another bug.

PR #542, opened 2015-06-26
Change model query in place for Journal ListView, fixes #4852

samdroid-apps:
Instead of replacing the ListModel, we now just change the result set
query.

This means that the buggy scroll position setting code is no longer
needed.  Infact, a lot of code is removed

Steps to test:

* Go to the Journal
* Open an item in the details view
* Change title or star status
* Click back
* Notice how it has updated and the scroll position preserved

Other things tested:

* Changing mountpoints
* Selections and changing mountpoints
* Selection preservation after refresh
* Changing the filter

### DIFF
--- a/src/jarabe/journal/listmodel.py
+++ b/src/jarabe/journal/listmodel.py
@@ -75,16 +75,14 @@ class ListModel(GObject.GObject, Gtk.TreeModel, Gtk.TreeDragSource):
 
     _PAGE_SIZE = 10
 
-    def __init__(self, query):
+    def __init__(self):
         GObject.GObject.__init__(self)
 
+        self._result_set = None
         self._last_requested_index = None
         self._cached_row = None
-        self._query = query
         self._all_ids = []
-        t = time.time()
-        self._result_set = model.find(query, ListModel._PAGE_SIZE)
-        logging.debug('init resultset: %r', time.time() - t)
+        self._query = None
         self._temp_drag_file_path = None
         self._selected = []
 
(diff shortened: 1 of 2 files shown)


samdroid-apps:
Ok, I will add the feature tag because this does do a lot of refactoring.

@godiard, @tchx84  Could this get merged for 106?  If not, maybe we could merge the more self-contained https://github.com/sugarlabs/sugar/pull/520 and target this for 108?  Or will we just ship 106 with the bug?

samdroid-apps:
Ok, let's aim to 108 for this patch!

quozl:
Is there a feature page?  I don't understand your pull request yet; it doesn't look like a feature, but rather an optimisation.  But I don't understand the impact.  Is there a bug number that would tell me more?

samdroid-apps:
Yes it is an optimization, but it also fixes http://bugs.sugarlabs.org/ticket/4852

quozl:
Thanks, I've reproduced 4852 now.

tchx84:
Is this a replacement for https://github.com/sugarlabs/sugar/pull/520?

godiard:
@tchx84, no, is a complement. This patch solves the case of going to the detail view, modify the metadata (change star, title or description) and go back to the listview

src/jarabe/journal/listmodel.py
```diff
@@ -92,8 +90,19 @@ def __init__(self, query):
         # avoid hitting D-Bus and disk.
         self.view_is_resizing = False
 
+    def set_query(self, query, clear_selection=True):
```
samdroid-apps: If we were merging this along side #520, we would need to reset the self._updated_entries table here.

godiard: Sorry @samdroid-apps I am slow with reviews this week. Could you prepare a new pr with the patch in #520, and this one modified to make easier test all together?

samdroid-apps: Can do, see #564

samdroid-apps:
Replaced with #564

PR #543, opened 2015-06-26
Main method names

godiard:
This is a cleanup, low priority.


tchx84:
The first patch seems necessary, since it makes no sense to leave that function now with that name after all the locales setup was moved somewhere else.

The second patch is unnecessary, specially at this stage of the release.

Please send a new PR with only the function rename, so we can fix the naming problem.

godiard:
@tchx84, ok sent https://github.com/sugarlabs/sugar/pull/544

PR #546, opened 2015-06-26
Don't check window type before change view - Fixes #4795

godiard:
The code removed was introduced on commit 9406c42e to avoid change
views when the My Settings window was opened, but that didn't worked,
and we added the control of modal windows on the shell on 14956762
to solve that case.
But this control have the lateral effect of block change the views
when a popup is opened. Remove this control solves the problem.

### DIFF
--- a/src/jarabe/model/shell.py
+++ b/src/jarabe/model/shell.py
@@ -463,12 +463,6 @@ class ShellModel(GObject.GObject):
         if old_level == new_level:
             return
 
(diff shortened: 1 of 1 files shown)


quozl:
Reviewed.  Tested on XO-1 with Sugar 0.105.2, does fix #4795.  Thanks!

tchx84:
Merged.

PR #547, opened 2015-07-01
Set default language when read .i18n file - Fixes #4868

godiard:
If the file is empty or corrupted, use the default language.


tchx84:
Tested and reviewed.

samdroid-apps:
Can reproduce bug.

Fix tested.  Merged.

PR #548, opened 2015-07-02
Add a dbus method to provide information about a bundle path

godiard:
This is used in the toolkit to provide the get_bundle functionality.


godiard:
This is needed by https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/236

tchx84:
In general both PRs looks OK, or least I can't notice anything else. I left one more comment  in https://github.com/sugarlabs/browse-activity/commit/7315fe7d9d951965a7ccd6341d9bc684517cbad8

PR #551, opened 2015-07-07
Bundle of Clipboard changes

samdroid-apps:
Feature Page:  http://wiki.sugarlabs.org/go/Features/Replace_GtkMenu

Requires:  https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/238 (and therefore https://github.com/sugarlabs/sugar-artwork/pull/69)

See the toolkit pull request for screenshots.

Hopefully I will also have a patch for the journal palettes.  But there are treeviews and invokers involved, and those 2 things have never gone well for me :smile:

I also think that this patch has a net loss of 3 lines of code!  Yay!

### DIFF
--- a/src/jarabe/frame/clipboardicon.py
+++ b/src/jarabe/frame/clipboardicon.py
@@ -19,6 +19,7 @@ import logging
 
 from gi.repository import Gtk
 from gi.repository import Gdk
+from gi.repository import GdkPixbuf
 from gi.repository import SugarExt
 
 from sugar3.graphics.radiotoolbutton import RadioToolButton
@@ -132,9 +133,9 @@ class ClipboardIcon(RadioToolButton):
             return
 
         if cb_object.get_icon():
-            self._icon.props.icon_name = cb_object.get_icon()
+            self._icon.props.file = cb_object.get_icon()
             if self._notif_icon:
-                self._notif_icon.props.icon_name = self._icon.props.icon_name
+                self._notif_icon.props.icon_filename = self._icon.props.file
         else:
             self._icon.props.icon_name = 'application-octet-stream'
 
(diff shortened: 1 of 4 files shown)


samdroid-apps:
I really enjoy putting unrelated fixes together :smile:  Seriously though, I didn't want to write new GtkMenu code and cause conflicts.

I pushed 2 new commits:

```
commit ead7b100d44d0fe182b91b0cad52fead4417b6bf
Author: Sam Parkinson <sam.parkinson3@gmail.com>
Date:   Thu Jul 9 12:48:06 2015 +1000

    Better represent journal objects on the clipbaord
    
    Currently the clipboard just shows things like "Text Clipping" for
    journal objects.  However, journal objects should have a sense of
    wholeness and be better repesented in the clipboard [1].
    
    This patch shows metadata about journal objects when they are coppied
    to the frame.  It also shows a resume menu item.
    
    Test Cases:
    
    * Drag from journal to frame, click resume
    * Drag journal entry from the frame to the documents folder
    
    [1] <wiki.sugarlabs.org/go/Design_Team/Specifications/Clipboard>

commit cc3775e88c52a972c55614dca96411011249d910
Author: Sam Parkinson <sam.parkinson3@gmail.com>
Date:   Thu Jul 9 12:40:41 2015 +1000

    Recycle paths during journal copying, fixes #4876
    
    Use the same path if the same journal entry is requested for drag
    data twice.  This means that the old path doesn't get deleted and the
    clipboard always gets a good path.
    
    <http://bugs.sugarlabs.org/ticket/4876>
```

godiard:
All is good material, but putting in the same pr make more difficult merge. Anyway we need wait for @tchx84 to see where/when merge.

samdroid-apps:
Ok, sorry about that.  If you want to you can just merge the commits separately

samdroid-apps:
Oops, found an id10t error :smile:

```
     def can_resume(self):
-        return self._get_journal_object is not None
+        return self._get_journal_object() is not None
```

PR #552, opened 2015-07-07
Don't let the label expand the icons in the clipboard tray

samdroid-apps:
| **Before** | **After** |
| ---------- | ---------- |
| 1before | 1after |

### DIFF
--- a/src/jarabe/frame/clipboardicon.py
+++ b/src/jarabe/frame/clipboardicon.py
@@ -56,6 +56,10 @@ class ClipboardIcon(RadioToolButton):
         self.set_icon_widget(self._icon)
         self._icon.show()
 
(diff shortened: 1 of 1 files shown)


godiard:
@samdroid-apps, I spent time trying to understand why this is happening and check if there are another way to solve it. I sent a pr to solve the issue in the toolkit to avoid the problem appear in other places, and added a explanation with the source of the problem. 
Let me know what you think.
In general, would be good if you explain more in the commit headers the reason of the changes. Is the only comment I can do, other than that is veeeeery good :) I will try review and merge the other prs.

godiard:
...and I filled a ticket (#4875). Tickets are a permanent history for us, comments in prs are temporary or difficult to obtain in the future.

samdroid-apps:
Merged alternative patch

PR #553, opened 2015-07-08
Delay reset home cursor at startup if window is not created - Fixes #…

godiard:
…4873

This patch delay the reset of the Clock cursor set at Sugar startup
until the window is created. The bug is probably originated by a
race condition and is difficult to reproduce.

### DIFF
--- a/src/jarabe/desktop/favoritesview.py
+++ b/src/jarabe/desktop/favoritesview.py
@@ -622,13 +622,23 @@ class CurrentActivityIcon(CanvasIcon):
             self.props.xo_color = self._home_activity.get_icon_color()
 
             if self._home_activity.is_journal():
-                self.get_window().set_cursor(None)
+                if self.get_window():
+                    self.get_window().set_cursor(None)
+                else:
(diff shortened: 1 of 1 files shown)


godiard:
@quozl, I didn't find a way to reproduce the problem, testing on a XO-1
I tried removing .sugar, setting many activities as favorites or only two.
If you have a way to reliably reproduce this, would be good to know.
cc this information in the ticket.

quozl:
Reviewed.

samdroid-apps:
@quozl did you test?

quozl:
No, and nor do I think it necessary.  A full test would require the problem to be easily reproducible, and as you know it isn't.  Gonzalo's patch does fix a theoretical race condition.  Do you have a doubt that you aren't expressing?

samdroid-apps:
No, I have no  doubts.  I just thought you were able to test it and reproduce it.

I'll test if it has any effect on the normal (non racy) running of the shell and merge accordingly.

samdroid-apps:
Merged!

PR #554, opened 2015-07-08
Set a env variable SUGAR_VERSION

godiard:
Browse activity need know the Sugar version to communicate to ASLO.
Until now, we updated Browse activity with a constant,
but is not a good solution.
Other activities can use this version, by example Log.

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -46,6 +46,9 @@ gettext.bindtextdomain('sugar', config.locale_path)
 gettext.bindtextdomain('sugar-toolkit-gtk3', config.locale_path)
 gettext.textdomain('sugar')
 
+# publish sugar version in the environment
+os.environ['SUGAR_VERSION'] = config.version
+
 from dbus.mainloop.glib import DBusGMainLoop
 DBusGMainLoop(set_as_default=True)
 


godiard:
Example of use in Browse activity https://github.com/sugarlabs/browse-activity/commit/2da272b02534331b2825284a52d0750f5ddbe00f

samdroid-apps:
Works like a charm.  With both patches the browse user agent successfully says "SugarLabs/0.106".

@tchx84 Can I merge this?  Are we in the features part of the cycle for 0.108 or are we in the 0.106.1 part of the 0.106 cycle?

quozl:
Reviewed.

Good idea, one less call to config.  Please also change Log.

Agreed, it is up to @tchx84 if merge window has opened. If so, 0.106.1 will be a branch.  If not, we hold on to pull requests until merge window opens.

tchx84:
Even though the change is extremely small (and effective), is it a feature. Lets merge it in 0.108.

quozl:
@godiard, @tchx84, and @quozl discussed this in a meeting just now.

samdroid-apps:
@godiard, @tchx84, and @quozl, what was the result of that meeting?  Can we merge this?

quozl:
@samdroid-apps, no result, but I thought it was important enough to say we discussed it.  :smile:

I defer to @tchx84.

godiard:
I did push of this, after @tchx84 approval. I did a cherry-pick to have it sorted in the git log.

PR #555, opened 2015-07-10
Intro: don't try setup non existent page - Fixes #4872

godiard:


tchx84:
@quozl can you describe how to reproduce this?

quozl:
@tchx84, no.  shell.log was captured as part of evaluating a different problem, so this was an incidental discovery.  I do not think it needs to be reproduced.

godiard:
@quozl, @tchx84, The issue will not happen _every_ time, because done() emit a signal, then that processing is async. If the signal is processed fast enough, the window is closed before try to prepare the wrong page. The fix is clear, if is the last page, we can't process one page more.

samdroid-apps:
Does not break normal setup paging.

Merged.

PR #556, opened 2015-07-21
More robust read/write of file .i18n

godiard:
We have received a few reports of broken ~/.i18n files
Try to make the code more robust, using "with" to open the files,
and remove redundancy in the code.

### DIFF
--- a/extensions/cpsection/language/model.py
+++ b/extensions/cpsection/language/model.py
@@ -83,13 +83,11 @@ def _initialize():
 def _write_i18n(lang_env, language_env):
     path = os.path.join(os.environ.get('HOME'), '.i18n')
     if not os.access(path, os.W_OK):
-        print _standard_msg
-        fd = open(path, 'w')
-        fd.write('LANG="%s"\n' % _default_lang)
-        fd.write('LANGUAGE="%s"\n' % _default_lang)
-        fd.close()
-    else:
-        fd = open(path, 'w')
+        # set the default values
+        lang_env = _default_lang
+        language_env = _default_lang
+
+    with open(path, 'w') as fd:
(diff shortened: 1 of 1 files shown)


quozl:
Thanks.  But Nak.  Lots of issues.  Let's work through them.

The method _write_i18n looks wrong; if file is not present, any data in arguments is lost, instead the defaults are set.  User choice would be lost.

(Before your patch, _write_i18n still looks wrong; if checks if the file is writable, and if not tries to write defaults to it, and loses the user choice.)

Also, the code in both methods should not test access before opening the file for write.  There's no reason to test, and it introduces a potential race.  So remove the call to os.access and instead handle an exception if open fails.

The most likely cause of broken ~/.i18n files on any system is that kernel buffer cache has not been flushed before a power fail.  The way to force this is to call fsync or fdatasync before close, which is the flush method of the file object.  Expect a stall of between 100ms and 2000ms depending on system load.  So do it in an idle after rendering the ok button state change.

The code that calls os.path.join is different in the two methods; in one place a default value is given to os.environ.get, and in the other place no default is given.  Suggest you write a _get_i18n_path method to factor this.

I'd prefer if the file is not changed as a result of only displaying the control panel for keyboard or languages; it should only change if the user makes a change.  So I don't think get_languages should call _write_i18n.

sugar.in executes .i18n if it exists.  This may race with olpc-configure on an XO laptop.  The control panel uses the .i18n to show the current language; which because of the race is often different on first boot ... instead it should use the environment variables set in sugar.in.

Hope that helps!

godiard:
Thanks @quozl, very valuable analysis. The actual code is really puzzling..

PR #557, opened 2015-07-22
More robust i18n try2

godiard:
A new try, with improvements based in @quozl feedback, and solving more issues found while testing.
Replaces #556


quozl:
Is Sugar the only user of .i18n file and are there no other values that other desktops may insert?

godiard:
I will send a new try, with the following changes:
* remove close, is not needed.
* make the code reading the lines in i18n independent of the  order of the lines, if anybody else is using it, or a human edit the file.
* try catch around the code writing the file and log a exception if needed

PR #558, opened 2015-07-23
More robust i18n try3

godiard:
Replaces https://github.com/sugarlabs/sugar/pull/557

### DIFF
--- a/extensions/cpsection/language/model.py
+++ b/extensions/cpsection/language/model.py
@@ -25,7 +25,6 @@ import locale
 from gettext import gettext as _
 import subprocess
 
-
 _default_lang = '%s.%s' % locale.getdefaultlocale()
 _standard_msg = _('Could not access ~/.i18n. Create standard settings.')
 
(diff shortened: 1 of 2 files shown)


godiard:
@samdroid-apps, I addressed your comments in the next patch (Before open....)

extensions/cpsection/language/model.py
```diff
@@ -81,49 +81,54 @@ def _initialize():
...
+
+    new_lang_line = 'LANG="%s"\n' % lang_env
+    new_language_line = 'LANGUAGE="%s"\n' % language_env
+
+    if lang_line == new_lang_line and language_line == new_language_line:
+        return
 
+    try:
+        with open(path, 'w') as fd:
+            fd.write(new_lang_line)
```
quozl: Lines 105 to 110 do not preserve previous unrelated contents of file.  Since the file can contain locale environment variables like currency and date format, this is impolite to other software. :smile:

extensions/cpsection/language/model.py
```diff
@@ -81,49 +81,54 @@ def _initialize():
...
+    if 'LANGUAGE' in os.environ:
+        lang = os.environ['LANGUAGE']
+        lang = lang.strip()
+        if lang.endswith('UTF-8'):
+            lang = lang.replace('UTF-8', 'utf8')
+        langlist = lang.split(':')
+
+    if 'LANG' in os.environ:
+        lang = os.environ['LANG']
+        lang = lang.strip()
```
quozl: Lines 127 to 131 seem remarkably similar to lines 120 to 124 (not 125), so perhaps a local function inside get_languages would ease reading.

extensions/cpsection/language/model.py
```diff
@@ -81,49 +81,54 @@ def _initialize():
...
 
+    try:
+        with open(path, 'w') as fd:
+            fd.write(new_lang_line)
+            fd.write(new_language_line)
+            fd.flush()
+            # be sure all is flushed
+            os.fsync(fd)
+    except:
+        logging.exception('Error writing .i18n file')
```
quozl: Line 112 logs an exception only.  There's no UI response, so the user will think the change will be effective, but it won't be.  Why not catch the exception in caller and revert the change?

godiard:
Replaced by https://github.com/sugarlabs/sugar/pull/560

PR #560, opened 2015-07-27
More robust i18n try4

godiard:
Replaces https://github.com/sugarlabs/sugar/pull/558 addressing comments from @quozl.

### DIFF
--- a/extensions/cpsection/language/model.py
+++ b/extensions/cpsection/language/model.py
@@ -20,7 +20,6 @@
 # Parts of the code were reused.
 #
 
-import logging
 import os
 import locale
 from gettext import gettext as _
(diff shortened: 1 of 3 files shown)


tchx84:
Aside from my previous comment, I see changes on top of other changes (e.g, write method is changed in 3 different patches), shouldn't that be consolidated?

godiard:
@tchx84 , do you prefer consolidate all the patches in a single patch?

tchx84:
@godiard no, just in a way where the same function doesn't get completely re-written 3 times :)

quozl:
Reviewed.  I'm fine with the separate patches, as they show the sequence of change.  Don't like consolidation, as it hides the sequence.  Future bug analysis depends on sequence of change.

tchx84:
Alright, we are tied now, so its up to @godiard.

godiard:
Ok, I merged this pr. Even when the method is modified in different patches, every patch have a specific  purpose, and I have tried to consolidate the changes, and the result is too big.

PR #561, opened 2015-07-29
Fix copying web activities entries in the journal

tchx84:
Some web activities do not write the "mime_type" entry
in their metadata. This field is used when copying entries
to external devices, while finding a proper extension.

Therefore, make this particular piece of code a bit more
robust against that missing field.

To reproduce:
1. Open a new instance of Gears activity, put some content and close.
2. Go to the Journal, select the activity entry, and copy it to a external device.
3. Check shell.log.


godiard:
@tchx84 what was the result of copy a journal entry without mime_type defined?

tchx84:
@godiard Without the patch, nothing, it does not get copied because of the error. With the patch, just the title, which seems correct.

tchx84:
Sent a new PR with more clear description https://github.com/sugarlabs/sugar/pull/562

PR #563, opened 2015-07-29
Fix Journal cannot copy entries without mime_type

tchx84:
Some web activities do not write the "mime_type" entry
in their metadata. This field is used when copying entries
to external devices, while finding a proper extension.

Therefore, make this particular piece of code a bit more
robust against that missing field.


tchx84:
This was already merged.

PR #564, opened 2015-08-01
Journal treeview fast!

samdroid-apps:
Merges and replaces #542 (with changes/more testing) and #520 

```
commit 50635f86571122a2cab9f0e8264c70bac2518686
Author: Sam Parkinson <sam.parkinson3@gmail.com>
Date:   Sat Aug 1 21:10:07 2015 +1000

    Refresh model in place for Journal ListView, fixes #4852
    
    Instead of replacing the ListModel, we now just change the result
    set.  This means that the buggy scroll position setting code is no
    longer needed.  Infact, a lot of code is removed.
    
    However, the model is still replaced if the query is changed.  If the
    query is changed, the length of the model could change and this would
    effect the scroll bars.  It is eaiser to change the model in this
    case, as we have no need to maintain the scroll position.

commit 3299d64e4b58413ee3e5460a9aea5a1ed6e54e8e
Author: Sam Parkinson <sam.parkinson3@gmail.com>
Date:   Thu May 28 07:50:53 2015 +1000

    Update metadata in the treeview without reload all - Fixes #4852
    
    Implement a set_value method in the model, and move the logic
    to write to the datastore to the ListModel. To be able to
    update the model without trigger a update in the treeview,
    we pass the updated callback in the ListModel.setup() method,
    and the callback is disconnected and connected in every update.
    
    The treeview have the responsibility to redraw the cell after
    the change was done.
    
    The listmodel stores a dictionary with the entries uodated,
    to provide that data until the next update in the Journal.
    
    This patch is heavily based in the work of
    Gonzalo Odiard <godiard@gmail.com>
```

### DIFF
--- a/src/jarabe/journal/listmodel.py
+++ b/src/jarabe/journal/listmodel.py
@@ -92,6 +92,10 @@ class ListModel(GObject.GObject, Gtk.TreeModel, Gtk.TreeDragSource):
         # avoid hitting D-Bus and disk.
         self.view_is_resizing = False
 
+        # Store the changes originated in the treeview so we do not need
+        # to regenerate the model and stuff up the scroll position
+        self._updated_entries = {}
+
         self._result_set.ready.connect(self.__result_set_ready_cb)
         self._result_set.progress.connect(self.__result_set_progress_cb)
 
@@ -107,8 +111,9 @@ class ListModel(GObject.GObject, Gtk.TreeModel, Gtk.TreeDragSource):
     def __result_set_progress_cb(self, **kwargs):
         self.emit('progress')
 
-    def setup(self):
+    def setup(self, updated_callback=None):
         self._result_set.setup()
+        self._updated_callback = updated_callback
 
     def stop(self):
         self._result_set.stop()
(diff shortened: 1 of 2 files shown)


tchx84:
@samdroid-apps can you clarify this: does the patch b19549d requires the previous a05776d patch?

samdroid-apps:
@tchx84 no, they are just merged together as @godiard requested

samdroid-apps:
I tested this patch just then.  I don't think we should merge it, mainly because of "Refresh model in place for Journal ListView, fixes #4852" (commit 50635f86571122a2cab9f0e8264c70bac2518686).

There was a reason the model is wholesale replaced when there is a big change.  The treeview does not deal with the model length getting shorter without any signals.  It is fine with changing the data, fine with adding new items, but as soon as you delete something it goes weird.

Testcase:

1.  Select 4 items in the journal
2.  Delete them
3.  Scroll down
4.  See 4 rows worth of damage

That commit is bad.  The other commit seems self contained and very good and much more self contained.  I apologise for merging them together.

godiard:
Don't apologise, I give yout that idea. Thanks for testing and find this issue.
As we have less time to do tests and less people testing, probably is a good idea merge only low risk changes, and merge big changes only when we are sure.

PR #565, opened 2015-08-07
Use Sugar palette menus for view source, not Gtk.Menu

samdroid-apps:
*Let's make sugar 108 the most paletteable release ever!*


godiard:
That was the reason because was so ugly :)

PR #571, opened 2015-09-15
Add setting for frame trigger size, fixes #4835

samdroid-apps:
When sugar is run inside a window, the frame trigger is a very target to hit.
This can be improved by making the trigger area bigger, however it is only
needed in some environments, not those where sugar is fullscreen.

Ticket URL  <https://bugs.sugarlabs.org/ticket/4835>

BTW:  Not 100% sure if this is a bug fix or a feature.  It's pretty small, but whatever :smile:

### DIFF
--- a/data/org.sugarlabs.gschema.xml
+++ b/data/org.sugarlabs.gschema.xml
@@ -152,6 +152,11 @@
             <summary>Corner Delay</summary>
             <description>Delay for the activation of the frame using the corners.</description>
         </key>
+        <key name="trigger-size" type="i">
+            <default>1</default>
+            <summary>Trigger Size</summary>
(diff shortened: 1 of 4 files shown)


quozl:
It is a feature that fixes a bug.

The feature will also assist in environments that are not virtualised.

Reviewed, like it.

quozl:
Thanks for the fix, I've cherry-picked it for an OLPC fork.

quozl:
Conflict resolved on OLPC fork; apply() method will need code added for handling trigger_size sid, in the same way as in f481ecdf8.

samdroid-apps:
Replaced by https://github.com/sugarlabs/sugar/pull/601 where the merge conflicts are resolved

PR #573, opened 2015-10-06

i5o:
I wasn't able to get the Duplicate option working, but the busy cursor is working..

I get (tested with Moon / TurtleArt activity)

``Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/jarabe/view/viewsource.py", line 397, in async_copy_activity_tree
    customizebundle.generate_bundle(nick, new_basename)
  File "/usr/lib/python2.7/site-packages/jarabe/view/customizebundle.py", line 84, in generate_bundle
    bundlebuilder.cmd_dist_xo(config, None)
  File "/usr/lib/python2.7/site-packages/sugar3/activity/bundlebuilder.py", line 354, in cmd_dist_xo
    packager = XOPackager(Builder(config, options.no_fail))
AttributeError: 'NoneType' object has no attribute 'no_fail'``

I think the duplicate option is totally broken?
Log: http://fpaste.org/275403/44146581/

### DIFF
--- a/src/jarabe/view/viewsource.py
+++ b/src/jarabe/view/viewsource.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2008 One Laptop Per Child
 # Copyright (C) 2009 Tomeu Vizoso, Simon Schampijer
 # Copyright (C) 2011 Walter Bender
-# Copyright (C) 2014 Ignacio Rodriguez
+# Copyright (C) 2014-15 Ignacio Rodriguez
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


godiard:
Do you have  updated toolkit? I think that erroe was fixed on https://github.com/sugarlabs/sugar-toolkit-gtk3/commit/100a6b56b65319999905fb5c91b5a1862308c208

i5o:
Image using the olpc os 13.2.5 so I assume that its the last version, I
will do a yum update and report

godiard:
@i5o, the change was done after 0.106, I don't know if is packaged.

samdroid-apps:
I don't get the same error but copy still crashes for me :)

If the copy crashes, the cursor should return to normal.  However, when I tested, it stayed spinning :(

@i5o, could you fix that?

i5o:
Sure, I will

PR #574, opened 2015-10-07

i5o:
Finally tested on sugar-build.

I get errors, but I think it's not because of this patch.
I need to look on the mechanism of duplication.


samdroid-apps:
Looks good to me, merged

PR #575, opened 2015-10-07
Ensure duplicated xo bundle is placed in expected directory

samdroid-apps:
This fixes a crash during the duplicate function in view source.

Steps to reproduce:

1. Open view source for an activity
2. Duplicate
3. See error in shell.log
4. See that there wasn't an xo bundle created in the journal

@i5o, does this fix your issue in sugar-build?

Note to self:  reopen https://github.com/samdroid-apps/sugar/commit/63348be2485a8b84f321fedc322715f3678f4399 after merge


i5o:
Thanks.

I've tested with few activities, and I don't get errors when duplicating

Merged

PR #576, opened 2015-10-07
Confirmation alert and Copying alert for Duplicate  in source view

i5o:
This patch adds a Confirmation alert when Duplicate (in view source)
    is clicked.
    Also adds a "Copying" alert, that is showed when the copy start
    until it finish.
    
    By Sam:
    Added alerts for when the Duplicate activity already exists
    and after the activity is duplicated
    
    Fixes #3672

### DIFF
--- a/src/jarabe/view/viewsource.py
+++ b/src/jarabe/view/viewsource.py
@@ -36,6 +36,7 @@ from gi.repository import Gio
 from sugar3.graphics import style
 from sugar3.graphics.icon import Icon
 from sugar3.graphics.xocolor import XoColor
+from sugar3.graphics.alert import NotifyAlert
 from sugar3.graphics.toolbutton import ToolButton
 from sugar3.graphics.palettemenu import PaletteMenuBox
 from sugar3.graphics.palettemenu import PaletteMenuItem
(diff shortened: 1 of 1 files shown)


src/jarabe/view/viewsource.py
```diff
@@ -400,12 +439,34 @@ def async_copy_activity_tree():
                                         new_basename),
                                     symlinks=True)
                     customizebundle.generate_bundle(nick, new_basename)
+
+                    if copy_alert:
+                        self.get_toplevel().remove_alert(copy_alert)
+
+                    alert = NotifyAlert(10)
+                    alert.props.title = _('Duplicated')
+                    alert.props.msg = _('The activity has been duplicated.')
```
samdroid-apps: I don't think full stops (periods, .s) are the convention for Alert messages.

src/jarabe/view/viewsource.py
```diff
@@ -371,6 +385,31 @@ def __init__(self, file_name, document_path, title, bundle=False):
...
+                GObject.idle_add(self.__copy_to_home_cb,
+                                 None,
+                                 new_alert)
+
+            new_alert = Alert()
+            new_alert.props.title = _("Duplicate activity")
+            new_alert.props.msg = _("Copying...")
+
+            self.get_toplevel().add_alert(new_alert)
+            GObject.idle_add(internal_callback, new_alert)
```
samdroid-apps: Why do we idle_add something to idle_add something else?

i5o: Ah.
That happens when you try to port another patch to here :laughing: 

It was:

+            def internal_callback():
+                self.emit('remove-alert', alert)
+                GObject.idle_add(self.__copy_to_home_cb, new_alert)

It should nt idle_add now. fixing it

samdroid-apps:
I will test...

i5o:
Thanks

samdroid-apps:
Merged

PR #577, opened 2015-10-07
Use a icon for when the battery is not present

i5o:
Now the colors goes along with the new battery icons.
Requires: https://github.com/sugarlabs/sugar-artwork/pull/73
Before:
Before the patch, no icon, just white

With patch
With the patch, new icon using xocolors

### DIFF
--- a/extensions/deviceicon/battery.py
+++ b/extensions/deviceicon/battery.py
@@ -28,7 +28,6 @@ from sugar3.graphics.icon import get_icon_state
 from sugar3.graphics.tray import TrayIcon
 from sugar3.graphics.palette import Palette
 from sugar3.graphics.palettemenu import PaletteMenuBox
-from sugar3.graphics.xocolor import XoColor
 
 from jarabe.frame.frameinvoker import FrameWidgetInvoker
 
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Hum, this doesn't make it very obvious (in my opinion) that you don't have a battery in your computer.  Is that what we want though?

Also, when the speakers are muted they go all white.  Should speakers muting be analogous to the battery not being there?

quozl:
Design questions can be tagged [DESIGN] and posted to sugar-devel@?

My opinion is that the battery icon should disappear when the battery is removed.  There's no need to say it isn't there.

Speakers aren't removed, so an icon should remain.  Of course, on some systems, speakers are removable, but that's not common enough yet to spend time on.  Sugar doesn't have support for additional Bluetooth or USB speakers.

My other job is having to explain your changes to users in release notes of next release of an operating system that has Sugar.  That's why [DESIGN] discussions are useful to me.

samdroid-apps:
+1 for the idea of removing the icon.  Thats... too logical

i5o:
Yes, removing the icon looks more logical for me, and I think for users too So I will close, and will start a DESIGN thread for the deviceicon removal when battery is not present

PR #578, opened 2015-10-09
Clear journal search with escape key

nemesiscodex:
As mentioned in the issue #4892

> search entry box is cleared with escape key in
* neighbourhood view,
* friends view, and;
* home view (list and circle),
>
but not in journal.
when a journal entry is expanded, the escape key returns to the main journal view. (_key_press_event_cb in journalactivity.py)
please also clear the search entry. ;-)

Regards

### DIFF
--- a/src/jarabe/journal/journalactivity.py
+++ b/src/jarabe/journal/journalactivity.py
@@ -287,6 +287,7 @@ class JournalActivity(JournalWindow):
 
         keyname = Gdk.keyval_name(event.keyval)
         if keyname == 'Escape':
+            self._main_toolbox.clear_query()
             self.show_main_view()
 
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Looks good code wise, haven't tested.

Could you change the commit message a little?  If you put "fixes #4892" first line, instead of on the 2nd line?

nemesiscodex:
Like that? :sweat_smile:

samdroid-apps:
Merged!  I would usually put the fixes #ABCD at the end, but it doesn't matter too much

nemesiscodex:
I will keep that in mind for the next one :smile: thanks!

quozl:
added pr to ticket

PR #579, opened 2015-10-09
Do not crash if brightness manager can not get a path

samdroid-apps:
If the brightness manager does not get authentication from polkit,
it does not get a path.  This causes a crash because no file
mointor is created, causing sugar to crash when calling a method
of None.

### DIFF
--- a/src/jarabe/model/brightness.py
+++ b/src/jarabe/model/brightness.py
@@ -116,7 +116,8 @@ class Brightness(GObject.GObject):
 
     def set_brightness(self, value):
         # do not monitor the external change we are about to trigger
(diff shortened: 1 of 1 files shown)


samdroid-apps:
This does hide the brightness stuff and only leaves the screenshot in the display device icon.

When you explain the patch like that, it is probably a bad idea because we shouldn't aim to support every weird install that I can create.

tchx84:
That patch does not hides it, it simply avoids trying to set values, and yeah, this is probably an "unnatural" scenario, but still, _we_ use it, so it might be worth covering it...

PR #580, opened 2015-10-09
Hide battery icon when battery is not present

i5o:
Right now it shows a full white battery when it's not present
more logical would be if its not showed.

This patch does that

### DIFF
--- a/extensions/deviceicon/battery.py
+++ b/extensions/deviceicon/battery.py
@@ -28,7 +28,6 @@ from sugar3.graphics.icon import get_icon_state
 from sugar3.graphics.tray import TrayIcon
 from sugar3.graphics.palette import Palette
 from sugar3.graphics.palettemenu import PaletteMenuBox
-from sugar3.graphics.xocolor import XoColor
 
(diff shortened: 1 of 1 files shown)


samdroid-apps:
I will test this

samdroid-apps:
Doesn't cause any issues when the battery is not removed.

Lemme test what happens when you remove the battery

samdroid-apps:
Will not merge.  It doesn't work.

If I take out my battery while sugar is running, it doesn't take the battery out of the frame.

i5o:
Are you using it on a Xo or Sugar-Build?

It works for me in a XO

samdroid-apps:
Further testing showed that the battery removal code does not work on sugar-build before or after this patch.  That's funny, because it seems to have access to the battery fine (it shows the %)

samdroid-apps:
Works really nicely on XO4.  It even animates when you remove that battery!

Please refrain from adding "This patch does that" to the end of your commit message for next time.  I should have seen that before I merged the patch, but whatever.

PR #581, opened 2015-10-09
Erasing and Erased alert for 'Erase' option in List View

i5o:
This add an Erasing alert while the activity is being erased
also adds a "Erased" notifyalert for when the erase process is done.

The idea is to give feedback to user, not just to update the listview.

### DIFF
--- a/src/jarabe/desktop/activitieslist.py
+++ b/src/jarabe/desktop/activitieslist.py
@@ -33,6 +33,7 @@ from sugar3.graphics import style
 from sugar3.graphics.icon import Icon, CellRendererIcon
 from sugar3.graphics.xocolor import XoColor
 from sugar3.graphics.alert import Alert
+from sugar3.graphics.alert import NotifyAlert
 from sugar3.graphics.palettemenu import PaletteMenuItem
 from sugar3.graphics.scrollingdetector import ScrollingDetector
 from sugar3.graphics.palettewindow import TreeViewInvoker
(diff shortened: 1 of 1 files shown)


samdroid-apps:
What is the rational behind this patch?

Removing an activity is very fast, so the user is probably still going to be at the list view when the activity is removed.

Since it is so fast, do we really need an alert to say it is doing something?

i5o:
Maybe it could be useful with old hardware.
I tested with SimpleWikipedia activity and it tooks few seconds (XO-4)

I will test on XO-1 and reply

2015-10-09 19:16 GMT-02:00, Sam <notifications@github.com>:
> What is the rational behind this patch?
>
> Removing an activity is very fast, so the user is probably still going to be
> at the list view when the activity is removed.
>
> Since it is so fast, do we really need an alert to say it is doing
> something?
>
> ---
> https://github.com/sugarlabs/sugar/pull/581#issuecomment-146987930


-- 
Ignacio Rodríguez
SugarLabs at Facebook

i5o:
it tooks few more seconds on XO-1, I tried to remove Measure activity and Scratch..

Maybe I should test with Wikipedia..
So maybe instead of "Erasing" alert, we just could set the cursor to watch while erasing, and turn it to normal after it

samdroid-apps:
Yeah, sounds good

i5o:
Doing it

PR #582, opened 2015-10-09
Check if remove_alert is present in toplevel

i5o:
Sometimes the toplevel is changed to DocumentButton,
and there is no remove_alert on DocumentButton.

The idea is to check if there is remove_alert on toplevel,
to prevent a false function call

I made this patch because of this error in logs after duplicate activity (appears -I think- 10 seconds after the duplication process)

`
  File "/home/broot/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/view/viewsource.py", line 470, in __alert_response_cb
    self.get_toplevel().remove_alert(alert)
AttributeError: 'DocumentButton' object has no attribute 'remove_alert'
`

### DIFF
--- a/src/jarabe/view/viewsource.py
+++ b/src/jarabe/view/viewsource.py
@@ -467,7 +467,9 @@ class DocumentButton(RadioToolButton):
             self.get_toplevel().add_alert(alert)
 
     def __alert_response_cb(self, alert, response_id):
-        self.get_toplevel().remove_alert(alert)
+        toplevel = self.get_toplevel()
+        if hasattr(toplevel, 'remove_alert'):
+            toplevel.remove_alert(alert)
 
     def __keep_in_journal_cb(self, menu_item):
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Does the alert get removed automatically?

i5o:
Nop,

if I don't call remove_alert it's not automatical

2015-10-09 19:14 GMT-02:00, Sam <notifications@github.com>:
> Does the alert get removed automatically?
>
>
> ---
> https://github.com/sugarlabs/sugar/pull/582#issuecomment-146987634


-- 
Ignacio Rodríguez
SugarLabs at Facebook

samdroid-apps:
I've made a ticket for this bug:  https://bugs.sugarlabs.org/ticket/4898

samdroid-apps:
I think we should wait until we understand what this bug is before we patch it.  Them maybe we can make an even better patch!

PR #585, opened 2015-10-12
Update README.md

FGrose:
Fix typos and avoid a numbers controversy.
Provide some web links (for the GitHub front page).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,17 +2,20 @@ Sugar
 =====
 
 Sugar is the core component of a worldwide effort to provide every
-child with equal opportunity for a quality education. Available in
+child with an equal opportunity for a quality education. Available in
 more than twenty-five languages, Sugar’s Activities are used every
-school day by nearly three million children in more than forty
-countries.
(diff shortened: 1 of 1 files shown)


quozl:
Reviewed.  I suggest making both links look similar in style, the way you have them at the moment one has https prefix the other does not.

samdroid-apps:
Closing.  We need the links in the same style please.  Actually maybe we could even have text labels instead of urls.

quozl:
@FGrose, in case it isn't obvious, this project has a policy of never re-working submitted changes that fail review, but rather it gives the opportunity to the original contributor to work with feedback and submit again.  If you've no time to do that, say so, and one of us will take over.  Until then we may consider the area of change off-limits because you've expressed interest.  Thanks for your contribution.

FGrose:
ok, good to know.  See https://github.com/sugarlabs/sugar/pull/591.
I'm happy if a maintainer wants to amend the commit.

PR #586, opened 2015-10-13
add logout failure alert, for review

quozl:
DO NOT MERGE

Issues are:

 - alert is lost if F1 then F3 key is pressed, seems to be a problem generally with alerts on the home window,

 - control panel restart is not yet handled

 - logout shortcut key is not yet handled, but it is developer only feature.

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -399,9 +399,40 @@ class ControlPanel(Gtk.Window):
                 self._section_view.restart_alerts
             self._show_main_view()
         elif response_id is Gtk.ResponseType.APPLY:
-            session_manager = get_session_manager()
-            session_manager.logout()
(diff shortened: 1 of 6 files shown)


tchx84:
@quozl which sugar version are you using? Can you rebase? I can't get it to work, and can't be 100% sure is not related to the resolved conflicts.

quozl:
Sorry, it was on an old checkout.  Now rebased.  Not tested.  The area of conflict is unrelated to the problem.

Steps to reproduce are on the mailing list design post day before yestre; simplest method is to start Terminal, then start xclock or xterm, then click on the Terminal and press Ctrl-Z, this sends a SIGSTOP to the X app child.  Now try to logout.

Not tested in Sugar Build; not applicable.

quozl:
Replaced by https://github.com/sugarlabs/sugar/pull/587 and https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/261.

PR #587, opened 2015-10-14
add logout failure alert, for review

quozl:
DO NOT MERGE

Issues are:

 - alert is lost if F1 then F3 key is pressed, seems to be a problem generally with alerts on the home window, advice is welcome,

 - logout shortcut key is not yet handled, but it is developer only feature.

To cause a logout failure alert, start Terminal, then start xclock or xterm, then click on the Terminal and press Ctrl-Z, this sends a SIGSTOP to the child process. Now switch to home view or control panel and try to logout, restart, or shutdown.

Other test cases:

 - check there is a busy cursor during startup and it goes away once journal is ready,

 - check there is a busy cursor during shutdown, restart, or logout, if there is enough time to see it,

 - make an activity not respond, check the logout failure alert appears,

 - click cancel on the alert, switch to the hung activity, unhang it, resume working; the system shall not quit,

 - click cancel on the alert, switch to the hung activity, switch to home view, try logout again, click cancel, switch to hung activity, unhang it, resume working; the system shall not quit,

 - click continue on the alert; the system shall quit,

 - allow the timer to expire; the system shall quit.

Replaces https://github.com/sugarlabs/sugar/pull/586

Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/261

References http://lists.sugarlabs.org/archive/sugar-devel/2015-October/050877.html

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -399,9 +399,40 @@ class ControlPanel(Gtk.Window):
                 self._section_view.restart_alerts
             self._show_main_view()
         elif response_id is Gtk.ResponseType.APPLY:
-            session_manager = get_session_manager()
-            session_manager.logout()
-            # FIXME: deal more cleanly with no response to save yourself
+            self.busy()
+            self._section_toolbar.accept_button.set_sensitive(False)
+            self._section_toolbar.cancel_button.set_sensitive(False)
+            get_session_manager().logout()
+            GObject.timeout_add_seconds(4, self.__quit_timeout_cb)
+
+    def __quit_timeout_cb(self):
+        self.unbusy()
+        alert = Alert()
+        alert.props.title = _('Warning')
+        alert.props.msg = _('An activity is not responding.')
+
+        icon = Icon(icon_name='dialog-ok')
+        alert.add_button(Gtk.ResponseType.ACCEPT, _('Lose unsaved work'), icon)
+        icon.show()
+
+        icon = Icon(icon_name='dialog-cancel')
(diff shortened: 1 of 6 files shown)


samdroid-apps:
> alert is lost if F1 then F3 key is pressed, seems to be a problem generally with alerts on the home window, advice is welcome,

I didn't look at the code, but I speculate:  alerts are attached to canvases, and each zoom view is a different canvas

> logout shortcut key is not yet handled, but it is developer only feature.

Why doesn't this just throw fire on everything and kill it anyway?

> To cause a logout failure alert, start Terminal, then start xclock or xterm, then click on the Terminal and press Ctrl-Z, this sends a SIGSTOP to the child process. Now switch to home view or control panel and try to logout, restart, or shutdown.

Can reproduce, your patch does fix it.

However, the timer overflows the label:

[https://gyazo.com/79c27262c3548b15f3aa6c7c89868e90](https://gyazo.com/79c27262c3548b15f3aa6c7c89868e90)

> check there is a busy cursor during startup and it goes away once journal is ready,

This happens without the patch for me

quozl:
> I didn't look at the code, but I speculate: alerts are attached to canvases, and each zoom view is a different canvas

Fixed since, not yet pushed, turns out HomeWindow alerts are unused and buggy.

> Why doesn't this just throw fire on everything and kill it anyway?

Noted, thanks.

> However, the timer overflows the label:

You want 9 second timeout?  I can do that.

quozl:
Replaced by https://github.com/sugarlabs/sugar/pull/592

quozl:
Reviewed comments, all replied to or carried over to next pull request.

PR #588, opened 2015-10-15
check before remove on control panel section end

quozl:
A critical warning appears in shell.log when a control panel section is removed from screen:

```(main.py:6713): Gtk-CRITICAL **: gtk_container_remove: assertion 'gtk_widget_get_parent (widget) == GTK_WIDGET (container) || GTK_IS_ASSISTANT (container)' failed```

Fix the warning by checking the object is in the container before trying to remove it.

Reproducer: choose My Settings, then any icon, such as About Me, then press ```ESCAPE``` key once.  Examine shell.log.

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -111,7 +111,8 @@ class ControlPanel(Gtk.Window):
 
     def _set_canvas(self, canvas):
         if self._canvas:
-            self._main_view.remove(self._canvas)
+            if self._canvas in self._main_view:
+                self._main_view.remove(self._canvas)
         if canvas:
             self._main_view.add(canvas)
         self._canvas = canvas

### CODE REVIEW

File: src/jarabe/controlpanel/gui.py [hunk 1]
```diff
@@ -111,7 +111,8 @@ def _calculate_max_columns(self):
 
     def _set_canvas(self, canvas):
         if self._canvas:
-            self._main_view.remove(self._canvas)
+            if self._canvas in self._main_view:
```
samdroid-apps: Can you combine the 2 ifs

samdroid-apps:
Looks good after testing.

But what does not removing it cause?  Is it removed in another way?  Or are we leaking widgets?

quozl:
Thanks for the review and testing.

> Can you combine the 2 ifs

Yes, that's possible.  Might also remove the first ```if``` because ```None``` will never be in the ```_main_view``` children.  But as it stands, it explains the possibility to the reader, and the reader is the one that will have the most trouble understanding.  The compiler won't have any trouble.

> But what does not removing it cause?

An error.  I/O to disk for logging.  Additional delay during that I/O.

> Is it removed in another way? Or are we leaking widgets?

Yes, it was already removed.  No, we're not leaking widgets.  Once all references to children widgets have gone away, they will be deleted.

quozl:
Not sure why you have closed this pull request, could you please make it more obvious?  I'll have to delay another pull request until I understand.

samdroid-apps:
Sorry, I didn't see that you forced pushed the changes as per my comments.  Github does not post a message relating to force pushes, so please make a comment.

samdroid-apps:
Sorry again, you didn't make any force pushes or changes.

My comments were that you should combine the 2 if statements and sent a new patch.  Since you have answered my other questions, I'll merge it once you do that

quozl:
Replaced by https://github.com/sugarlabs/sugar/pull/593

>  I didn't see that you forced pushed the changes as per my comments.

Despite a push with --force, GitHub does not offer a re-open here, so I'm having to open a new one and cross linking it with this one.  My apologies to other reviewers.

PR #590, opened 2015-10-15
Use sugar palettes for activity invites rather than GtkMenu palettes

samdroid-apps:
I might also look into putting the name of the inviter or the name of the activity instance in the menu, but that would require model changes.


quozl:
Commit message doesn't say why this change is being made.  It presumes knowledge about the difference between MenuItem and PaletteMenuItem.  The ignorant, such as myself, would have to go and research.  I think we need a commit history that explains why a change was made.

PR #592, opened 2015-10-16
fixes for logout hang

quozl:
A series of fixes for failure to logout, restart or shutdown, when an activity is hung.  See the individual commits for narrative detail.

To cause a logout failure alert, start Terminal, then start ```xclock``` or ```xterm```, then click on the Terminal and press Ctrl-Z, this sends a SIGSTOP to the child process. Now switch to home view or control panel and try to logout, restart, or shutdown.

Other methods are to send SIGSTOP to any activity process.

Other test cases:

 - check there is a busy cursor during startup and it goes away once journal is ready,

 - check there is a busy cursor during shutdown, restart, or logout, if there is enough time to see it,

 - make an activity not respond, check the logout failure alert appears,

 - click cancel on the alert, switch to the hung activity, unhang it, resume working; the system shall not quit,

 - click cancel on the alert, switch to the hung activity, switch to home view, try logout again, click cancel, switch to hung activity, unhang it, resume working; the system shall not quit,

 - click continue on the alert; the system shall quit,

 - allow the timer to expire; the system shall quit.

Replaces https://github.com/sugarlabs/sugar/pull/587

Requires https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/263 and https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/266.

References http://lists.sugarlabs.org/archive/sugar-devel/2015-October/050877.html

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -24,7 +24,7 @@ from gi.repository import Gdk
 
 from sugar3.graphics.icon import Icon
 from sugar3.graphics import style
-from sugar3.graphics.alert import Alert
+from sugar3.graphics.alert import Alert, TimeoutAlert
 
 from jarabe.model.session import get_session_manager
 from jarabe.controlpanel.toolbar import MainToolbar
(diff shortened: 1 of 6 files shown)


samdroid-apps:
This looks like a really good patch set.  I have code reviewed it.

* Did you finish writing the message of commit d5296ef805ca0ded2b26bd46283065c04e2c8cc8?
* Different time out in the control panel and in the home view (4 sec vs 3 sec)

Otherwise it looks really good.

I'm not 100% sure on the alert implementation.  I think that we are suffering from the fact that there is no way to push a critical alert to users in sugar and having to implement the alert in 100 places.  There was the same issue with the max running activities patch from memory - it only displays alerts in the journal and home view.

The previously proposed solution was to show the alerts in activities by adding a new dbus api or something.  But this did not get merged because it added a new api and that wasn't well received.

This is something we seriously need a design discussion about, but the design discussion is separate from this patch, which does everything consistently!

samdroid-apps:
>  * Different time out in the control panel and in the home view (4 sec vs 3 sec)

Just to add a note - this should be a constant not a magic number too

samdroid-apps:
And the 30sec timeout should probably have a constant as well,  that's all :smile:

quozl:
Thanks for the review.

> Did you finish writing the message of commit d5296ef?

You're right, there's a missing 'o.' on the end of the first paragraph, and a missing "." on the end of the last, but it is trivial, so please wait for others to review and me to agree before closing this pull request.  I'd like to give everyone a chance to review.

> Different time out in the control panel and in the home view (4 sec vs 3 sec)

Yes, because the control panel is invoked from the home view, and on an XO-1 the control panel steals more memory, causing any activities to take longer to respond to save-yourself message, since they will have to page fault to bring contents of virtual memory into physical memory.

I don't think this needs to be a comment or commit detail though.

> should be a constant not a magic number too

I disagree, and would like other developers to give an opinion.  @tchx84, @godiard?  The problem with constants with only single point of use is that it takes longer for someone new to read the code and locate the constant declaration.  I'm fine with constants that have multiple uses and are near to the code that uses them, but otherwise they waste valuable developer time.  We want to be welcoming to new developers.

> I'm not 100% sure on the alert implementation.

As you will have seen from your review, the implementation of alerts in Sugar is inconsistent.  But beyond fixing the HomeWindow alert for this patch series, it didn't seem like a good time to embark on such a project.

I was happy to try to make the busy cursor more consistent, following the GSoC or GCI contributions, and this series does achieve some of that.  But there are some more opportunities there as well.

quozl:
@samdroid-apps, looks like @tchx84, @godiard, and others are busy.  Will you merge the patch if I redo it with constants, or are my arguments above persuasive?  If it needs to be redone, exactly which constants do you want moved away from where they are used, and what is the reason for the move?

samdroid-apps:
Na looks fine.  I'll merge it soon

quozl:
thanks.

quozl:
I've cherry-picked this for an OLPC fork.

PR #594, opened 2015-10-16
Use sugar palettes for activity invites rather than GtkMenu palettes

samdroid-apps:
The GtkMenu palettes are not reccomended for simple palettes as
they do not look good as the sugar palettes.  This is not an
issue that will be addresses with the GtkMenu, so consumers must
be changed.

### DIFF
--- a/src/jarabe/frame/activitiestray.py
+++ b/src/jarabe/frame/activitiestray.py
@@ -32,7 +32,6 @@ from sugar3.graphics.radiotoolbutton import RadioToolButton
 from sugar3.graphics.toolbutton import ToolButton
 from sugar3.graphics.icon import Icon, get_icon_file_name
 from sugar3.graphics.palette import Palette
-from sugar3.graphics.menuitem import MenuItem
 from sugar3.graphics.palettemenu import PaletteMenuBox
 from sugar3.graphics.palettemenu import PaletteMenuItem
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Replaces #594.  No changes besides the commit message

quozl:
Replaces #594?  Think you mean https://github.com/sugarlabs/sugar/pull/590.

Who recommends?

What does it look like before and after?

samdroid-apps:
**Before**
[https://gyazo.com/65b50bb41c9d5487006044998209a453](https://gyazo.com/65b50bb41c9d5487006044998209a453)

**After**
[https://gyazo.com/55696ab3dab7e14bae881197ab24a6ec](https://gyazo.com/55696ab3dab7e14bae881197ab24a6ec)

Yes, it does replace #590

> Who recommends?

I reccomend it.  @godiard said on the ml a while ago:

> We don't want invest more time in the Gtk.Menu palettes,
> but would be great find a fix to ship Sugar in F22 on shape.

quozl:
Oh, okay, so 

> The GtkMenu palettes are not reccomended for simple palettes as they do not look good as the sugar palettes.

Means the invite join menu uses icon sizes, drop shadow, border and spacing which are inconsistent with other Sugar menus.

Now I understand.

Reviewed.

tchx84:
Tested and Merged.

PR #595, opened 2015-10-24
sugar.in:  Fix typos in the debug file here document.

FGrose:
Remove stray single quote.
End sentence with a period.


samdroid-apps:
Good fix.  Annoying bug!

PR #596, opened 2015-10-29
serial number helper

quozl:
On commodity x86 and x86_64 hardware the serial number is available to
root in /sys/class/dmi/id/product_serial, but Sugar does not run as
root, so per commit 612842c the string "Not available" is shown.

Add a polkit helper to read the serial number.

### DIFF
--- a/bin/sugar-serial-number-helper
+++ b/bin/sugar-serial-number-helper
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+device=/sys/class/dmi/id
+
+read value < $device/product_serial
+echo "$value"
--- a/extensions/cpsection/aboutcomputer/model.py
+++ b/extensions/cpsection/aboutcomputer/model.py
@@ -23,6 +23,7 @@ import errno
 import time
 
 from gi.repository import Gio
+from gi.repository import GLib
 
 from jarabe import config
 from jarabe.model.network import get_wireless_interfaces
@@ -35,6 +36,7 @@ _DMI_DIRECTORY = '/sys/class/dmi/id'
 _logger = logging.getLogger('ControlPanel - AboutComputer')
 _not_available = _('Not available')
 
+_serial_no = None
 
 def get_aboutcomputer():
     msg = 'Serial Number: %s \nBuild Number: %s \nFirmware Number: %s \n' \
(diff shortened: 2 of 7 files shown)


bin/sugar-serial-number-helper
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+device=/sys/class/dmi/id
+
+read value < $device/product_serial
+echo "$value"
```
samdroid-apps: Could this use cat an be more simple?  Do we need the device variable?

extensions/cpsection/aboutcomputer/model.py
```diff
@@ -46,11 +48,38 @@ def print_aboutcomputer():
...
+
+    def _helper():
+        binary = 'sugar-serial-number-helper'
+        for path in os.environ['PATH'].split(os.pathsep):
+            binary_path = os.path.join(path, binary)
+            if os.path.exists(binary_path):
+                return binary_path
+        return None
+
+    helper = _helper()
```
samdroid-apps: I don't think closures are in favour around here.  Can't it just be a loop with a `break` statement?

extensions/cpsection/aboutcomputer/model.py
```diff
@@ -46,11 +48,38 @@ def print_aboutcomputer():
...
+    if helper is None:
+        return _not_available
+
+    cmd = 'pkexec %s' % helper
```
samdroid-apps: pkexec seems to automatically resolve the full path of the executable.  Maybe it is better to let pkexec do it instead of doing the exact same thing ourselves except probably not dealing with as many edge cases?

samdroid-apps: eg. to test run `pkexec sugar-backlight-helper --get-max-brightness`.  Works fine for me (fc24 rawhide).

samdroid-apps:
Looks great, just a few little things.

bin/sugar-serial-number-helper
quozl: Using ```cat``` would add delay while another binary is started as a subprocess of the shell.  ```read``` is a built-in for the shell.  The device variable follows the style of ```sugar-backlight-helper```.

extensions/cpsection/aboutcomputer/model.py
quozl: Again, this copied the style set by the backlight helper.

extensions/cpsection/aboutcomputer/model.py
quozl: Good idea, but again, I followed the existing style.  Would you like to patch the backlight helper?

quozl:
Thanks for the review.  Please make it clearer which little things are questions for clarification, and which you think need to be fixed before merge.

PR #597, opened 2015-10-30
Recycle paths during journal copying, fixes #4876

samdroid-apps:
The frame requests the drag data form the journal twice, meaning
that the path it recives first is deleted.  This commit makes
the clipboard reuse paths if they are reuqested sequentially, as
the frame does.

Ticket <https://bugs.sugarlabs.org/ticket/4876>


tchx84:
Tested and Merged!

PR #598, opened 2015-10-31
Add zoom guesture for desktop views

samdroid-apps:
Centeral the the nvaigation in Sugar is the zoom metaphor.  This
metaphor does not extend to the zoom guesture on multi touch
devices.

This commit adds a zoom guesture controller to the homewindow.
The user can then use the pinch to zoom guesture to switch between
the home, group and mesh views.

This was tested on commodity hardware using fc24 rawhide.

### DIFF
--- a/src/jarabe/desktop/homewindow.py
+++ b/src/jarabe/desktop/homewindow.py
@@ -21,6 +21,7 @@ from gi.repository import GLib
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import GdkX11
+from gi.repository import SugarGestures
 
 from sugar3.graphics import style
 from sugar3.graphics import palettegroup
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Does nothing on an xo4 running olpc os 14.

quozl:
> Does nothing on an xo4 running olpc os 14.

What is olpc os 14?

samdroid-apps:
The fc20 build.  Is that 14?

quozl:
Oh, right, you mean 14.1.0, not yet released, and has blocker bugs that haven't had progress for many months.  I was afraid you might have been referring to a deployment build or the Sugar Labs testing build 41002SL.

i5o:
"ping" for the PR, are we going to push this?

quozl:
I'm still puzzled as to why it does nothing on an OLPC XO-4 running 14.1.0, according to Sam.  I've not tested it.  Does it work on an OLPC XO-4 running 13.2.6?  That's probably the most stable touchscreen platform Sugar is run on at the moment.  Hard to test if it doesn't work.

samdroid-apps:
I don't currently have access to an XO4 with 13.2.6.  You are welcome to
test.

PR #601, opened 2015-10-31
Add setting for frame trigger size, fixes #4835

samdroid-apps:
Replaces #571 

When sugar is run inside a window, the frame trigger is a very target to hit.
This can be improved by making the trigger area bigger, however it is only
needed in some environments, not those where sugar is fullscreen.

Ticket URL https://bugs.sugarlabs.org/ticket/4835


samdroid-apps:
Yeah, I'll fix that

PR #603, opened 2015-11-03
fix missing scroll bar in view source

quozl:
On Fedora 18, scroll bars were missing, see [4903](https://bugs.sugarlabs.org/ticket/4903).  On Fedora 20, impossible arguments were given to X server, resulting in desktop instability, see [12872](http://dev.laptop.org/ticket/12872).

Regression introduced in 4be3f9e and [4273](https://bugs.sugarlabs.org/ticket/4273), in the SourceDisplay class, where a [GtkBox](https://developer.gnome.org/gtk3/stable/GtkBox.html) is inserted between the [GtkScrolledWindow](https://developer.gnome.org/gtk3/stable/GtkScrolledWindow.html) and any child.

Fix is

- remove GtkBox intermediary; unnecessary,

- add methods for managing a single child of the GtkScrolledWindow, which will be either a [GtkSourceView](https://developer.gnome.org/gtksourceview/stable/GtkSourceView.html) or a [GtkEventBox](https://developer.gnome.org/gtk3/stable/GtkEventBox.html) in a [GtkViewport](https://developer.gnome.org/gtk3/stable/GtkViewport.html),

- remove some [GtkWidget.show](https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-show) calls; made redundant by [.show_all](https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-show-all),

### DIFF
--- a/src/jarabe/view/viewsource.py
+++ b/src/jarabe/view/viewsource.py
@@ -706,17 +706,22 @@ class SourceDisplay(Gtk.ScrolledWindow):
         self.props.hscrollbar_policy = Gtk.PolicyType.AUTOMATIC
         self.props.vscrollbar_policy = Gtk.PolicyType.AUTOMATIC
 
-        self._box = Gtk.Box()
-        self.add(self._box)
-        self._box.show()
-
         self._file_path = None
+
+    def _replace(self, child):
+        self._remove_children()
+        self.add(child)
+
+    def _replace_with_viewport(self, child):
+        self._remove_children()
+        self.add_with_viewport(child)
+
+    def _remove_children(self):
+        for child in self.get_children():
+            self.remove(child)
 
     def _set_file_path(self, file_path):
         self._file_path = file_path
-
-        for child in self._box.get_children():
-            self._box.remove(child)
 
         if self._file_path is None:
             self._show_no_file()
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Still results in X crashing for me on fc20.  Happens after it loads the source, but same as before (it loads the code, then it loads more signalled by the scroll bar getting smaller, then crash).  Tested on the evil moon activity that caused a crash before.

I'm not sure if this patch was meant to fix that, maybe don't include it in the commit message if it wasn't?

Otherwise, it doesn't break sugar on fc24 and looks good to me code wise.

quozl:
Interesting, thanks for the test.  I've tested again just now on Fedora 20, and can't fault it.  With Moon, Memorize, Measure and Maze.

Can you please confirm you have a file /usr/lib/python2.7/site-packages/jarabe/view/viewsource.py with md5sum 3585f26e8afdf99f6ed718b0a597c77d?

Can you please save "rpm -qa | sort" into a gist or pastebin and link here?  My test system package list is [here](http://dev.laptop.org/~quozl/z/1Ztro3.txt).

quozl:
Rebased.

samdroid-apps:
Works great on fc20 XO4.  Merging

PR #604, opened 2015-11-03
sugar.in: fix shell posix and quoting

quozl:
Pushing upstream a downstream patch that has general applicability.

Borrowed from
https://anonscm.debian.org/cgit/pkg-sugar/sugar.git/

Signed-off-by: James Cameron <quozl@laptop.org>

### DIFF
--- a/bin/sugar.in
+++ b/bin/sugar.in
@@ -2,7 +2,7 @@
 
 sugardatadir=@prefix@/share/sugar/data
 
-if [ "$(id -u)" -eq 0 -o "$(id -ru)" -eq 0 ] ; then
+if [ "$(id -u)" -eq 0 ] || [ "$(id -ru)" -eq 0 ] ; then
 	echo Refusing to run as root.
 	exit 3
 fi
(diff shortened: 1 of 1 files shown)


samdroid-apps:
GitHub suggests the authors GitHub account name, so ping @jonassmedegaard, you'll probably want to drop this patch downstream.

jonassmedegaard:
I don't understand how your sentence above, samdroid-apps, is related to (I guess) the fix_shell_posix_and_quoting.patch applied in Debian: There are no "Github account name" in that patch.

samdroid-apps:
GitHub seemed to say the patch was made by you, and that you had the username jonassmedegaard.

You might want to drop the patch downstream now that it is merged upstream.

Apologies for the bad phrasing.

quozl:
@samdroid-apps, nothing be done until a new upstream release is being packaged.

@jonassmedegaard, I had carefully attributed author during commit.  Sam noticed the correlation.  But Sam probably doesn't know that a Debian developer would _certainly_ notice this once the new upstream release was made, because the previous patch wouldn't apply.  :smile_cat:

jonassmedegaard:
Quoting Sam (2015-11-04 10:14:30)
> GitHub seemed to say the patch was made by you, and that you had the 
> username jonassmedegaard.

That's right :-)


> You might want to drop the patch downstream now that it is merged 
> upstream.
> 
> Apologies for the bad phrasing.

Ahhh.  Makes perfect sense now :-D

 - Jonas

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

PR #605, opened 2015-11-04
serial number helper

quozl:
On commodity x86 and x86_64 hardware the serial number is available to
root in /sys/class/dmi/id/product_serial, but Sugar does not run as
root, so per commit 612842c the string "Not available" is shown.

Add a polkit helper to read the serial number.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ data/sugar.xml
 data/sugar-xo.gtkrc
 data/sugar-emulator.desktop
 data/org.sugar.brightness.policy
+data/org.sugar.serial-number.policy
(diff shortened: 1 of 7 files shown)


samdroid-apps:
This is WIP right?

quozl:
Thanks for discussion.

> This is WIP right?

Exactly what work are you proposing is in progress?  See #596, where you closed the previous to stop discussion.

samdroid-apps:
No sorry, I'm just having issues reading today

Sorry, I thought there were more commits

quozl:
Rebased.

quozl:
Please remove the "needs work" label.

samdroid-apps:
Worked great when I ran without any polkit-password-getter and showed Not Available as expected.

Polkit-gnome is really buggy in rawhide, so I can not test that at this time.  But it worked before :smile:

PR #606, opened 2015-11-04
add accelerator for control panel open

quozl:


samdroid-apps:
Why?

Why alt-shift-m of all keys?  Why is m for controlpanel?

quozl:
Thanks for discussion.

> Why?

Because of [#4660](http://bugs.sugarlabs.org/ticket/4660).  See #497, where this was first proposed, and you closed it to stop discussion.

> Why alt-shift-m of all keys? Why is m for controlpanel?

"[M]y Settings".  Because nobody came up with a better key.

quozl:
Rebased.

samdroid-apps:
That makes a ton of sense now

PR #608, opened 2015-11-04
sugar-install-bundle: multiple bundles

quozl:
Accept multiple bundles on the command line.

Was requested in 2011 by Thomas Gilliard, a solution was
proposed by me but was accepted ada284445e9a3ce96a0c9df4c299021254b0668d
without complete testing, and caused http://dev.laptop.org/ticket/11178
so it was reverted in a hurry.

### DIFF
--- a/bin/sugar-install-bundle
+++ b/bin/sugar-install-bundle
@@ -9,19 +9,19 @@ DBusGMainLoop(set_as_default=True)
 
 
 def cmd_help():
-    print 'Usage: sugar-install-bundle [ bundlename ] \n\n\
-    Install an activity bundle (.xo). \n'
+    print 'Usage: sugar-install-bundle activitybundle.xo [...] \n\n\
+    Install activity bundles\n'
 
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Hum, you chmoded the script from 644 to 755.  That is something I haven't noticed, but all your scripts are 755 vs the rest which are 644.

755 seems to make more sense, but I'm not sure why they are 644.  Maybe somebody else knows?

quozl:
Apologies, I should not have committed the chmod.  It was done to test the scripts in isolation and when copying it to the target system for testing.  The bin_SCRIPTS variable in bin/Makefile.am ensures that the file will be chmod'd during "make install" step, so the mask it has in git is ignored.

Since the mask in git is ignored, I don't think it needs to be changed.

There's a good argument for setting the rest of the scripts to +x so that they can be run within the context of the source tree; developers need no barriers to their experimentation, and it would be otherwise harmless.

PR #600, opened 2015-10-31
Specify library version prior to import, fixes #4900

samdroid-apps:
Gi has been emitting warnings into the log to do this for some
time now.  Not doing this has also caused bugs, such as #4900.

Ticket URL <https://bugs.sugarlabs.org/ticket/4900>

CC @i5o

PR #609, opened 2015-11-05
Specify library version prior to import, fixes #4900

samdroid-apps:
Replaces #600, no longer assumes Maliit is installed.

Gi has been emitting warnings into the log to do this for some
time now.  Not doing this has also caused bugs, such as #4900.

Ticket URL <https://bugs.sugarlabs.org/ticket/4900>

### DIFF
--- a/extensions/deviceicon/frame.py
+++ b/extensions/deviceicon/frame.py
@@ -27,7 +27,9 @@ import jarabe.frame
 _ICON_NAME = 'module-keyboard'
 _HAS_MALIIT = False
 
+import gi
 try:
+    gi.require_version('Maliit', '1.0')
     from gi.repository import Maliit
 except ImportError:
     logging.debug('Frame: can not create OSK icon: Maliit is not installed.')
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -52,6 +52,15 @@ os.environ['SUGAR_VERSION'] = config.version
 from dbus.mainloop.glib import DBusGMainLoop
 DBusGMainLoop(set_as_default=True)
 
+# define the versions of used libraries that are required
+import gi
+gi.require_version('Gtk', '3.0')
+gi.require_version('Gst', '1.0')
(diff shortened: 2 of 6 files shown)


i5o:
Testing

i5o:
shell.log http://paste.fedoraproject.org/287120/67079611
*I think we will need to change all instances?*

org.laptop.WebActivity-1.log http://paste.fedoraproject.org/287121/70802114 (just sharing it, no idea abou what happened).

quozl:
For your (shell.log)[http://paste.fedoraproject.org/287120/67079611]

Check the patch is properly applied?

Line 57 of src/jarabe/main.py is either the import Gtk of the pre-patch condition, or the require_version in the post-patch condition.

i5o:
I don't have fedora right now, but I will get it ASAP and report back

samdroid-apps:
This one has come up in the field:  https://socialhelp.sugarlabs.org/t/unable-to-run-sugar-on-arch-linux-due-to-pygiwarning/138/3

samdroid-apps:
Could we look into getting this done for 110?  This patch fixes so much log noise, and log noise is very painful for development.

davelab6:
Added to https://wiki.sugarlabs.org/go/0.110/Feature_List - I see there is a Github issue tracker milestones that this is part of, but it seems Github Issues is deactivated...

samdroid-apps:
Yes, we use trac for issues.  The milestones feature is only available to the github org members anyways.

I will also make the feature as orphaned.  IDK If I will do it for this cycle, I don't know if I will have time yet.  You are welcome to give it a go though.

davelab6:
> we use trac for issues

I don't think we should, and started a thread on that topic, http://lists.sugarlabs.org/archive/sugar-devel/2016-April/051788.html :)

quozl:
Log noise is also painful for deployment, as it scales power and disk usage across a large number of devices.  I'm also in favour of this patch.

PR #610, opened 2015-11-05
File transfer fixes

samdroid-apps:
Fix for https://bugs.sugarlabs.org/ticket/4907 and auto-dismissing of completed transfers from the sender.

Still more to improve, eg. sending all the metadata so we get the icon, so we can have the bundlelauncher to launch it...  but it kind of works nice now :)

### DIFF
--- a/src/jarabe/frame/activitiestray.py
+++ b/src/jarabe/frame/activitiestray.py
@@ -524,6 +524,7 @@ class OutgoingTransferButton(BaseTransferButton):
     """
     def __init__(self, file_transfer):
         BaseTransferButton.__init__(self, file_transfer)
+        file_transfer.connect('notify::state', self.__notify_state_cb)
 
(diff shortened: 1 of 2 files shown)


quozl:
I'm fine with the second patch 9392e7b, well explained, well executed.  But the first patch fad0d41 is a bit opaque.  Why is the notification being removed?  The notification serves a clear purpose; it ensures the learner now knows that bandwidth is back to normal and the task is complete.  Just disappearing doesn't indicate that.

samdroid-apps:
Idk, but I don't see any reason for the notification to remain.  It does
not clearly indicate that the transfer is complete, it doesn't say
"completed" anywhere.  I think the learner will know if the bandwidth has
returned to normal more easily by the absence of the icon.  Plus, I don't
want 1 billion icons in my activities tray.

samdroid-apps:
Anyway, I'll seperate these into 2 patches

samdroid-apps:
Replaced by #611

PR #612, opened 2015-11-12
background control panel, drop idle task

quozl:
An idle task for loading images may continue to run after apply or undo.

- save the source identifier for the idle task,

- wipe the source identifier when the idle task is finished,

- remove the idle task if it is still running on apply or undo.

Fixes [#4911](https://bugs.sugarlabs.org/ticket/4911).


tchx84:
Reviewed, tested and merged.

PR #613, opened 2015-11-12
background control panel, alpha level callbacks

quozl:
Gio.Settings changes are memory and processor expensive, because the [instance persists](http://lazka.github.io/pgi-docs/Gio-2.0/classes/Settings.html#Gio.Settings) until the change is flushed to disk.

Extra changes were happening during button creation, and for every click of a button.  By default, the first button is created active by Gtk+.  (See [Gtk.RadioButton](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/RadioButton.html#Gtk.RadioButton)).

When the code tries to set another button as active, the clicked callback for the first button is triggered.  When followed by apply, this can lead to incorrect setting.

Changes in this patch are to:

- ignore callbacks for a button changing from active to inactive, by checking get_active(),

- suppress callbacks during button creation, by connecting the toggled signal after the buttons have been created,

- simplify the button creation code.

Fixes [#4912](https://bugs.sugarlabs.org/ticket/4912).

### DIFF
--- a/extensions/cpsection/background/view.py
+++ b/extensions/cpsection/background/view.py
@@ -79,19 +79,27 @@ class Background(SectionView):
 
         alpha_box = Gtk.HBox()
         alpha_buttons = []
-        for i in ['000', '020', '040', '060', '080']:
+        alpha_icons = [
+            [1.0, 'network-wireless-000'],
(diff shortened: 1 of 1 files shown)


tchx84:
Reviewed, tested, and merged.

PR #614, opened 2015-11-17
frame key double press consistency

quozl:
When the frame key is pressed rapidly, some presses are ignored, because the animation position is used to determine if the frame is visible.

To fix,

- remove class _KeyListener as it is not required, see also 3e28f56,

- add a flag for whether the frame is wanted or not, self._wanted,

- add a toggle() method, which calls show() or hide() according to last known wanted state, and call it from the mouse enter corner or frame key press methods,

- do not test frame visibility in show() or hide(), depend on _wanted,

- in the gesture handler, use toggle() instead of show() or hide(),

Test case:
```
	xdotool key F6 key F6 key F6
```
should be equivalent in final effect to
```
	xdotool key F6
```
Fixes [#4806](https://bugs.sugarlabs.org/ticket/4806).

### DIFF
--- a/src/jarabe/frame/frame.py
+++ b/src/jarabe/frame/frame.py
@@ -54,17 +54,6 @@ class _Animation(animator.Animation):
         self._frame.move(current)
 
 
-class _KeyListener(object):
-    def __init__(self, frame):
-        self._frame = frame
-
-    def key_press(self):
-        if self._frame.visible:
-            self._frame.hide()
-        else:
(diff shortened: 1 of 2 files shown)


samdroid-apps:
Another way to reproduce the issue:  move your mouse in and out of the corners with about 0.5sec between each corner activation

samdroid-apps:
Great patch, it makes sugar feel much more responsive.  I'll merge it.

PR #616, opened 2015-11-25
request for comment, not for merging, home view redraw excess

quozl:
- do not let the view children of the home window take focus; this prevents redraw, proving it is the focus change causing [#4914.2](https://bugs.sugarlabs.org/ticket/4914),

 - but as a side effect focus is always in the search box, and the blinking cursor is a bother; it consumes power,

 - gtk-cursor-blink Gdk.Settings might help with the blinking cursor, but don't know how to set them, changing the gtk-widgets.css files did nothing,

 - perhaps a better solution may be to ensure there are no focus rectangles drawn by children of the home window,

Any ideas welcome.

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -161,6 +161,7 @@ class ControlPanel(Gtk.Window):
         self._table.set_border_width(style.GRID_CELL_SIZE)
 
         self._scrolledwindow = Gtk.ScrolledWindow()
+        self._scrolledwindow.set_can_focus(False)
         self._scrolledwindow.set_policy(Gtk.PolicyType.AUTOMATIC,
                                         Gtk.PolicyType.AUTOMATIC)
         self._scrolledwindow.add_with_viewport(self._table)
--- a/src/jarabe/desktop/activitieslist.py
+++ b/src/jarabe/desktop/activitieslist.py
@@ -55,6 +55,7 @@ class ActivitiesTreeView(Gtk.TreeView):
 
     def __init__(self):
         Gtk.TreeView.__init__(self)
+        self.set_can_focus(False)
 
         self._query = ''
 
@@ -442,6 +443,7 @@ class ActivitiesList(Gtk.VBox):
         Gtk.VBox.__init__(self)
 
         self._scrolled_window = Gtk.ScrolledWindow()
+        self._scrolled_window.set_can_focus(False)
         self._scrolled_window.set_policy(Gtk.PolicyType.NEVER,
                                          Gtk.PolicyType.AUTOMATIC)
         self._scrolled_window.set_shadow_type(Gtk.ShadowType.NONE)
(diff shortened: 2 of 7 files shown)


godiard:
@quozl, maybe you can use https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Settings.html#Gtk.Settings.props.gtk_cursor_blink_timeout

godiard:
I think this is a example of set a Gtk.Setting:

https://github.com/sugarlabs/sugar/blob/master/src/jarabe/main.py#L346

samdroid-apps:
Works great, seems to do it's job.

Can we remove some of the focus steeling code from the homeview not that you can't focus anything other than the search bar?  (Or does that code only exsist in my mind?)

quozl:
Force pushed for further review only.

@samdroid-apps, yes, there is some focus stealing code that could be removed, if we go with this.

@godiard, thanks, that fixes the blinking cursor, brings idle redraws to nothing.

I also extended the patch to exclude focus from the control panel scrolled window.

I'm still worried about keyboard focus accessibility.  [#1969](https://bugs.sugarlabs.org/ticket/1969) and [#4891](https://bugs.sugarlabs.org/ticket/4891).  Almost the opposite of this work.  We would need to bring back a visible focus rectangle, accept the consequent redraws, and fix some of the problems.  Our gtk-widgets.css makes any focus rectangle invisible.  Gtk+ seems to redraw an entire widget when redrawing the focus rectangle, instead of splitting the drawing into segments.

After your next comments, it is probably time for me to post to sugar-devel@ to get other input.

samdroid-apps:
Yes the accessibility is an issue.  But does the homeview have keyboard focus accessibility now?  I speculate that making sugar work with keyboard focus would be a large task.  I'm also going to speculate that we could just drop down a palette of search suggestions more easily than fixing the whole focus model :smile:

I speculate that redrawing the entire widget on focus is always going to be needed - an entry looks very different focused vs unfocused.

godiard:
@quozl, imho, accessibility in the home view is implemented with the search box. Keyboard navigation was never implemented and would be a non trivial task.

quozl:
@godiard, using a shorter gtk-cursor-blink-timeout would increase redrawing and decrease battery run time.  Thanks for the input, I'll fork.

godiard:
@quozl, gtk_cursor_blink_timeout set "the time after which the cursor stops blinking, in seconds. The timer is reset after each user interaction."  not how fast the cursor blinks.

quozl:
@godiard, thanks, easily confused with gtk-cursor-blink-time which is duration of each blink.  So this seems good;

```
settings.set_property('gtk-cursor-blink-timeout', 5)
```

and on XO-1 to manipulate perception of responsiveness

```
settings.set_property('gtk-cursor-blink-timeout', 5)
settings.set_property('gtk-cursor-blink-time', 500)
```

also found that turning off the blink can make me think there is a vertical bar character.

godiard:
+1

PR #617, opened 2015-11-26
fix control panel layout

quozl:
Settings icons in the control panel were not evenly distributed, on screens with 1200 pixels horizontal, because some icon titles were very long.

Remove "Configuration" and "Configure your".

(the words are redundant; everything in my settings is a configuration).

### DIFF
--- a/extensions/cpsection/modemconfiguration/__init__.py
+++ b/extensions/cpsection/modemconfiguration/__init__.py
@@ -18,4 +18,4 @@ from gettext import gettext as _
 
 CLASS = 'ModemConfiguration'
 ICON = 'module-modemconfiguration'
-TITLE = _('Modem Configuration')
+TITLE = _('Modem')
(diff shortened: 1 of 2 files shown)


samdroid-apps:
I think this is a really good change.  With the patch:

[https://gyazo.com/b116aeb1ce4ae25463031eccd9b8f7e7](https://gyazo.com/b116aeb1ce4ae25463031eccd9b8f7e7)

Without the patch:

[https://gyazo.com/a7544fb5de266685e2094f5b93665d26](https://gyazo.com/a7544fb5de266685e2094f5b93665d26)

Note:  there are lots of gtk issues on gtk3-3.19.3-1.fc24.x86_64.  They are unrelated to this patch.

quozl:
Interesting, thanks.  From you *after* screenshot, still some extra width between second and third column, so there is an opportunity to make the icons evenly spaced by centres instead of spaced according to their label width.

PR #618, opened 2015-11-26
fix resume mode alt key handling in favourites

quozl:
Sometimes after stopping an activity that was started new with the Alt key, the visibility notify event is missed, and the favourites view is without user colour because the Alt key is presumed to be down.

Workaround is to briefly press and release the Alt key, and the colours return.

Fix is to abandon the visibility notify event, which also happens to be deprecated, in favour of a timeout that checks for Alt key release.

Also refactored the common code for detecting the Alt key in an event.

A behaviour change is that the Alt key has to be pressed while the home view is shown.  Test case; show journal, expose frame, press Alt, click on home view icon, which results in home view appearing as if the Alt key is not down.

### DIFF
--- a/src/jarabe/desktop/homewindow.py
+++ b/src/jarabe/desktop/homewindow.py
@@ -18,6 +18,7 @@ from gettext import gettext as _
 import logging
 
 from gi.repository import GLib
+from gi.repository import GObject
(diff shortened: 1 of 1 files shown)


quozl:
Will fork.

PR #619, opened 2015-11-26
remove transition box

quozl:
The transition box significantly slows use of Sugar when using keyboard shortcuts to change the zoom view.

While it is useful for the zoom analogy in the original design, it doesn't seem that this analogy has really caught on.  With the transition box removed, the sense of zooming does still occur, just really fast.


samdroid-apps:
Maybe you can just remove the transitionbox when using the keyboard shortcuts.

Animations are a really useful tool to explain the flow of interaction through a user interface, something which the zoom animation helps with.  You could make the argument that the frame animation uses resources, but it also explains where the frame is coming from and the fact that you can get rid of it (it's over the top of the current activity).  Removing the animation probably won't help explain the zoom analogy to users, which will probably make it harder to catch on.

Also animations look nice, and nice looking things are nice.  Removing animations makes sugar look more scary to users who don't understand it - as when you see some video you just see a clunky system rather than a nice smooth one.

This needs a design discussion, which doesn't appear to have been held.  I'll close this pull request, but feel free to re open it if you want to hold a discussion.  You should probably also email sugar devel, make a feature page or anything else that would stall the process.

As you may have guessed, I'm gonna give this a NAK.  But my opinion is no means the only, and maybe somebody else should review this.

CC @walterbender @godiard @tchx84

godiard:
@quozl, @samdroid-apps, first, I didn't tested this patch, then I don't know how much affect the user experience. In my personal opinion, this animation in particular do not add too much, and could be removed. Maybe we can add a config "disable_animations" and avoid this and others (like the palettes or the activity startup icon) when configured. XO-1 images could set the config by default, and keep the option to advanced or impatient users. Just my two cents.

quozl:
No worries, will fork.

PR #620, opened 2015-11-26
Frame trigger setting:  correct setting name when reading setting

samdroid-apps:
Fixes @quozl's comment on the previous patch.


tchx84:
This patch is included in @quozl recent PR https://github.com/sugarlabs/sugar/pull/632, therefore I close it.

PR #622, opened 2015-12-05
metacity shouldn't use compositing

rcunning01:
Compositing in metacity adds extra unneeded gradients to the Sugar user interface.

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -181,7 +181,7 @@ def _start_window_manager():
     settings = Gio.Settings.new('org.gnome.desktop.interface')
     settings.set_string('cursor-theme', 'sugar')
 
-    _metacity_process = subprocess.Popen(['metacity', '--no-force-fullscreen'])
+    _metacity_process = subprocess.Popen(['metacity', '--no-force-fullscreen', '--no-composite'])
 
     screen = Wnck.Screen.get_default()
     screen.connect('window-manager-changed', __window_manager_changed_cb)


samdroid-apps:
What do you mean by gradients?  Do you mean the shadows?  Screenshot please!

rcunning01:
Like this:
Screenshot

rcunning01:
Compare to this:
Screenshot

samdroid-apps:
So hum, I need to really evaluate this from a graphics perspective.  I'm kind of a shadows fan, but this might fix some of the bugs where each window of the frame puts a shadow over it's friends.

But I can't do that now, because gtk3-3.19.3-2.fc24.x86_64 has some other issues that make it look horrible:

[https://gyazo.com/e26aa7cebd8564f27f0d03e25c2c1655](https://gyazo.com/e26aa7cebd8564f27f0d03e25c2c1655)

tchx84:
I don't think Sugar was supposed to have these shadows in the first place, but I would like you hear to heat your thoughts @quozl @godiard @i5o @samdroid-apps

samdroid-apps:
Yeah sounds correct.  Plus shadows much with the palettes sometimes (eg. on the right frame panel).

Does this have a performance effect?  It's probably good right?

Does disabling compositing effect our ability to run arbitrary windowed X apps.

i5o:
I like the shadows, they looks pretty good.

But I think it's not in the original sugar designs https://wiki.sugarlabs.org/go/Design_Team/Designs/Frame , maybe we can add them and skip the designs for a moment?

quozl:
@tchx84, i'm fine with removing these new shadows.

tchx84:
@quozl alright.

@samdroid-apps (as far I understand) having these effects puts an extra cost on the rendering layer, specially if you don't have proper GPU support. But I haven't find anything conclusive in the metacity docs.

@i5o I personally don't see the point of using these shadows (specially IF they affect performance)

@rcunning01 before we make the final decision, we should check these other things to fully understand what else get affected by this change (for good and for bad). e.g, testing it with media players, etc.

@rcunning01  Can you help us with that research?

samdroid-apps:
From my very unscientific tests, WebKit2 watching a YouTube views gives me like 1fps when compositing is disabled.  I haven't directly compared with compositing, but I believe that it was much better in the past.  So, I'm saying no to the patch at this point.

Also, please go onto a new line - lines are limited to 80chars by pep8.

quozl:
There's probably a better way to remove shadows than turning off compositing.  The shadows will have been specified *somewhere*.

samdroid-apps:
Great.  Closing.  Please look for another solution.

PR #623, opened 2015-12-10
Favorites view name can be random

163gal:
GCI Task: https://codein.withgoogle.com/tasks/4788542454628352/
Now the favorite views can have names (Or if they doesn't have one, they receive one).

### DIFF
--- a/src/jarabe/model/desktop.py
+++ b/src/jarabe/model/desktop.py
@@ -15,6 +15,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
+from gettext import gettext as _
+
 from gi.repository import GObject
 from gi.repository import Gio
 
@@ -37,6 +39,7 @@ class DesktopViewModel(GObject.GObject):
         self._number_of_views = 1
         self._view_icons = None
         self._favorite_icons = None
+        self._view_labels = None
 
         self._settings = Gio.Settings(_DESKTOP_CONF_DIR)
         self._ensure_view_icons()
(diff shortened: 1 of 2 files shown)


quozl:
Thanks for your contribution.  I've reviewed the patch.

Why are there two assignments to variable favorite_icons, one of which is a getter of get_view_labels?

I don't understand how to test the patch.  How can a favourite view have a name?

The patch description should say what the problem is, how it is fixed, and how it can be tested.

I suggest you make changes and do a force push for further review.  I'm not inclined to review a different pull request, as I'll lose my comments.

i5o:
Hey @quozl 

I will fix the description and those things, its for a gci task.
This is for the feature "multiple homeviews", before gsettings port it was possible to change the homeviews names (using  the activity "Manage homeviews": https://github.com/i5o/manage-homeviews )
Looks like it was lost after the gsettings port.

src/jarabe/model/desktop.py
```diff
@@ -58,6 +61,11 @@ def get_favorite_icons(self):
 
     favorite_icons = GObject.property(type=object, getter=get_favorite_icons)
 
+    def get_view_labels(self):
+        return self._view_labels
+
+    favorite_icons = GObject.property(type=object, getter=get_view_labels)
```
samdroid-apps: Typo here

samdroid-apps:
Please make a patch for manage home views so that i can test.

i5o:
@samdroid-apps I recommend you to test with https://wiki.sugarlabs.org/go/Features/Multiple_home_views gsettings instructions for Sugar 102+
adding view-label key..

That is another GCI task, and I think student aren't able to work on two at the same time.

quozl:
@i5o in IRC referenced https://wiki.sugarlabs.org/go/Features/Multiple_home_views which in my opinion may need editing which may form part of the overall work, and would be a great thing to mention in the patch description.

That this pull request is associated with a GCI task, or that coordinating multiple students is required, shouldn't be part of the pull request discussion.

163gal:
I'm fixing the typo.

PR #624, opened 2015-12-10
Add the ability for change homeviews names

163gal:
Before the GConf to Gsettings port changing the homeviews names
was possible.

This patch adds the ability to change it again using gsettings
config, "view-label".

How to test:
Follow the instruction for "gsettings instructions for Sugar 102+" in
https://wiki.sugarlabs.org/go/Features/Multiple_home_views
and add the key "view-label" to the homeview you want.

This is a GCI task:
https://codein.withgoogle.com/tasks/4788542454628352/

### DIFF
--- a/src/jarabe/desktop/viewtoolbar.py
+++ b/src/jarabe/desktop/viewtoolbar.py
@@ -205,7 +205,7 @@ class FavoritesButton(RadioToolButton):
     def __init__(self, favorite_view):
         RadioToolButton.__init__(self)
 
-        self.props.tooltip = _('Favorites view %d' % (favorite_view + 1))
+        self.props.tooltip = desktop.get_view_labels()[favorite_view]
         self.props.accelerator = _('<Ctrl>%d' % (favorite_view + 1))
         self.props.group = None
         self.props.icon_name = desktop.get_view_icons()[favorite_view]
--- a/src/jarabe/model/desktop.py
+++ b/src/jarabe/model/desktop.py
@@ -15,6 +15,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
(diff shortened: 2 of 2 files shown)


quozl:
Your test instructions are too complex.  Reference to an already completed feature was confusing.  There was no feature page for changing home view name.  Please review and contribute to my [simplified instructions in a new feature page](https://wiki.sugarlabs.org/go/Features/Change_home_view_name).

i5o:
I thought I replied this, but looks like I didn't.

You can use "Manage Home views" activity for add/remove homeviews, modify homeviews names, icons, etc. all with a graphical interface.

https://github.com/i5o/manage-homeviews

samdroid-apps:
Yes, this is a good patch.  One interesting thing is that while I can instantly add a homeview without restarting sugar, I must restart to see the name change come into effect.

quozl:
@i5o, @samdroid-apps, yes, it would be better if restart is not needed, and also manage-homeviews should be brought into sugar as a control panel, not an activity.

i5o:
I would work on the control panel thing as feature then.

I don't know if we have time (I mean, land for 0.108)

2015-12-20 19:28 GMT-03:00, James Cameron <notifications@github.com>:
> @i5o, @samdroid-apps, yes, it would be better if restart is not needed, and
> also manage-homeviews should be brought into sugar as a control panel, not
> an activity.
>
> ---
> https://github.com/sugarlabs/sugar/pull/624#issuecomment-166159489


-- 
Ignacio Rodríguez
SugarLabs at Facebook

samdroid-apps:
Yeah, I think that if you do a pretty restrained design we can definatly
get it merged!

On Mon, Dec 21, 2015 at 9:51 AM, Ignacio Rodríguez <notifications@github.com

quozl:
@i5o, that's a question for @tchx84 as release engineer.  :grin:

PR #434, opened 2014-12-11
Hide 'Invite to <activity>' menu option if the activity cant be shared

i5o:
With this changes, the 'Invite to <activity>' option in the
buddy menu only show if max_participants is <= 1.
max_participants is a field off activity.info, the default is 0
so, if you want you activity sharable set max_participants
to one number > 1

Fixes SL#2702

PR #625, opened 2015-12-11
Fix "Invite to <activity>" shown even for activities that don't support sharing

erilyth:
A fix for the issue #2702 (https://bugs.sugarlabs.org/ticket/2702)

### DIFF
--- a/src/jarabe/view/buddymenu.py
+++ b/src/jarabe/view/buddymenu.py
@@ -29,6 +29,7 @@ from sugar3.graphics.palettemenu import PaletteMenuItem
 from sugar3.graphics.icon import Icon
 from sugar3.graphics.alert import TimeoutAlert
 from sugar3.graphics import style
+from sugar3.bundle.activitybundle import ActivityBundle
 
 from jarabe.model import shell
 from jarabe.model import friends
@@ -174,9 +175,11 @@ class BuddyMenu(Palette):
             buddy_activity_id = buddy_activity.activity_id
         else:
             buddy_activity_id = None
-
+        if activity is not None:
+            bundle_activity = ActivityBundle(activity.get_bundle_path())
         if activity is None or activity.is_journal() or \
-           activity.get_activity_id() == buddy_activity_id:
+           activity.get_activity_id() == buddy_activity_id or \
+           bundle_activity.get_max_participants() <= 1:
             self._invite_menu.hide()
         else:
             title = activity.get_title()
@@ -220,3 +223,4 @@ class BuddyMenu(Palette):
                     raise
         else:
             logging.error('Invite failed, activity service not ')
+


godiard:
Looks nice. Check your code with pep8 and do a test with activities that do not have the field max_participants in the file activity.info. (This field was a recent addition, not all the activities has been updated yet)

erilyth:
In the activitybundle, max_participants is initialized to 0, so currently the "Invite" option is not displayed if max_participants is not set in the activity.info file. Should this be displayed for activities which don't have max_participants set as well?

quozl:
Reviewed this and also @i5o's abandoned #434.

Agree with @godard that commit should describe tests with activities that do not have the field max_participants.

Please check that you are handling the journal activity correctly; the code in #434 is quite different to yours.

In my opinion the invite option should be hidden for activities that do not have max_participants set.  This is also consistent with #434.

Please add to the commit message; what was wrong, what is fixed, how it is fixed. how to test.

Please also rebase your patches on this pull request to create one patch, and then force push.  (Because merging two patches will make history harder to use for regression analysis.)

erilyth:
In #434 , The journal hides the invite menu as well, and the same thing is being done here, so I think it should be fine.

erilyth:
Any updates on this?

quozl:
@erilyth, I know how you feel; I've several pull requests waiting for review and merge.  The next 0.108 date milestone is 28th December.  Thanks for the force push, I've reviewed the commit.  

I continue to agree with @godiard that you should test with activities that do not have the field max_participants.  You may have done this, I'm not sure.  The best place to put it is in the commit message, because that is what will be used later when someone reports a bug.  Write something like "Tested with activity Name-99 which does not have field max_participants."

My other comments above, you have already integrated.

erilyth:
Ok, sure quozl :). I will add that to the commit message and force push it.

src/jarabe/view/buddymenu.py
```diff
@@ -24,11 +24,13 @@
 from gi.repository import GObject
 import dbus
 
+
```
samdroid-apps: Why this newline?

samdroid-apps:
This looks good, I will test.  I'd appreciate it though if you kept the summary (the 1st line) of your commit message less thatn 80chars.  Then it displays correctly in most places.  Maybe like "Do not show invite option for activities that do not support sharing"  (if that is less than 80chars :smile:).

@quozl  My tests lead me to believe that this patch makes activities without max_partcipants not show the invite menu.  I agree that this is a good thing, as older activities collab is probably broken anyway (from my fedora rawhide centric view of the world).

I will merge once you fix up the commit message @erilyth!

erilyth:
Done :)

quozl:
I continue to wish for the commit message to refer to the testing that has been done, rather than depend on GitHub to keep the pull request discussion.  At the moment the commit message does not do that.

erilyth:
Updated the commit message, added test details for the write and read activities.

quozl:
Thanks.  I've no further wishes.

samdroid-apps:
Great!  It still works as intended.  Sorry for taking soo long to merge this.

PR #627, opened 2015-12-14
fix resume mode alt key handling in favourites

quozl:
Sometimes after stopping an activity that was started new with the Alt
key, the visibility notify event is missed, and the favourites view is
without user colour because the Alt key is presumed to be down.

Workaround is to briefly press and release the Alt key, and the colours
return.

Fix is to abandon the visibility notify event, which also happens to be
deprecated, in favour of a timeout that checks for Alt key release.

Also refactored the common code for detecting the Alt key in an event.

### DIFF
--- a/src/jarabe/desktop/homewindow.py
+++ b/src/jarabe/desktop/homewindow.py
@@ -18,6 +18,7 @@ from gettext import gettext as _
 import logging
 
 from gi.repository import GLib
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Ah, this makes sense now.  You don't poll constantly, you only poll for alt up!

great.  Merging.

PR #628, opened 2015-12-14
reduce home view redraw excess

quozl:
- do not let the view children of the home window take focus; this
  prevents redraw, proving it is the focus change causing #4914.2,

- as a side effect focus is always in the search box, so the blinking
  cursor is turned off quickly,

However, breaks and makes harder any solution for two tickets:

https://bugs.sugarlabs.org/ticket/1969 (Keyboard navigability of the
Sugar UI) closed by Daniel Narvaez for being non-specific, and would
require visible focus rectangles to be restored.

https://bugs.sugarlabs.org/ticket/4891 (no journal accessibility keys)
which turns out to be mentioned already by 1969.

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -162,6 +162,7 @@ class ControlPanel(Gtk.Window):
         self._table.set_border_width(style.GRID_CELL_SIZE)
 
         self._scrolledwindow = Gtk.ScrolledWindow()
+        self._scrolledwindow.set_can_focus(False)
         self._scrolledwindow.set_policy(Gtk.PolicyType.AUTOMATIC,
                                         Gtk.PolicyType.AUTOMATIC)
         self._scrolledwindow.add_with_viewport(self._table)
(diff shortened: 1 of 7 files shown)


samdroid-apps:
Hum, how do I test?

So far I have:

{With,Without} the patch, run sugar.  Open Gtk+ Inspector and turn on flashing on redraws.  Switch between home view and buddy view and notice that there is lots of flashing no matter what.

quozl:
Gtk+ Inspector is not available on the Fedora 18 builds where this problem matters most, so per 4914 it should be tested with set_debug_updates.  Wouldn't be surprised that later Gtk+ does lots of flashing no matter what, but it isn't in my test scope.  You're welcome to test that, but use the test method below;

- apply set_debug_updates in homewindow.py and restart jarabe,
- make the home view visible,
- try to change the focus widget using tab or shift-tab,
- popup and popdown the buddy menu, and observe the redraw area,

Without the patch, the focus widget can be either the search entry box or the home view, and when it is the home view the redraw of the invisible focus indicator causes redraw area to cover the home view; i.e. most of the display.

With the patch, the focus widget is always the search entry box, and the redraw area is the search entry box and palette only.

On an XO-1 this reduces the power cost and display latency of calling up the buddy menu and activity palettes.

quozl:
Ping?  Can we get this merged now?

PR #629, opened 2015-12-14
enhance favourites

quozl:
I've tested this work by Martin, and submit it for review and merge.


quozl:
Rebased against master and force pushed.  GitHub identifies the rebased commits as added by @tchx84, but does not identify me as rebase author.

samdroid-apps:
Hum, this seems to fail importing previous favourites.  I found this by applying this patch to sugar, then noticing that every activity was now marked as favourite (but not terminal or log).  It also touched many activities that I had never unfavourited, but were unfavourited by default.

I speculate this was due to the _convert_old_favourites not accounting for the fact that many of the favourites aren't stored if they don't differ from the default.  However, I have not touched this code in ages, so I am interested to hear how you believe that this could be solved.

Alternatively, we could just try and impose the change on users and force them to go and change their favourites.  But that would be badly received.

quozl:
@tchx84, what do you think?  (re: experience on upgrade for an existing user).

@samdroid-apps, which particular users are you asking about?  Fedora SoaS users don't upgrade.  Developers can ignore the problem.

I've been handling this already for OLPC OS with a couple of patches (in a fork of sugar prior to .tar.xz) that blend the olpc-os-builder favourites with this code.  In effect these add the following to data/activities.hidden:

OLPC Library content bundles

```
org.laptop.biology
org.icdlbooks.icdl
org.jamendo.music
org.laptop.wp-nature-images
info.dicts.dictionary-ml
org.laptop.wikibooks
info.dicts.wikislice-chemistry
gov.loc.wdl
info.dicts.worldfactbook.small
```

OLPC OS 13.2.6 nominated hidden bundles

```
org.laptop.ImageViewerActivity
org.laptop.sugar.ReadActivity
org.laptop.TamTamEdit
org.laptop.TamTamSynthLab
```

@samdroid-apps, can you tell me if the favourites change you observed is any different from the above?

quozl:
Ping?  Can we get this merged now?

samdroid-apps:
Ok, so disregarding the issue that happens on upgrade (as we decided was a non issue from above), I ran the following tests:

* Tested that adding a favourited activity to the `activities.hidden` file does not cause it to hide
* Tested that hating an activity is persistent
* Tested that downloading an activity via browse causes it to be installed and favourited

I will merge it as it doesn't seem to break anything too big.  However if anybody wishes to test further cases please do so!

PR #630, opened 2015-12-15
Fix remaining icon_size references

quozl:
icon_size was changed to pixel_size, yet several references remained,
causing extra log data.


godiard:
You need change the constant defining the size too, see https://developer.sugarlabs.org/release-notes/102.md.html#icon%20size

quozl:
Have done.  Sadly GitHub does not permit reopen, so a new pull request #634 was opened.

PR #632, opened 2015-12-15
Avoid restart in frame control panel

quozl:
Use the Gio.Settings 'changed' signal to avoid having to restart Sugar when the frame control panel is used.

Also fixes an interesting problem on real hardware, see c92dcf3 for details.

Includes #620.

### DIFF
--- a/extensions/cpsection/frame/model.py
+++ b/extensions/cpsection/frame/model.py
@@ -38,10 +38,6 @@ def set_corner_delay(delay):
     delay: 100 (100 milliseconds)
     never: 1000 (disable activation)
     """
-    try:
-        int(delay)
-    except ValueError:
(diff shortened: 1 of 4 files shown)


samdroid-apps:
Reviewed, not tested.

I believe that the following is unrelated to your patch:

[https://gyazo.com/b850d1a6d825c14eaba842612869d852](https://gyazo.com/b850d1a6d825c14eaba842612869d852)

quozl:
Yes.

tchx84:
Tested and merged, @quozl please add a couple lines to our Features page (just to keep track)

quozl:
Added [line to feature page](https://wiki.sugarlabs.org/index.php?title=0.108/Feature_List&diff=96699&oldid=96686).

PR #633, opened 2015-12-15
Add drag from notification to act like a clipboard entry

erilyth:
If a user starts a drag action, then the timeout is reset to 5 seconds, if necessary we can increase the timer to whatever value needed.

### DIFF
--- a/src/jarabe/frame/clipboardicon.py
+++ b/src/jarabe/frame/clipboardicon.py
@@ -20,6 +20,7 @@ import logging
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import SugarExt
+from gi.repository import GObject
 
 from sugar3.graphics.radiotoolbutton import RadioToolButton
 from sugar3.graphics.icon import Icon
@@ -75,6 +76,9 @@ class ClipboardIcon(RadioToolButton):
 
     def _drag_data_get_cb(self, widget, context, selection, target_type,
                           event_time):
+        frame = jarabe.frame.get_view()
+        self._timeout_id = GObject.timeout_add(
+            jarabe.frame.frame.NOTIFICATION_DURATION, lambda: frame.remove_notification(self._notif_icon))
         target_atom = selection.get_target()
         target_name = target_atom.name()
         logging.debug('_drag_data_get_cb: requested target %s', target_name)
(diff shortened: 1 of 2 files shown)


samdroid-apps:
I don't think this is the best way to do this.  This change should only affect the clipboard code, not the frame code.

Why don't you deal with adding the drag callbacks and targets before sending the icon to the frame to make it a notification?  In https://github.com/sugarlabs/sugar/pull/633/files#diff-be714cfaae37541410117916f40be501R166  You could probably reuse the callbacks for the other icon there as well, meaning we have less code (yay!)

erilyth:
Yup, that surely wasn't the best way to do it :), its a lot simpler now. How do I go about increasing the timeout without changing code in frame.py.

samdroid-apps:
Sorry for the slow review.

Maybe for the timeout you can do the following:

* Make add_notification return the timeout id (please make sure to document this)
* In the drag_begin_cb, `GObject.source_remove(timeout_id)`
* In the drag_data_get_cb, re add the timeout, which should be like:

        timeout_id = GObject.timeout_add(duration, lambda: frame.remove_notification(icon))

erilyth:
I added the timeout as you specified

src/jarabe/frame/clipboardicon.py
```diff
@@ -75,6 +76,8 @@ def get_object_id(self):
 
     def _drag_data_get_cb(self, widget, context, selection, target_type,
                           event_time):
+        frame = jarabe.frame.get_view()
+        self._timeout_id = GObject.timeout_add(5000, lambda: frame.remove_notification(self._notif_icon))
```
samdroid-apps: Can you please use the constant for the timeout here

samdroid-apps:
Oh, this is good!  I tested and it works super well!

Please see the comment about the constant

Please squash all the commits into 1

samdroid-apps:
Then I will merge :)

erilyth:
Used the constant for the timeout and squashed commits into 1

src/jarabe/frame/clipboardicon.py
```diff
@@ -75,6 +76,8 @@ def get_object_id(self):
 
     def _drag_data_get_cb(self, widget, context, selection, target_type,
                           event_time):
+        frame = jarabe.frame.get_view()
+        self._timeout_id = GObject.timeout_add(jarabe.frame.frame._NOTIFICATION_DURATION, lambda: frame.remove_notification(self._notif_icon))
```
samdroid-apps: 1.  Lines should be max 80chars
2.  The _ in a variable signals it is private.  Maybe we need to change _NOTICICAION_DURATION to just NOTIFICATION_DURATION?

erilyth:
Ran a PEP-8 check and changed variable from private to public.

quozl:
How to test?

samdroid-apps:
1.  Copy something to the clipboard
2.  Drag from the notification icon that pops up in the bottom left

PR #635, opened 2015-12-17
Configurable launcher pulsing icon interval

quozl:
On some computers, the activity startup pulsing icon steals resources, increasing startup time.

Add a launcher-interval setting, and use it in pulsing icon.

Test case:
```
gsettings set org.sugarlabs.desktop launcher-interval 500
```
Then restart Sugar.  The pulsing icon should be slow.

Measurements:

On one of the oldest computers that Sugar is used, an OLPC XO-1, CPU usage of the X server and Jarabe were measured during launch of Moon activity using top(1):

- launcher-interval=50 ms, gave 42% usage,

- launcher-interval=100 ms, gave 21% usage,

- launcher-interval=500 ms, gave 8% usage,

- launcher-interval=1000 ms, gave 4% usage.

Test configuration was Fedora 18 with Sugar 0.107.0, and commit d349379 (launcher, reduce redraw area).

References:

- olpc-os-builder/v7.0 [xo1: reduce pulsing icon rate](http://dev.laptop.org/git/projects/olpc-os-builder/commit/?h=v7.0&id=28b669a1004207b537fe82e2211a44acfbb60a6a)

### DIFF
--- a/src/jarabe/view/launcher.py
+++ b/src/jarabe/view/launcher.py
@@ -17,6 +17,7 @@
 import logging
 from gettext import gettext as _
 
+from gi.repository import Gio
 from gi.repository import Gtk
 from gi.repository import Gdk
 
@@ -25,6 +26,9 @@ from sugar3.graphics import style
 
 from jarabe.model import shell
 from jarabe.view.pulsingicon import PulsingIcon
+
+
+_INTERVAL = 100
 
 
 class LaunchWindow(Gtk.Window):
@@ -55,7 +59,8 @@ class LaunchWindow(Gtk.Window):
         self._activity_id = activity_id
 
         self._activity_icon = PulsingIcon(file=icon_path,
-                                          pixel_size=style.XLARGE_ICON_SIZE)
+                                          pixel_size=style.XLARGE_ICON_SIZE,
+                                          interval=_INTERVAL)
         self._activity_icon.set_base_color(icon_color)
         self._activity_icon.set_zooming(style.SMALL_ICON_SIZE,
                                         style.XLARGE_ICON_SIZE, 10)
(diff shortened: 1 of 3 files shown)


src/jarabe/view/launcher.py
```diff
@@ -27,6 +28,13 @@
 from jarabe.view.pulsingicon import PulsingIcon
 
 
+try:
+    settings = Gio.Settings('org.sugarlabs.desktop')
+    _INTERVAL = settings.get_int('launcher-interval')
+except:
+    _INTERVAL = 100
```
samdroid-apps: Do we need this try/except?  Shouldn't this always be there?

quozl:
Thanks for the review.

@godiard wrote:
> ... to not execute in import

My error, has been moved to setup().

@godiard wrote:
> Needed here?

My error, removed redundant import Gio.

@godiard wrote:
> try don't change the API

Good idea.  Pulser is not used outside pulsingicon.py.  Let's deprecate it, make it local only.  We're not yet at API freeze.  PulsingIcon API change is adding a keyword argument.

@samdroid-apps wrote:
> Shouldn't this always be there?

Defensive programming.  It won't be there when the schema is unchanged, which happens if patches are applied to a running system without calling glib-compile-schemas.  Your patch to add trigger-size caused days of pain, and I am holding a local patch that catches that exception.  But as you think it is not necessary to defend against this, I've removed it, and will hold a local patch.

Forced push.

quozl:
@godiard wrote:
> You can keep the API compatible doing:

Okay, done.

samdroid-apps:
Hum, seems to not preserve the speed.  Eg. with 20ms:  https://vid.me/e/zccX

(Yes, that video is 100% legit speed)

However, I don't see this as an issue that should prevent this patch from merging.  The only use case is really for slow XO1s, it really doesn't make sense to waste cpu cycles even on a good computer to have smooth launch animations, despite being pretty.

samdroid-apps:
Also there is a lot of gtk3 breakage in that video.  But it is kind of different from how it broke 2 weeks ago

I might try and have a look at fixing it, hopefully we can ship a release that looks correct with gtk for once :)

quozl:
Yes, it also changes the speed; that's because the speed is set by the interval.

I'm not seeing any significant Gtk+ breakage in the Fedora 18 environment.  Several irritating messages in logs because of changes to sugar-artwork:gtk3/theme/{gtk.css,gtk-widgets.css.em} that are not recognised by the earlier Gtk+.

PR #636, opened 2015-12-21
Neighborhood buddies aren't placed randomly

163gal:
* Neighborhood buddies are placed according to their key
* Bug ticket 381 (https://bugs.sugarlabs.org/ticket/381)
* This is a GCI 2015 task: https://codein.withgoogle.com/tasks/6733138625560576/

### DIFF
--- a/src/jarabe/desktop/favoriteslayout.py
+++ b/src/jarabe/desktop/favoriteslayout.py
@@ -172,7 +172,15 @@ class SpreadLayout(ViewLayout):
         for child in children:
             if not self._grid.is_in_grid(child):
                 width, height = self._get_child_grid_size(child)
-                self._grid.add(child, width, height, None, None, locked=False)
+                x = y = None
+                if hasattr(child, "get_buddy"):
+                    md5hash = hashlib.md5(child.get_buddy().get_key())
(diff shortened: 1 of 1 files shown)


src/jarabe/desktop/favoriteslayout.py
```diff
@@ -172,7 +172,15 @@ def allocate_children(self, allocation, children):
         for child in children:
             if not self._grid.is_in_grid(child):
                 width, height = self._get_child_grid_size(child)
-                self._grid.add(child, width, height, None, None, locked=False)
+                x = y = None
+                if hasattr(child, "_buddy"):
```
samdroid-apps: You typically don't want to access _ methods or props outside the class.  Maybe add a function "get_placement_key" and add that to the child class?

samdroid-apps:
Please link to the bug ticket in your commit description

samdroid-apps:
It looks fine from a code review perspectve.

From a gci perspective, the bug isn't just about buddies, it's also about wifi points, activities, etc.  Maybe make a general "get_placement_key" function and add it to buddys, wifi points, shared activities, etc.

163gal:
Let's do that then.

PR #637, opened 2015-12-21
Neighborhood icons are no longer placed randomly

163gal:
* Neighborhood icons are placed according to their data
* Bug ticket 381 (https://bugs.sugarlabs.org/ticket/381)
* This is a GCI 2015 task: https://codein.withgoogle.com/tasks/6733138625560576/

### DIFF
--- a/src/jarabe/desktop/favoriteslayout.py
+++ b/src/jarabe/desktop/favoriteslayout.py
@@ -172,7 +172,17 @@ class SpreadLayout(ViewLayout):
         for child in children:
             if not self._grid.is_in_grid(child):
                 width, height = self._get_child_grid_size(child)
-                self._grid.add(child, width, height, None, None, locked=False)
+                x = y = None
+
+                if hasattr(child, "get_positioning_data"):
+                    md5hash = hashlib.md5(child.get_positioning_data())
+                    digest = abs(hash(md5hash.digest()))
+                    w = (self._width - (width * 3)) / 4
+                    h = (self._height - (height * 3)) / 4
+                    x = ((digest & 0xFFFFFFFF) % w) * ((digest >> 126) + 1)
+                    y = ((digest >> 32) % h) * (((digest >> 124) & 0b11) + 1)
+
+                self._grid.add(child, width, height, x, y, locked=False)
 
             requisition = child.get_preferred_size()[0]
             rect = self._grid.get_child_rect(child)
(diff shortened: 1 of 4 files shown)


src/jarabe/desktop/favoriteslayout.py
```diff
@@ -172,7 +172,36 @@ def allocate_children(self, allocation, children):
...
+                    x = ((digest & 0xFFFFFFFF) % w) * ((digest >> 126) + 1)
+                    y = ((digest >> 32) % h) * (((digest >> 124) & 0b11) + 1)
+                elif hasattr(child, "activity_id"):
+                    md5hash = hashlib.md5(str(child.activity_id))
+                    digest = abs(hash(md5hash.digest()))
+                    w = (self._width - (width * 3)) / 4
+                    h = (self._height - (height * 3)) / 4
+                    x = ((digest & 0xFFFFFFFF) % w) * ((digest >> 126) + 1)
+                    y = ((digest >> 32) % h) * (((digest >> 124) & 0b11) + 1)
+                self._grid.add(child, width, height, x, y, locked=False)
```
samdroid-apps: there is a lot of duplicated code there!  Maybe instead of checking for each type of object, you could add a `get_positioning_data` function or something to each object.  That would return the data, whether it is the buddy key,network_has, channel or activity id.

quozl:
Very repetitive code.  Please refactor; extract the repetition into a separate method.

163gal:
Removed repetitive code, now it uses "get_positioning_data" method.
Thanks Sam!

samdroid-apps:
Code looks good, I will test

samdroid-apps:
Great!  It would also be really good to look into doing this with the buddies view.  It's kind of weird how a buddy on the left in the neighbour hood view can be on the right in the buddies.

quozl:
On the neighbourhood view confine the buddies to be less than a third of the maximum radius from centre.  On the friends view use the same positions, but with radius multiplied by three.  In the transition box, animate the buddy icons so that they glide into positions.

quozl:
Tested.

quozl:
@163gal, please tell me how to gather data for you to diagnose this regression.

@samdroid-apps, can you revert meanwhile, unless you can see what the problem might be?

samdroid-apps:
Oh, let's learn more about what is going wrong before we race to revert the patch

163gal:
Excuse me, my English is not so good.
What is the problem?

163gal:
I can't replicate the problem because I don't have an XO. I would need to see what's happening with the Y axis.

163gal:
I tested it on Fedora 23, and it seems to be a problem there. Testing...

163gal:
captura de pantalla de 2015-12-22 12-15-03

163gal:
#640

PR #638, opened 2015-12-22
Control panel section keyboard activation

quozl:
Adds keyboard navigation to control panel.

We maintain a list of selected options from search results, detect the return key and if unambiguous result, open section view.

Does not autocomplete like home view search.

### DIFF
--- a/src/jarabe/controlpanel/gui.py
+++ b/src/jarabe/controlpanel/gui.py
@@ -85,6 +85,7 @@ class ControlPanel(Gtk.Window):
             'size-changed', self.__size_changed_cb)
 
         self._busy_count = 0
+        self._selected = []
 
     def __realize_cb(self, widget):
         self.set_type_hint(Gdk.WindowTypeHint.DIALOG)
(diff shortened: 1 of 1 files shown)


samdroid-apps:
> Does not autocomplete like home view search.

Any reason why¿  consistency is nice.

quozl:
I could make some reasons up.  Any reason why?  Perhaps you could do that part and make a pull request with both features?

samdroid-apps:
Na, this is fine.  I think that the list of cpanels is always visible, so your searching on that whereas in the home view you are searching on an invisible list.

samdroid-apps:
No, sorry, I mis-interpreted the patch.  Testing, say I type `back`, and press enter.  Nothing happens :(  I'll hack soomething on top of that.

samdroid-apps:
My apologies, that was a bad test.  There is both background and data.  Evidently the disabled styles are broken again.

quozl:
Thanks.  I didn't see that effect on Fedora 18.  :grin:

PR #639, opened 2015-12-22

erilyth:
We can drag anything from the clipboard to the journal and it gets stored in the journal. This works if we drag from the notification as well.

### DIFF
--- a/src/jarabe/frame/activitiestray.py
+++ b/src/jarabe/frame/activitiestray.py
@@ -24,6 +24,7 @@ import os
 from gi.repository import GObject
 from gi.repository import Gio
 from gi.repository import Gtk
+from gi.repository import Gdk
 
 from sugar3.graphics import style
 from sugar3.graphics.tray import HTray
@@ -50,6 +51,10 @@ from jarabe.frame.frameinvoker import FrameWidgetInvoker
 from jarabe.frame.notification import NotificationIcon
 from jarabe.frame.notification import NotificationButton
 from jarabe.frame.notification import NotificationPulsingIcon
+from jarabe.frame.clipboardmenu import ClipboardMenu
+from jarabe.frame.clipboardobject import ClipboardObject
+from jarabe.frame.clipboardtray import _ContextMap
+from jarabe.frame.clipboardobject import Format
 import jarabe.frame
 
 
@@ -63,11 +68,23 @@ class ActivityButton(RadioToolButton):
         self._home_activity = home_activity
         self._notify_launch_hid = None
 
+        self._context_map = _ContextMap()
+        self._cb_object = None
         self._icon = NotificationPulsingIcon()
         self._icon.props.base_color = home_activity.get_icon_color()
         self._icon.props.pulse_color = \
             XoColor('%s,%s' % (style.COLOR_BUTTON_GREY.get_svg(),
                                style.COLOR_TOOLBAR_GREY.get_svg()))
+        if self._home_activity.is_journal():
+            self._icon.drag_dest_set(Gtk.DestDefaults.ALL,
+                                     [], Gdk.DragAction.COPY)
+            self._icon.drag_dest_add_text_targets()
+            self._icon.drag_dest_add_image_targets()
+            self._icon.drag_dest_add_uri_targets()
+            self._icon.connect('drag_drop', self.__drag_drop_cb)
+            self._icon.connect('drag_data_received',
+                               self.__journal_drag_data_received_cb)
+
         if home_activity.get_icon_path():
             self._icon.props.file = home_activity.get_icon_path()
         else:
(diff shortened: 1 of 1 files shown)


src/jarabe/frame/activitiestray.py
```diff
@@ -309,12 +310,34 @@ def __activity_added_cb(self, home_model, home_activity):
             group = None
 
         button = ActivityButton(home_activity, group)
+        #If its the journal, the activity name is None
+        if home_activity.get_activity_name() == None:
```
erilyth: Is there a better way to check if we are dragging onto the journal only and not any other activity?

samdroid-apps: Looks fine

samdroid-apps:
Hum, I'm not 100% sure that this is the best way to do this from a UX perspective.

Maybe you could make the journal icon a drag source (in the same way as the clipboard tray is) and then it would accept anything (a drag from the clipboard, dragging directly to the journal, etc).  Is that even possible?

erilyth:
If we set it as the source and make sure to save the data received then it would work, but I thought this method would be better since we could add drag to other activities as well in the future easily. I will try do it as you specified and update the pr

erilyth:
Where exactly is the journal icon being created? Its the ClipboardIcon.py for clipboard icons but I am unable to find that for the journal icon

erilyth:
I added a drag source to the journal icon, and it stores objects that are dragged onto it

src/jarabe/frame/clipboardicon.py
```diff
@@ -176,11 +178,27 @@ def show_notification(self):
...
 
+    def _drag_end_cb(self, widget, context):
+        '''
+        If we are dropping onto the journal, only then we save the object in the journal,
+        the destination window (Gdk.X11Window) of the journal has the ID >= 8388800
+        '''
+        logging.debug(context.get_dest_window().get_xid())
+        print (context.get_dest_window().get_xid())
+        if context.get_dest_window()!= None and self.drag_inprocess == True and \
+                context.get_dest_window().get_xid()>=8388800 and context.get_dest_window().get_xid()<=8389000:
```
erilyth: I need a better way to do this, since this might fail at times.
Any ideas how I can check if the function is called only due to the drop on the journal and nowhere else?

samdroid-apps: You should not try and detect when a clipboard item is dropped on the journal.  You need to only change the code for the journal activity icon in the frame.

The journal icon should deal with getting the drag data and saving it.  See how the clipboard tray does it with the drag-data-received signal:  https://github.com/sugarlabs/sugar/blob/master/src/jarabe/frame/clipboardtray.py#L173

erilyth: So, should I use the selection that I receive from the clipboard and store it in the journal?

samdroid-apps: Yes

erilyth:
The drag_data_received() function attached to the journal activity icon now stores the data in the journal.

src/jarabe/frame/activitiestray.py
```diff
@@ -90,6 +100,53 @@ def __init__(self, home_activity, group):
...
+            tmp_dir = os.path.join(env.get_profile_path(), 'data')
+            f, file_path = tempfile.mkstemp(dir=tmp_dir)
+            try:
+                os.write(f, data)
+            finally:
+                os.close(f)
+        
+        if file_path != None:
+            jobject = datastore.create()
+            jobject.metadata['title'] = 'test'
```
erilyth: Is the title passed via the selection or do I need to pass it by myself and retrieve it?

src/jarabe/frame/activitiestray.py
```diff
@@ -90,6 +100,53 @@ def __init__(self, home_activity, group):
...
+            return
+        logging.debug('Saved to Journal')
+        self._drag_save = True
+        data = None
+        file_path = None
+        if selection.get_text() != None:
+	    data = selection.get_text()
+        if len(selection.get_uris()) != 0:
+            file_path = selection.get_uris()[0]
+        if selection.get_pixbuf() != None:
```
samdroid-apps: Hum, I think that you should change the order of priorities, as the clipboard can have multipul formats in it.

Maybe put images 1st, links 2nd and text 3rd?  Or just do whatever the clipboard does.

src/jarabe/frame/activitiestray.py
samdroid-apps: No it is not.  The clipboard just seems to make up a title.

Maybe you can reuse the logic from the clipboard?  I don't like the idea of duplicating the code.

samdroid-apps:
Besides the comments, it works pretty good!  Good job!

src/jarabe/frame/activitiestray.py
erilyth: I wanted to do that but for it I needed to pass the clipboard object to the drag destination, but I was unable to do that, since I was only able to pass text, uris, etc but not all objects. Could you help me with this?

erilyth:
I made the appropriate changes, and now I use the clipboard code to get the storage done without me having to do it manually :)

erilyth:
Now check all the formats and choose the best one to add to the journal

samdroid-apps:
Hum, are you clearing the clipboard object between drags?  I seem to be getting the same jobject created when I drag different text to the journal icon

erilyth:
It was happening because one last call of drag_data_received was happening after storing the previous entry into the journal. It is fixed now

samdroid-apps:
Awesome, this works great now.  Please run flake8 on your code, squash the commits and also prefix the callbacks with __, eg:

`drop_cb becomes __drop_cb`  - this is just the style that we use :)

Anyway, there is also a bigger issue that you need to fix with this patch.  To reproduce:

1.  Open browse
2.  Goto home page (with sugarlabs logo image)
3.  Drag that image into journal icon
4.  Notice that nothing is added the the journal

Issue:  the drag drop callback (when you try to save it to the journal) is called BEFORE the drag data signals, meaning the context map is empty when you try to save it to the journal.

Can you fix that as well?

erilyth:
Sorry for taking so long, I ran flake8, added the __ to callbacks and squashed the commits. I'll take a look at how I can fix the error with the drag of the image.

erilyth:
@samdroid-apps I tried to fix the images not getting added to the journal error you mentioned, but came across this.
When I drag images from the internet to my clipboard, the object stored cannot be opened with any activity unlike other normal images which can be opened with the browse or image viewer activity. The error I see in the log is that the "os.write(f,data)" fails when I call the copy_to_journal function since "Type Error: must be convertible to a buffer, not Pixbuf".
Now these images I am unable to drag them to the journal (they are not added to the journal) whereas other images I have stored or screenshots etc have no problem when I try to add them.
I'm not quite sure why this is happening, any thoughts? 
screenshot from 2016-03-16 12-43-24
screenshot from 2016-03-16 12-43-27

quozl:
Looking at _copy_to_journal function, my guess is that the object has no formats, but you might add temporary debugging prints to the function to gather more useful information about your problem.

samdroid-apps:
You'll need to make a special case in the clipboard menu code for pixbufs - as they have their own functions for saving as an image file (jpeg, etc).

erilyth:
@samdroid-apps Do you mean to save the image using a Gtk/Gdk image save function like www.pygtk.org/pygtk2reference/class-gdkpixbuf.html#method-gdkpixbuf--save ?
In the _write_to_temp_file function in clipboard menu code, instead of just doing a normal os.write should I use the pixbuf save function?
I tried this but it was not saving anything (just like before)

samdroid-apps:
Yep, you need to use a function like http://lazka.github.io/pgi-docs/index.html#GdkPixbuf-2.0/classes/Pixbuf.html#GdkPixbuf.Pixbuf.savev if it is a pixbuf.

erilyth:
Refer to new PR, https://github.com/sugarlabs/sugar/pull/678

PR #641, opened 2015-12-22
Added background for ViewSource's ImageBox

ezequielpereira:
- Added a grid background for ViewSource's ImageBox, so they are visible even when they are completely white
- This is a GCI 2015 task: https://codein.withgoogle.com/tasks/5721811266306048/
- Depends on: https://github.com/sugarlabs/sugar-artwork/pull/85
- Bug ticket 4909: https://bugs.sugarlabs.org/ticket/4909

### DIFF
--- a/src/jarabe/view/viewsource.py
+++ b/src/jarabe/view/viewsource.py
@@ -787,8 +787,7 @@ class SourceDisplay(Gtk.ScrolledWindow):
     file_path = property(_get_file_path, _set_file_path)
 
     def _show_image_viewer(self, icon=None, image=False):
-        media_box = Gtk.EventBox()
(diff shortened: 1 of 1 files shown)


quozl:
Forgot to mention bug 4909, which contains an easy test case.

163gal:
Added to the description.

quozl:
Thanks.  Cherry-picked for OLPC builds.  Don't see why this isn't merged by Sugar Labs.

i5o:
Okay it does not work on fedora 25, maybe due to Gtk changes, but I'll merge it anyway.

@quozl reviewed, and it worked for me in older versions of fedora/gtk.

PR #642, opened 2015-12-23
Fixed Ad-hoc icon mixup on the frame

ezequielpereira:
- Non-Sugar Ad-hoc networks are shown like access points
- Fixes ticket 3006: https://bugs.sugarlabs.org/ticket/3006
- This is a GCI 2015 task: https://codein.withgoogle.com/tasks/6255669527183360/

### DIFF
--- a/extensions/deviceicon/network.py
+++ b/extensions/deviceicon/network.py
@@ -536,7 +536,7 @@ class WirelessDeviceView(ToolButton):
         else:
             state = network.NM_DEVICE_STATE_UNKNOWN
 
-        if self._mode != network.NM_802_11_MODE_ADHOC and \
+        if self._mode != network.NM_802_11_MODE_ADHOC or \
                 network.is_sugar_adhoc_network(self._ssid) is False:
             if state == network.NM_DEVICE_STATE_ACTIVATED:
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Your commit message title is not descriptive, it should be "Show non-sugar ad-hoc networks as ad-hoc networks (not access points)"

You also haven't addressed the issue where there are no icons for ad-hoc on channel 8, etc.

163gal:
My changes make the non-sugar ad-hoc networks appear just like normal access points, I don't think your title is suitable.
The Ad-hoc on channel 8 (For example) will just have an icon like the normal access points.

samdroid-apps:
Cool, how should I test this then?

163gal:
Create an Ad-hoc network on channel 8 using another computer, then try to connect to it from Sugar.

i5o:
Can anyone test this? I only have one laptop for test :-/

quozl:
Can be closed.

Was already committed as b86b623fa6ef53c6317f95edaec5bb3412cb8aa8.  Tested, and does fix both issues in ticket 3006.

quozl:
Yay, thanks for responding.

PR #646, opened 2016-01-02
Add forward and back buttons in help view, fixes #4913

samdroid-apps:
This allows the user to navigate, so that they do not get lost
if they click a hyperlink that takes them off the social help
website.  It also helps users feel safe if navigating the help
documentation, as they will not get lost.

https://bugs.sugarlabs.org/ticket/4913

### DIFF
--- a/src/jarabe/view/viewhelp.py
+++ b/src/jarabe/view/viewhelp.py
@@ -169,11 +169,11 @@ class ViewHelp(Gtk.Window):
         self.connect('hide', self.__hide_cb)
         self.connect('key-press-event', self.__key_press_event_cb)
 
-        toolbar = Toolbar(title, has_local_help)
-        box.pack_start(toolbar, False, False, 0)
-        toolbar.show()
-        toolbar.connect('stop-clicked', self.__stop_clicked_cb)
-        toolbar.connect('mode-changed', self.__mode_changed_cb)
+        self._toolbar = Toolbar(title, has_local_help)
+        box.pack_start(self._toolbar, False, False, 0)
+        self._toolbar.show()
+        self._toolbar.connect('stop-clicked', self.__stop_clicked_cb)
+        self._toolbar.connect('mode-changed', self.__mode_changed_cb)
 
         session = WebKit.get_default_session()
         cookie_jar = SoupGNOME.CookieJarSqlite(
@@ -186,6 +186,7 @@ class ViewHelp(Gtk.Window):
         self._webview.set_full_content_zoom(True)
         self._webview.connect('resource-request-starting',
                               self._resource_request_starting_cb)
+        self._toolbar.bind_webview(self._webview)
 
         scrolled_window = Gtk.ScrolledWindow()
         scrolled_window.add(self._webview)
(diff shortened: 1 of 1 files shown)


quozl:
Is a new feature that fixes a bug.  Reviewed.

samdroid-apps:
@quozl yes, it is a feature from a ui perspectve.  However it is otherwise a very self contained small patch, so maybe it should not be t

samdroid-apps:
Trying again with the realization that close and comment is not the same as clear:

@quozl yes, it is a feature from a ux perspective.  However it otherwise a small, self contained patch, so maybe in the release perspective it should be treated as not a feature.  I do not know, and this is probably a matter for @tchx84.

godiard:
+1 from my part.
Just one comment: what happen with the buttons sensitive state when the user change from help to social help, and back?

samdroid-apps:
It's just the same webview, so the history is the same.

Maybe I should keep separate history for social/non help.  Or maybe just change the social/not RadioToolButton state based on if the user is in normal help, or the internet?

samdroid-apps:
Ah, I just force pushed an update that changes which radio tool button is selected as the user navigates.

It gets into weird cycles though. eg like:

1.  Open help
2.  Go to social help
3.  Wait for social help load
4.  Go back twice (to man page help)
5.  Go forwards
6.  Infinite weird cycle thing

godiard:
@samdroid-apps, yes, is expected. I only can imagine two solutions: if a api is available, reset the history when change from help to social help. If that is not possible, use two different webview widgets.

samdroid-apps:
@godiard good idea.  I ported it to use the back forward list to save/restore state, it works better now!  Please feel free to test.

In other news, this reminded me that WebKit2 session state apis got merged [1]!  I'll prepare some patches when new webkit2 builds flow into rawhide, but it looks good for the webkit2 port.  We could probs even get backwards compatible (loading back forward lists) in browse, because it is just a gvariant or something for the session state (which we can construct).  I'll test.

[1]  http://webkitgtk.org/reference/webkit2gtk/unstable/WebKitWebView.html#webkit-web-view-get-session-state

tchx84:
I tested it, but something seems broken:

```
  File "/home/tch/Devel/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/view/palettes.py", line 154, in __view_help__cb
    setup_view_help(self._home_activity)
  File "/home/tch/Devel/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/view/viewhelp.py", line 137, in setup_view_help
    viewhelp = ViewHelp(activity, window_xid)
  File "/home/tch/Devel/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/view/viewhelp.py", line 201, in __init__
    self._help_state = ([self._help_url], 0)
AttributeError: 'ViewHelp' object has no attribute '_help_url'
1453548660.591153 DEBUG root: NotificationBox.__destroy_cb
```

Is there another patch I should add?

samdroid-apps:
Thanks for the testing @tchx84, I did not handle the case where there was no local help available for the activity.

I have force pushed a fix.  Now tested against when both social or local help are unavailable.

davelab6:
What is the next action for this PR?

samdroid-apps:
It needs to be tested and reviewed by somebody other than myself.

PR #647, opened 2016-01-09
Fixed untranslated languages on My Settings

ezequielpereira:
- Now the languages names are translated to the language they represent
- Depends on sugarlabs/sugar-build#37
- Fixes SL#4449
- This is a GCI 2015 task: https://codein.withgoogle.com/tasks/4902925285457920/

### DIFF
--- a/extensions/cpsection/language/view.py
+++ b/extensions/cpsection/language/view.py
@@ -23,6 +23,8 @@ from gi.repository import Gdk
 from gi.repository import GObject
 import gettext
 
+from icu import Locale
+
 from sugar3.graphics import style
 from sugar3.graphics.icon import Icon
 
@@ -50,6 +52,7 @@ class Language(SectionView):
         self._labels = []
         self._language_dict = {}
         self._country_dict = {}
+        self._locale_code_dict = {}
         self._language_buttons = []
         self._country_buttons = []
         self._language_widgets = []
@@ -66,6 +69,7 @@ class Language(SectionView):
             if language not in self._language_dict:
                 self._language_dict[language] = _translate_language(language)
                 self._country_dict[language] = [[code, country]]
+                self._locale_code_dict[language] = code[:2]
             else:
                 self._country_dict[language].append([code, country])
 
@@ -132,16 +136,20 @@ class Language(SectionView):
         language_palette = []
         key_list = self._language_dict.keys()
         for language_key in sorted(key_list):
+            locale = Locale(self._locale_code_dict[language_key])
+            name = locale.getDisplayName(locale).title()
             language_palette.append(
-                {'label': self._language_dict[language_key],
+                {'label': name,
                  'index': len(self._language_buttons),
                  'callback': self._language_changed})
 
         new_language_widget = set_palette_list(language_palette)
         if locale_language is None:
             locale_language = 'English'
+        locale = Locale(self._locale_code_dict[locale_language])
+        name = locale.getDisplayName(locale).title()
         new_language_button = FilterToolItem(
-            'go-down', 'go-up', locale_language, new_language_widget)
+            'go-down', 'go-up', name, new_language_widget)
         country_list = self._build_country_list(locale_language)
 
         new_country_widget = set_palette_list(country_list)
@@ -311,7 +319,9 @@ class Language(SectionView):
         i = item['index']
 
         for language_key in self._language_dict.keys():
-            if self._language_dict[language_key] == item['label']:
+            locale = Locale(self._locale_code_dict[language_key])
+            name = locale.getDisplayName(locale).title()
+            if name == item['label']:
                 new_country_list = \
                     self._build_country_list(language_key, idx=i)
                 break


samdroid-apps:
How will this generate a list of lang names to be translated and put on potlee?

ezequielpereira:
It won't... Maybe I misunderstood the task?

samdroid-apps:
I think so.  My interpretation has been to display the language names as they would be in their language.  Eg. Don't display Indonesian or Spanish (current behaviour), say Bahasa Indonesia or Espanol. the ticket has a good comparison of data sources for this.

extensions/cpsection/language/view.py
```diff
@@ -132,16 +134,20 @@ def _add_row(self, locale_code=None):
         language_palette = []
         key_list = self._language_dict.keys()
         for language_key in sorted(key_list):
+            locale = Locale(self._country_dict[language_key][0][0][:2])
```
samdroid-apps: `[language_key][0][0][:2]` woah!

What does this mean?

Can we use one of the variables like locale_code?  Or can we give this a name?

samdroid-apps:
I have not tested but this appears to fix the ticket.

Adding the pyicu package as a dep is not an issue size wise, tested on clean SoaS22 iso which said it added less than a MB installed and did not have any sub-deps (only the pyicu package was needed).  It is also within reason; discourse on the ticket found it to be fastest and most extensive list of languages.  Do you foresee any issues with this new dep (@Quozl?)

Is it too late in the cycle to add a new dep?  (@tchx84?)

extensions/cpsection/language/view.py
ezequielpereira: It was a workaround to obtain the language code from the language name (i.e. 'es' from 'spanish').
Now I created a dictionary with all the codes (Just for clarity).

i5o:
Sam, can you please review this task after the patch is merged?
https://codein.withgoogle.com/dashboard/task-instances/6331956593688576/?sp-page=1

tchx84:
@ezequielpereira, @samdroid-apps, a couple questions:

1. What is _the_ problem here? I mean, why aren't these names being translated? What would be the solution without using pyICU?
2. If there is no other way, have you checked that the proper version of pyICU is packaged for Fedora>=18 and Ubuntu>=14.04?
3. if you look at [1] you will see that using pyICU was already considered, but it was not chosen as the solution. Have you checked what would be the cost of using it?

[1] https://bugs.sugarlabs.org/ticket/4449

ezequielpereira:
@tchx84:
1. They aren't being translated because they have no translation in Sugar .po files. A solution without using PyICU could be adding the translations to Sugar .po files.
2. PyICU 1.4 is available for Fedora >= 18, PyICU 1.5 is available for Ubuntu >= 14.04. The latest version of PyICU is 1.9.2, although for this fix the available versions are enough.
3. The cost of using PyICU is a new dependence (pyicu itself), with an installed size of less than 700Kb. I'm not sure why wasn't it chosen as the solution, it is small and the fastest.

tchx84:
@samdroid-apps @ezequielpereira my feeling is that PyICU does a lot more than just "translate" these strings properly, it is probably a whole i18n system by itself (parallel to gettext). That might be what @manuq meant when he said [1]:

    "Using ICU is a radical change and takes more resources and adds dependencies."

If I am correct, we should not add a whole new i18n system just for one bug that can be fixed in other ways.

[1] https://bugs.sugarlabs.org/ticket/4449#comment:16

ezequielpereira:
@tchx84 It should be possible then to create the necessary .po files (I don't know how to do that though). Or I can make a dictionary in  extensions/cpsection/language/view.py with all the languages translated.

tchx84:
@ezequielpereira @samdroid-apps Yes, in fact it has been done in the past [1].

[1] https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/speech.py

ezequielpereira:
@tchx84 I will do that if you think it is better (So we don't need to use PyICU at all).

walterbender:
+1 to including the language names. But we should document this somewhere
so that when new languages are added we can know to update this file. The
good news is that the worst case is that the fall back is to not have the
strings translated -- our current situation.

On Mon, Jan 11, 2016 at 9:33 AM, Ezequiel Pereira <notifications@github.com>
wrote:

> @tchx84 <https://github.com/tchx84> I will do that if you think it is
> better (So we don't need to use PyICU at all).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sugarlabs/sugar/pull/647#issuecomment-170569301>.
>



-- 
Walter Bender
Sugar Labs
http://www.sugarlabs.org

godiard:
Remember we already included a list of languages in sugar-toolkit-gtk3,
used by the text to speech code:

https://github.com/sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/speech.py#L49

On Mon, Jan 11, 2016 at 11:53 AM, Walter Bender <notifications@github.com>
wrote:

> +1 to including the language names. But we should document this somewhere
> so that when new languages are added we can know to update this file. The
> good news is that the worst case is that the fall back is to not have the
> strings translated -- our current situation.
>
> On Mon, Jan 11, 2016 at 9:33 AM, Ezequiel Pereira <
> notifications@github.com>
> wrote:
>
> > @tchx84 <https://github.com/tchx84> I will do that if you think it is
> > better (So we don't need to use PyICU at all).
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/sugarlabs/sugar/pull/647#issuecomment-170569301>.
> >
>
>
>
> --
> Walter Bender
> Sugar Labs
> http://www.sugarlabs.org
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sugarlabs/sugar/pull/647#issuecomment-170577907>.
>



-- 
Gonzalo Odiard

ezequielpereira:
Those languages are translated like "Spanish" to "Español" and "Greek" to "Ελληνική"? That's what we want to do.

godiard:
One issue is that the languages names are taken from espeak, then some are
strange like English_wmids or Vietnam_sgn

On Mon, Jan 11, 2016 at 12:16 PM, Ezequiel Pereira <notifications@github.com
> wrote:

> Those languages are translated like "Spanish" to "Español" and "Greek" to
> "Ελληνική"? That's what we want to do.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sugarlabs/sugar/pull/647#issuecomment-170584093>.
>



-- 
Gonzalo Odiard

ezequielpereira:
My idea was to make a dict like this one: https://gist.github.com/ezequielpereira/2aee4f3fa1f6a37cd68a

godiard:
@ezequielpereira, then you show every language in they own language?

ezequielpereira:
@godiard Yes, because if they are in English, the people could not know which one to choose because it could be written different (For example, Spanish in Spanish is "Español", it is very different from the English spelling).

godiard:
@ezequielpereira, have sense. Probably is a good idea for GCI, and could be included in the next Sugar version ( @tchx84 can confirm, but I think is too late for 0.108)

FGrose:
Why assume that the user can read or understand the language of the Sugar instance they receive or first start to use?

Would it not be better to assume that the user desiring to set the interface to a different language will recognize that language written in the native script and spelling?

tchx84:
@godiard  @ezequielpereira String freeze due is on Friday [1], so it will depend on the complexity of the patch and the quality of our review.

[1] http://wiki.sugarlabs.org/go/0.108/Roadmap

ezequielpereira:
I can add the dictionary with the languages translated in their own languages, but I don't know if you all agree with that or want to do something different.

godiard:
@ezequielpereira, @tchx84, in this case, where there are changes visible for the user, I think would be good discuss the proposal in the list. I agree with the proposal, but @FGrose can disagree. Of course it is possible work in a patch to try, but would not be good include a change like this at the last moment without a proper discussion. Not all the changes are Features, but this one looks as a Feature.

ezequielpereira:
@godiard Okay, then, should I close the PR's and wait for a resolution?

godiard:
No need to close.

FGrose:
Consider the situation where you receive or step up to a computer where the language had been changed to a script you could not recognize.  How would you find your native or another language?

Consider also how Wikipedia presents alternate languages, https://secure.wikimedia.org/wikipedia/en/wiki/Main_Page.

We try to follow that pattern in the Sugar Labs wiki sidebar, https://wiki.sugarlabs.org/go/Welcome_to_the_Sugar_Labs_wiki,
and also center align the names so there is no bias for left-to-right or right-to-left scripts.

samdroid-apps:
Yep, this change missed the feature freeze, however we could target it for this release.  Maybe start a mailing list thread to document the proposed approach and feature?

davelab6:
Was a mailing list thread started?

i5o:
ping

quozl:
As @FGrose pointed out; one of the purposes of the language list is so you can find your own language, and the best way to do that is using your own language.  I'm frequently failing to find English and tend to fix it by deleting the .i18n file and restarting.  Ingles isn't what works for me.  I think of it as a brand of fridge.  :grin:

yatinmaan:
I was looking around and found [this small library](https://github.com/janpipek/iso639-python) (~20KB) which is almost perfect for this use-case. 
Its just a text file and an python wrapper to get the data from file.
```python
import iso639

iso639.to_name('sv')
u'Swedish'

iso639.to_native('sv')
u'svenska'
```

**BUT**

[The data](https://github.com/janpipek/iso639-python/blob/master/iso639/languages_utf-8.txt) is a little incomplete and stored in a weird-flexible-format.
```
ISO-693-2/B | ISO-693-2/T | ISO-693-1 | NAME | NATIVE-NAME
aar         |             |aa         | Afar |Afaraf
spa         |             |es         |Spanish; Castilian|español
arc         |             |           |Official Aramaic (700-300 BCE); Imperial Aramaic (700-300 BCE)|            
```

So , would it be better to include the whole library and push any improvements upstream or re-factor its data (MIT License) for our own use (say as a dictionary) and implement the interface ourselves ?

samdroid-apps:
On Tue, 2017-01-17 at 09:23 -0800, yatinmaan wrote:
> I was looking around and found this small library (~20KB) which is
> almost perfect for this use-case.
> Its just a text file and an python wrapper to get the data from file.
> import iso639
> 
> iso639.to_name('sv')
> u'Swedish'
> 
> iso639.to_native('sv')
> u'svenska'
> BUT
> The data is a little incomplete and stored in a weird-flexible-
> format.
> ISO-693-2/B | ISO-693-2/T | ISO-693-1 | NAME | NATIVE-NAME
> aar         |             |aa         | Afar |Afaraf
> spa         |             |es         |Spanish; Castilian|español
> arc         |             |           |Official Aramaic (700-300
> BCE); Imperial Aramaic (700-300 BCE)|            
> So , would it be better to include the whole library and push any
> improvements upstream or re-factor its data (MIT License) for our own
> use (say as a dictionary) and implement the interface ourselves ?

The data is not MIT licensed in this project.  According to the readme,
it is from US Library of Congress data:

http://www.loc.gov/standards/iso639-2/php/code_list.php

It looks very complete to me (from the table on the US LOC site).  The
only column with gaps is the "ISO 639-1 Code" which is the 2 letter
language code.  I assume that is because not every language has a 2
letter code?

> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
-- 
Thanks,
Sam

https://www.sam.today/

Latest blog post: How they track you: Email Service Provider Edition - http://localhost/blog/how-they-track-you:-email-service-provider-edition/

yatinmaan:
Sorry, I thought the library's licence applied to the data as well. The native names thought are from Wikipedia so are licensed under CC-SA.

i5o:
Isn't this the chance to build our own "language" module?

Like sugar.misc.language that have functions to get current language,
and set current language etc.
AFAIK developers then to use os environ or something like that to get
current language, in activities.

On 1/18/17, yatinmaan <notifications@github.com> wrote:
> Sorry, I thought the library's licence applied to the data as well. The
> native names thought are from Wikipedia so are licensed under CC-SA.
>
> --
> You are receiving this because you commented.
> https://github.com/sugarlabs/sugar/pull/647#issuecomment-273430629


-- 
Ignacio Rodríguez

yatinmaan:
@i5o Actually gettext provides a lot of that functionality.

gettext.textdomain() returns the current/global domain/.po file being used.
And
gettext.textdomain(other_domain) can be used to change the domain.
So if that is all we need I think it would be better to document gettext as a way get & set languages in the developer docs.
Infact we can create an custom domain an use that to "translate" language names into their native names in a similar way to how we use
```python
gettext.dgettext("iso_639",msg) #msg is the language code
```
to "translate" language codes (en) to language names (English)

yatinmaan:
So I created an native.po (in sugar/po/) with 
```
msgid "Spanish"
msgstr "Espanol"

msgid "Hindi"
msgstr "हिन्दी"
```
and used gettext to "translate" language names to their native names.
```python
locale = gettext._localedirs[gettext.textdomain()] #gets the locale of current domain
def _language_native(msg):
    native = gettext.translation(gettext.textdomain(),locale,languages=['native']) #gets translation from native.po
    res = native.gettext(msg)    
    return res
```


I choose this way so that translators can themselves add native name translations when future languages are added but I don't know if this advantage is big enough to warrant such an 'hacky' solution and wanted to know if this approach is acceptable.

**Some Questions:**
1. Is this the right platform to discuss this bug or should I do it on the mailing list ?

2.1 How should I arrange the translated names, right now  हिन्दी(Hindi) is placed among other 'H' names and Espanol is placed among other 'S' names (ie. in the place of Spanish)?

2.2 Wikipedia,it seems, sorts names in order of popularity/number of articles, can/should we do something like that?

Thanks,
Yatin Maan

i5o:
1, sure we can talk here
2.1, I think we should sort the languages by their name, I mean, the translated one (Espa;ol instead of Spanish).
2.2, It's possible, but that would be hard-coded. (Probably English and Espa;ol at the top)

quozl:
@yatinmaan: 1: yes.  2.1: sorted.  2.2: yes, with a toggle.

I'm also fine with discussion here; but this is a pull request from _over a year ago_ and the requirements or priorities may have changed since.

Sorting is very helpful for locating languages in a list; but it assumes enough literacy about alphabetic order, and you will also have to choose which alphabetic order to use; which depends on language.

Sorting by popularity is unfair to minorities; but a separate list of previously used or likely language choices could be provided instead.  Some UX have provided a map to limit choices, but that's a whole different problem.

For OLPC, I'm not very concerned about the language list feature because with the new laptops we're producing we are pre-loading Ubuntu, which asks for language during first boot and removes the unselected languages.  The language list in Sugar has very little use after that happens.  So it would appear the feature is more Fedora focused.

The ticket [#4449](https://bugs.sugarlabs.org/ticket/4449) is very out of date; hasn't been updated with some of the discussion in this pull request, so it would be a good idea to work toward consensus of requirements and open a new ticket, closing the old one.  There might be another ticket already.

The consensus seems to be; for each language, show the name in that language, and show the name in the configured language.

yatinmaan:
Using [this file](https://gist.github.com/anonymous/f06cd9d15899c275a171bb0460b49968) for translations (native names from wikipedia), I get 
sc3

### There's an issue here:

Some glyphs are missing here as well as in the logger activity
sc

BUT these glyphs **do** appear when the log is viewed in gedit on ubuntu, so maybe sugar is missing these fonts or something. What should I do about this?

### As for sorting the language list : 

- If we sort them using alphabetical order then how do you order the languages that don't use the Latin alphabet ?
 - If we use [name in current language]([native name]) the "Ingles(English)" issue mentioned by @quozl above still persists.
- Maybe we should order them by popularity and implement a search bar at the top ?

**PS** : is https://bugs.sugarlabs.org active ?

Thanks,
Yatin  Maan

quozl:
Missing glyphs will be missing fonts on your system; you'll have to find an appropriate font.  On the other hand, to avoid having every font installed, you might detect missing font somehow and render the boxes as blank.

Yes, you can't order the languages that don't use the alphabet being used for ordering.  A recursive conflict.  Put them at the end of the list.

I don't understand what you mean about "Ingles(English)"; the name of English in English is English.

Yes, popularity is also useful, but is a different view.  As tabs?

Yes, bugs.sugarlabs.org is active; I've just tested it using my account.  What's your observation?

yatinmaan:
- If sugar is providing an language as an option wouldn't the fonts already exist ?

- Ingles(English) was supposing that the current system language is Spanish, **ie.** spanish-name(native-name) , but on second thought this system doesn't seem very useful as the languages positions will vary based on current system language.

- How about sorting them by their English names and showing both English and native names ?
ie. native-name(english-name)
sc4

- Arabic seems to have automatically oriented itself in a right-to-left format, is this coincidental, is there   need to center-align the names as suggested by someone above ?

- Is seems that most of the bugs on bugs.sugarlabs.org were reported 4-7 years ago, is it in use or am I searching wrong ?

- I notices another bug here, on opening the languages palette the window expands and moves the tick and cross button off-screen as can be seen in the screenshots below, should I open an issue here or report it on bugs.sugarlabs.org ?

sc2virtualbox_ubuntu_22_01_2017_21_17_49
sc3

Thanks,
Yatin Maan

quozl:
> * If sugar is providing an language as an option wouldn't the fonts already exist ?

I don't see why.

> * Arabic seems to have automatically oriented itself

Yes, Arabic is a right-to-left language implementation.

> most of the bugs on bugs.sugarlabs.org were reported 4-7 years ago, is it in use

Use the Timeline and the bugs mailing list to assess whether the service is in use enough for your needs.  Yes, the number of bug reports has declined precipitously, and the causes are less testing and an approval filter for creating tickets.

> open an issue here or report it

Issues feature of GitHub is not enabled for this repository.  Use bugs.sugarlabs.org.  See related bug [#4869](https://bugs.sugarlabs.org/ticket/4869) and the consequent changes in git.  You are welcome to fix the problem.  I'm not interested in fixing it myself because we don't plan to need this feature on our Ubuntu-based laptops, as I mentioned earlier.

chimosky:
Closed in #961.

PR #649, opened 2016-01-27
Fixed some russian strings and translated new ones.

rootacat:
Hi everybody,

I really like this project and I'd like to improve it further.

### DIFF
--- a/po/ru.po
+++ b/po/ru.po
@@ -108,8 +108,8 @@ msgid ""
 msgstr ""
 "Sugar это графический интерфейс пользователя, который вы сейчас видите. Это "
 "свободный программный продукт, защищенный лицензией GNU General Public "
-"License и вы можете свободно изменять и/или распространять его так как "
-"определено лицензией приведенной здесь."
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Hi @rootacat!  Thanks for contributing the translations.

Usually, people just do translations via http://translate.sugarlabs.org/ which is like a web interface for translations of sugar and sub-projects.

While it is preferable that T.SL.O is used in the future, I don't see a reason that this patch could not be merged?  @tchx84, I can test that the syntax works (aka, sugar still starts up in Russian language) if you're ok with the idea of merging a manual (modification of the PO files) translation patch?

tchx84:
@samdroid-apps go ahead, I am going to work on pootle this weekend. This patch is relatively small anyway, if anything happens we should be able to fix it. Double check random translations with translate.google.com?

rootacat:
Hi again, @samdroid-apps and tchx84! Nice to meet you! I didn't know about T .SL. O, but it is perfect idea! I don't need any translator, because my native language is Russian. In the future, I'll use T .SL. O.
Thank you!

PR #650, opened 2016-01-28
Set default collaberation model to salut (no jabber server)

samdroid-apps:
This allows for better activity support on newer systems, better
speed and privacy amongst other reasons.  It will not break
exsisting XO or schoolserver workflows.

The changes was discussed in [1].

[1]  http://lists.sugarlabs.org/archive/sugar-devel/2016-January/051187.html

@tchx84 is this ok for 108?

### DIFF
--- a/data/org.sugarlabs.gschema.xml
+++ b/data/org.sugarlabs.gschema.xml
@@ -165,7 +165,7 @@
     </schema>
     <schema id="org.sugarlabs.collaboration" path="/org/sugarlabs/collaboration/">
         <key name="jabber-server" type="s">
-            <default>'jabber.sugarlabs.org'</default>
+            <default>''</default>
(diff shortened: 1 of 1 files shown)


davelab6:
@tchx84 ping :)

samdroid-apps:
Reading through the ML thread it seems that there was agreement to merge this?  Would you agree with that reading @davelab6?

If so, I'll merge it before 0.109.0.0

davelab6:
Yes, that's my reading also. 

Great ideas from @quozl in http://lists.sugarlabs.org/archive/sugar-devel/2016-January/051200.html that we should be tracking in the issue tracker

samdroid-apps:
Didn't merge before 0.109.0.0, but it will be in 0.109.0.1!

PR #652, opened 2016-02-18
Implemented toggled sorting order

AbrahmAB:
For Sort view, I observed that it sorts in only one order (i.e. for sort by size its in descending order of size of journal entries, the list remains in descending order of sizes!) .So I added a feature of toggling the sorting order as well (like take for size on first click it shows list in descending order on second click list reverses to ascending order).

### DIFF
--- a/src/jarabe/journal/journaltoolbox.py
+++ b/src/jarabe/journal/journaltoolbox.py
@@ -695,6 +695,7 @@ class SortingButton(ToolButton):
         ToolButton.__init__(self)
 
         self._property = 'timestamp'
+        self.order_type='ascending'
         self._order = Gtk.SortType.ASCENDING
 
         self.props.tooltip = _('Sort view')
@@ -719,6 +720,7 @@ class SortingButton(ToolButton):
                                icon_name=icon)
             button.set_image(button_icon)
             button_icon.show()
+            
             button.connect('activate',
                            self.__sort_type_changed_cb,
                            property_,
@@ -727,9 +729,20 @@ class SortingButton(ToolButton):
             menu_box.append_item(button)
 
     def __sort_type_changed_cb(self, widget, property_, icon_name):
+        if(self._property == property_):
+              if(self.order_type=='ascending'):
+                   self.order_type='descending'
+              else:
+                   self.order_type='ascending'
+        else:
+             self.order_type='ascending'
+
         self._property = property_
-        # FIXME: Implement sorting order
-        self._order = Gtk.SortType.ASCENDING
+       
+        if(self.order_type=='ascending'):
+                   self._order = Gtk.SortType.ASCENDING
+        else:
+                   self._order=Gtk.SortType.DESCENDING
         self.emit('sort-property-changed')
 
         self.props.icon_name = icon_name
--- a/src/jarabe/journal/journaltoolbox.py
+++ b/src/jarabe/journal/journaltoolbox.py
@@ -693,17 +693,13 @@ class SortingButton(ToolButton):
 
     def __init__(self):
         ToolButton.__init__(self)
-
         self._property = 'timestamp'
         self.order_type='ascending'
         self._order = Gtk.SortType.ASCENDING
-
         self.props.tooltip = _('Sort view')
         self.props.icon_name = 'view-lastedit'
-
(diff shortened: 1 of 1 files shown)


src/jarabe/journal/journaltoolbox.py
```diff
@@ -727,9 +729,20 @@ def __init__(self):
             menu_box.append_item(button)
 
     def __sort_type_changed_cb(self, widget, property_, icon_name):
+        if(self._property == property_):
+              if(self.order_type=='ascending'):
+                   self.order_type='descending'
+              else:
+                   self.order_type='ascending'
+        else:
+             self.order_type='ascending'
```
samdroid-apps: Can you just directly set the _order property rather than adding the order_type property.

samdroid-apps:
Yeah, this works well.  I see what you were talking about with the indicator though, this may feel a bit confusing if the sort order randomly reverses.

In addition to the code comment, please run flake8 on your patch.  Flake8 will check for style guideline conformance.

While the click again to switch the sorting is nice and consistent with other operating systems, it is also not discoverable.   Maybe have a look into adding a check box with a label "reverse order" at the bottom of the palette, or maybe add radio buttons like the homeview type selector palette [1].  If you need help getting started with Gtk+, have a look around the other bits of the sugar code or maybe look at this good tutorial [2].

[1]  [https://gyazo.com/981f7f4efb22b53ce71edc60c75d23d5](https://gyazo.com/981f7f4efb22b53ce71edc60c75d23d5)
[2] http://python-gtk-3-tutorial.readthedocs.org/en/latest/button_widgets.html#togglebutton

AbrahmAB:
Adding a "Reverse Order" toggle button in Sort By palette seems a good idea.  I will make the required changes shortly.

AbrahmAB:
Done! 
I have used Gtk.ToggleButton for adding  "Reverse Order" at the bottom of the palette and checked for indentations and whitespaces using flake8. Can I send a patch now?
toggle_reverse

FGrose:
> While the click again to switch the sorting is nice and consistent with other operating systems, it is also not discoverable.

For the Journal sort, the up and down pointed triangles are a strong hint that two directions of sort order are possible.  I find the extra menu item distracting from the cleanliness/efficiency/simplicity of Sugar.  It also introduces an unnecessary translation burden.

For the Home View, this change feels like "featuritis" or "feature creep", https://en.wikipedia.org/wiki/Feature_creep as we already have a search box and the list of Activities is generally short.  If the Sugar learner doesn't discover the effect of a secondary invocation of the same view style, not much is lost.

What may be more rewarding, would be a feature to allow reordering of Activities in the circle or list views (like grouping is possible in the freeform view).  This may serve to conveniently group associated Activities in the visual plane.

samdroid-apps:
I am actually convinced by you @FGrose that this is reasonably intuitive and not too important if it is missed.  We can add it to the help activity I suppose :)

So sorry to distract you in the direction of a toggle button @AbrahmAB.  However, please fix up the other comments on this patch so that I can merge it (namely, directly use the _order variable rather than introducing a new one)

AbrahmAB:
Okay sure I will remove the toggle button and again send the patch :)

samdroid-apps:
Sorry, I didn't see that you pushed a new commit.  It looks like it fixes the variable issue.  Just 2 issues:

1.  You have a merge conflict in the commit :(  Feel free to make a new fresh branch, copy and paste changes over and start a new pull request if needed.
2.  Please don't remove the spacing in the `__init__`

AbrahmAB:
Okay @samdroid-apps I will make another pull request. Not much familiar with how git works :( . So can I close this pull request to make a proper pull request with other branch?

samdroid-apps:
Oh, you did this on master.  So first, make a backup of your code so that you don't loose it.

Then delete the sugar directory, and clone it again (`git clone https://github.com/sugarlabs/sugar`)

Now, before you make your change, go onto a new branch `git checkout -b branch-name`

Then commit your change per normal

Then push it like `git push fork branch-name` (amusing you added your fork like `git remote add fork git@github.com:you/sugar`)

AbrahmAB:
Okay I got it. Thanks @samdroid-apps  :)
Hereby I close the pull request. Sorry for my silly mistake .

PR #653, opened 2016-02-21
Implemented toggled sorting order

AbrahmAB:
As discussed with @samdroid-apps . I added a feature of toggled sorting order for list in the Journal. This will allow the list to be sorted in both ways.

### DIFF
--- a/src/jarabe/journal/journaltoolbox.py
+++ b/src/jarabe/journal/journaltoolbox.py
@@ -727,9 +727,15 @@ class SortingButton(ToolButton):
             menu_box.append_item(button)
 
     def __sort_type_changed_cb(self, widget, property_, icon_name):
+        if self._property == property_:
+            if self._order == Gtk.SortType.ASCENDING:
+                self._order = Gtk.SortType.DESCENDING
+            else:
+                self._order = Gtk.SortType.ASCENDING
+        else:
+            self._order = Gtk.SortType.ASCENDING
+
         self._property = property_
-        # FIXME: Implement sorting order
-        self._order = Gtk.SortType.ASCENDING
         self.emit('sort-property-changed')
 
         self.props.icon_name = icon_name


src/jarabe/journal/journaltoolbox.py
```diff
@@ -727,11 +724,20 @@ def __init__(self):
             menu_box.append_item(button)
 
     def __sort_type_changed_cb(self, widget, property_, icon_name):
+        if(self._property == property_):
```
samdroid-apps: Please run flake8 on this, but I don't think that it is accepted to put the parentheses in the if statement like C.

src/jarabe/journal/journaltoolbox.py
```diff
@@ -727,11 +724,20 @@ def __init__(self):
...
+        else:
+            self._order = Gtk.SortType.ASCENDING
+
         self._property = property_
-        # FIXME: Implement sorting order
-        self._order = Gtk.SortType.ASCENDING
+        if(self._order == Gtk.SortType.ASCENDING):
+            self._order = Gtk.SortType.ASCENDING
+        else:
+            self._order = Gtk.SortType.DESCENDING
```
samdroid-apps: Why do you need this?   This does nothing.

samdroid-apps:
Awsome, just a few cleaning up things to do.  Also, please send your patch in 1 commit rather than 2.

src/jarabe/journal/journaltoolbox.py
AbrahmAB: Okay I will remove the parentheses.

AbrahmAB:
All done. :)

samdroid-apps:
Thanks, but please run your code through flake8.  You still have some parenthesis around your ifs :(

Also, have a look at the diff from github.  You are still removing a few empty lines unrelated to your patch.  Please don't do this.

AbrahmAB:
I have made necessary changes again and resolved my silly mistakes.

PR #655, opened 2016-02-21
Fixed Issue: #4940 Journal entries name can be renamed to blank

AbrahmAB:
Fixed issue [#4940](https://bugs.sugarlabs.org/ticket/4940) renaming entries to blank or whitespace. Now  entry cannot be renamed to blank (or say whitespace). If tried the old name is retained.

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -747,7 +747,9 @@ class ListView(BaseListView):
 
     def __cell_title_edited_cb(self, cell, path, new_text):
         iterator = self._model.get_iter(path)
-        self._model[iterator][ListModel.COLUMN_TITLE] = new_text
+        if not new_text.isspace() and not new_text == '':
+            self._model[iterator][ListModel.COLUMN_TITLE] = new_text
+
         self.emit('title-edit-finished')
 
     def __editing_canceled_cb(self, cell):


samdroid-apps:
Looks good!  It works well.

However, please change your commit message to the title of the pull request that you used.  "Fixed Issue: #4940 Journal entries name can be renamed to blank" tells me a lot about what you did, where as "Fixed Renaming of entry" could refer to anything!

AbrahmAB:
Done! :)

src/jarabe/journal/listview.py
```diff
@@ -747,7 +747,11 @@ def __icon_clicked_cb(self, cell, path):
 
     def __cell_title_edited_cb(self, cell, path, new_text):
         iterator = self._model.get_iter(path)
-        self._model[iterator][ListModel.COLUMN_TITLE] = new_text
+        if new_text.isspace() or new_text == '':
+            return
+        else:
+            self._model[iterator][ListModel.COLUMN_TITLE] = new_text
+
         self.emit('title-edit-finished')
```
samdroid-apps: Hum, I just noticed that if you try and rename it to something empty, this signal will not be emitted.  This is an issue, eg. try:

1.  Go into name editing mode in the list view
2.  Rename to "" (empty)
3.  Press enter to confirm
4.  Try typing to search in the journal (don't click in the search entry, that should not be needed), notice that doesn't work though - the entry is not autofocused on keypress

AbrahmAB: @samdroid-apps Do you mean that if there is already existing entry it should be auto-focused on keypress?

Because according to my changes an entry cannot be renamed to empty ("") anyways. But if there is already an entry with empty name then should I make changes to search also such that it auto-focuses on empty named entries.

Is that what you are trying to suggest or there is something more to it?

samdroid-apps: You need to make sure that the "title-edit-finished" is always emitted when the textbox is dismissed.  Currently, if the text is set to "" (emtpy) the function returns and the signal is not emitted.

The signal deals with auto-focusing the search bar when you type.  To see that feature in action, open the journal and start typing - it will type into the search bar.  However, this function is broken if the title-edit-finished is not emitted, see the steps to reproduce that I provided in the previous comment.

AbrahmAB: Oh yes, I just saw the use of the signal 'title-edit-finished' in the journalactivity.py .
I will make the function emit the required signal by adding the line self.emit(..) in if block.
Hope that would be fine.

godiard: Better don't duplicate code.
Would be better change the 'if' condition to something like:
 ```
        if new_text != '' and not new_text.isspace():
             self._model[iterator][ListModel.COLUMN_TITLE] = new_text
```

AbrahmAB: Yeah, This is a better idea.
@samdroid-apps @godiard I have made the changes. Have a look

AbrahmAB:
@samdroid-apps I just found that user could save his work with no name (ie "" null)! 

Should I fix this?

AbrahmAB:
To fix this (ie Not allow user to rename or even name his/her work to empty text) we just need to add the condition to check the input text in sugar3/activity/widgets.py save_title() .

samdroid-apps:
I think that it is harder to deal with it in the activity title entry (as you talked about 3 days ago).  This is because here we have a clear feedback mechanism (the name does not change) whereas we don't in the activity title bar. Eg, if we reset it to the default ("%s activity") that would annoy users who were trying to delete the default title and add their own.  Maybe we need to think about a design for this one?

AbrahmAB:
@samdroid-apps Yes I just tried that. Its too annoying.
Couldn't this be solved by just checking for the title only when the user closes the activity ?

jouravla:
I found the ticket for this bug (https://bugs.sugarlabs.org/ticket/4940) and found that this issue still persists in the current version of Sugar.

quozl:
@jouravla, thanks, good point.

Yes, the issue persists in the detail view, because #655 was not the complete fix, and another fix was not yet merged in #663 before release.

I've updated the ticket.

I've reverted the patch in my OLPC branch.

quozl:
Revert of #655 submitted as #752.

PR #656, opened 2016-02-21
Fixed #3146: Dragging a Journal entry drags the whole white entry bar instead of icon

AbrahmAB:
Fixes issue [#3146](https://bugs.sugarlabs.org/ticket/3146) . Now only icon of entry can be dragged. 

Below is the result.
 
drag

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -23,6 +23,7 @@ from gi.repository import GObject
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import Pango
+from gi.repository import GdkPixbuf
 
 from sugar3.graphics import style
 from sugar3.graphics.icon import Icon, CellRendererIcon
@@ -191,10 +192,12 @@ class BaseListView(Gtk.Bin):
         model.deleted.connect(self.__model_deleted_cb)
 
     def enable_drag_and_copy(self):
-        self.tree_view.enable_model_drag_source(Gdk.ModifierType.BUTTON1_MASK,
-                                                [('text/uri-list', 0, 0),
-                                                 ('journal-object-id', 0, 0)],
-                                                Gdk.DragAction.COPY)
+        self.tree_view.drag_source_set(Gdk.ModifierType.BUTTON1_MASK,
+                                       [Gtk.TargetEntry.new(
+                                           'text/uri-list', 0, 0),
+                                        Gtk.TargetEntry.new(
+                                            'journal-object-id', 0, 0)],
+                                       Gdk.DragAction.COPY)
 
     def disable_drag_and_copy(self):
         self.tree_view.unset_rows_drag_source()
@@ -702,6 +705,13 @@ class ListView(BaseListView):
         return self._is_dragging
 
     def __drag_begin_cb(self, widget, drag_context):
+        path, _column = self.tree_view.get_cursor()
+        if path is None:
+            return
+
+        row = self.tree_view.get_model()[path]
+        _pixbuf = GdkPixbuf.Pixbuf.new_from_file(row[ListModel.COLUMN_ICON])
+        self.tree_view.drag_source_set_icon_pixbuf(_pixbuf)
         self._is_dragging = True
 
     def __drag_data_get_cb(self, widget, context, selection, info, time):


src/jarabe/journal/listview.py
```diff
@@ -711,6 +716,17 @@ def __drag_data_get_cb(self, widget, context, selection, info, time):
         model = self.tree_view.get_model()
         model.do_drag_data_get(path, selection)
 
+    def __button_press_event_cb(self, widget, event):
```
samdroid-apps: is the button press event the best way to do this?  Is there another event we could use?  Is there like a drag start event?

Does this mean that we will be loading the pixbuf every time I press a mouse button?  That seems expensive.

src/jarabe/journal/listview.py
```diff
@@ -711,6 +716,17 @@ def __drag_data_get_cb(self, widget, context, selection, info, time):
...
+        try:
+            path = self.tree_view.get_path_at_pos(event.x, event.y)[0]
+        except:
```
samdroid-apps: What are you excepting?  Catch-all except statements aren't the best - they hide errors

src/jarabe/journal/listview.py
```diff
@@ -711,6 +716,17 @@ def __drag_data_get_cb(self, widget, context, selection, info, time):
...
+            return
+
+        _iter = self._model.get_iter(path)
+        _model_value = self._model.get_value(_iter, 2)
```
samdroid-apps: Isn't there a COLUMN_ICON of whatever constant that you could use for here?

samdroid-apps:
Yeah, this works well.  But the solution needs some dicsussion.

Also, the icons are in just black and white.  This is the same issue as for the clipboard icons, so maybe copy it's TODO comment?

src/jarabe/journal/listview.py
AbrahmAB: So using if else instead would be preferable?

src/jarabe/journal/listview.py
AbrahmAB: Oh right! just replacing the lines
`_iter = self._model.get_iter(path)
+        _model_value = self._model.get_value(_iter, 2)`

with 

`row = self.tree_view.get_model()[path]
  _pixbuf = GdkPixbuf.Pixbuf.new_from_file(row[ListModel.COLUMN_ICON])`

would be better.

src/jarabe/journal/listview.py
quozl: No, the problem is that your *except* does not have the name of an exception, so it will hide all exceptions, including coding errors.

src/jarabe/journal/listview.py
AbrahmAB: The previously used `tree_view.enable_model_drag_source` returned the whole row to be drag. It didn't need any icon or something, the whole row would be seen lifted.

But as for using the `tree_view.drag_source_set` sets up a widget on which drag operation will start but icon wont be set automatically 

For starting the drag operation here we are using the `button-press-event`. 

So I have created another function for the `button-press-event` and also to set up an icon for indicating the drag.

src/jarabe/journal/listview.py
AbrahmAB: So should I use sys.exc_info() for returning exceptions

Or else should I use if else such that if no path found it returns back. Will that raise an exception?

Suggest me which one would be better @samdroid-apps @quozl

quozl: No, only you know which exception you are trying to handle.  Check Python language reference manuals for the try/except clause to see how to specify the exception.

AbrahmAB: Okay Got it!
The `get_path_at_pos` will return None when no path found.
I think this could be easily checked by `if path is None` .

quozl: Oh, both @samdroid-apps and I presumed you were handling an exception.  Now that I know your intent was to protect against None returned, yes, best to split the code into get of path, and extracting first.

src/jarabe/journal/listview.py
samdroid-apps: On Mon, Feb 22, 2016, 18:32 Abhijit Patel <notifications@github.com> wrote:

In src <https://github.com/sugarlabs/sugar/pull/656#discussion_r53593069>/
<https://github.com/sugarlabs/sugar/pull/656#discussion_r53593069>jarabe
<https://github.com/sugarlabs/sugar/pull/656#discussion_r53593069>/journal/
<https://github.com/sugarlabs/sugar/pull/656#discussion_r53593069>
listview.py
<https://github.com/sugarlabs/sugar/pull/656#discussion_r53593069>:

> @@ -711,6 +716,17 @@ def __drag_data_get_cb(self, widget, context,
selection, info, time): > model = self.tree_view.get_model() >
model.do_drag_data_get(path, selection) > > + def
__button_press_event_cb(self, widget, event):

The previously used tree_view.enable_model_drag_source returned the whole
row to be drag. It didn't need any icon or something, the whole row would
be seen lifted.

But as for using the tree_view.drag_source_set sets up a widget on which
drag operation will start but icon wont be set automatically

For starting the drag operation here we are using the button-press-event.

So I have created another function for the button-press-event and also to
set up an icon for indicating the drag.


Yes, but is this the best signal?  Isn't this called every time the user
interacts with the tree view, even if they are just clicking to start an
activity or something?  Can we change it so that we don't load an icon
(expensive operation?) every button event?

—
Reply to this email directly or view it on
<https://github.com/sugarlabs/sugar/pull/656/files#r53593069>GitHub
<https://github.com/sugarlabs/sugar/pull/656/files#r53593069>.

quozl:
The commit has english language error; "an journal entry" should be "a journal entry".

The commit description doesn't say what the problem is or how it is solved.

src/jarabe/journal/listview.py
AbrahmAB: Oh, That true! The function would be called even if user is just clicking to start an activity.

I just fixed this . The icon will be set after the `drag-begin` event.

AbrahmAB:
@samdroid-apps  Could I get a link from where to copy the TODO comment?

AbrahmAB:
@samdroid-apps @quozl I have made necessary changes in my previous commit .

samdroid-apps:
Awesome!  This is great, I've merged it.

Re the TODO, we don't need to add it to the code that bad on reflection.  Anyway, grep returns `src/jarabe/frame/clipboardicon.py:        # TODO: We should get the pixbuf from the icon, with colors, etc.`

quozl:
Regression on Fedora 18.

Cursor briefly has a "+" at the start of the drag, but the drag survives only a very short distance before the cursor returns to normal, and there is no floating icon following the cursor.  The "+" appears as the cursor enters the Documents icon, and the drop is successful.

Reverting the patch from my OLPC branch, which fixes the regression.

@AbrahmAB, have you any idea what might have caused this regression?  Fedora 18 on the OLPC XO uses GTK+ 3.6.4.

quozl:
@AbrahmAB, in enable_drag_and_copy, where you switched from Gtk.TreeView.enable_model_drag_source to Gtk.Widget.drag_source_set, should you also have switched from Gtk.TreeView.unset_rows_drag_source to Gtk.Widget.drag_source_unset in disable_drag_and_copy?  It is used in editing mode.  Unrelated to my previous comment.

AbrahmAB:
@quozl , The regression is caused due to this failure in creating pixbuf in [[1]](https://github.com/sugarlabs/sugar/blob/master/src/jarabe/journal/listview.py#L725) .
I tried setting the `_pixbuf` to None and got this error in shello.log
`Traceback (most recent call last):
  File "/home/broot/sugar-build/build/out/install/lib/python2.7/site-packages/jarabe/journal/listview.py", line 1063, in __drag_begin_cb
    self.tree_view.drag_source_set_icon_pixbuf(_pixbuf)
TypeError: Argument 1 does not allow None as a value`
Are you getting the same error in shell.log?
Mine distro is Ubuntu 14.04 uses Gtk+ 3.10.8.

quozl:
No, there is no such error in shell.log.  I also tried setting the icon from a fixed known path just after drag_source_set.  No error, but no change in behaviour.  It may be a bug in Fedora 18's GTK+ 3.6.4.  You might try that in a virtual machine?

davelab6:
> Reverted on OLPC branch.

@quozl where is the OLPC branch maintaned?

PR #657, opened 2016-02-22
About my computer: set copyright year 2016

quozl:
Dolphins patrolling the south pacific microplastic gyre have observed that the value returned by 

```python
import time
print time.strftime('%Y')
```

appears to have changed, and research suggests it has something to do with human tradition, yet the About My Computer text has not changed.  Rather than understand the tradition, the following patch is proposed:

```bash
sed --in-place 's/2015/2016/g' extensions/cpsection/aboutcomputer/view.py
```

Thus a rightness is brought to the world, albeit in a small way.


godiard:
Thanks dolphins!

davelab6:
lol! :D

PR #658, opened 2016-02-26
Fixed #4945: fixed the layout of alert message that appears on erasing an entry with lengthy titles

AbrahmAB:
Used Ellipsis for entries with lengthy titles instead of the whole title.

### DIFF
--- a/src/jarabe/journal/journaltoolbox.py
+++ b/src/jarabe/journal/journaltoolbox.py
@@ -581,8 +581,12 @@ class DetailToolbox(ToolbarBox):
         alert = Alert()
         erase_string = _('Erase')
         alert.props.title = erase_string
+        _erase_title = self._metadata['title']
+        if len(self._metadata['title']) > 25:
+            _erase_title = self._metadata['title'][:15] + '...'
+
         alert.props.msg = _('Do you want to permanently erase \"%s\"?') \
-            % self._metadata['title']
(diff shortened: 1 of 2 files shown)


quozl:
Your solution is not scalable to different font sizes and display widths.

Better to set the GtkLabel property *ellipsize*; search sugar and sugar-toolkit-gtk3 repositories for examples, and refer to the Python GI API reference for GtkLabel.

AbrahmAB:
Setting ellipsize in sugarlabs/sugar-toolkit-gtk3/blob/master/src/sugar3/graphics/style.py would be a better solution. 
Can I send another PR in sugar-toolkit-gtk3 to set the ellipsize in alert.py?

AbrahmAB:
It looks like the bug is in alert.py file that the message and title are not set to ellipsize. 

No need to make changes in sugar repo. Instead I would like to fix the bug in sugar-toolkit-gtk3 .
@quozl Can I close this PR now?

quozl:
Agreed, you will most likely fix the problem in sugar-toolkit-gtk3, so
when you have a new pull request add reference here and close.

AbrahmAB:
Done!
Reference to my pull request in sugar-toolkit-gtk3 [#310](https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/310).

PR #660, opened 2016-02-28
Fix frame toggle issue and journal button toggle issue

erilyth:
1)In the journal view, when we try to access the buttons on the bottom it becomes hard since the frame keeps popping up as soon as we go near the buttons.
2)Another issue in the journal view is that 'Journal' icon on the bottom left cannot be clicked on again if we move to either the 'Documents' or any of the mounted devices. This is because the frame gets called upon and the click is not detected by the button.

1)a)The frame starts popping up blocking the buttons
screenshot from 2016-02-29 00 43 11

1)b)The frame completely covers the buttons and I have to find other means to get to the buttons
screenshot from 2016-02-29 00 42 23

2)a)Select another icon,
screenshot from 2016-02-29 00 42 42

2)b)Unable to select the jounal icon again.
screenshot from 2016-02-29 00 42 48

Ensuring the frame pops up only when we reach the boundaries of the screen will solve both of these problems.

### DIFF
--- a/src/jarabe/frame/eventarea.py
+++ b/src/jarabe/frame/eventarea.py
@@ -69,8 +69,10 @@ class EventArea(GObject.GObject):
         self._edge_delay = min(settings.get_int('edge-delay'), _MAX_DELAY)
         self._corner_delay = min(settings.get_int('corner-delay'), _MAX_DELAY)
         ts = min(settings.get_int('trigger-size'), style.GRID_CELL_SIZE)
-        sw = Gdk.Screen.width()
-        sh = Gdk.Screen.height()
+        
+        #Ensure that the frame toggle gets called only when we reach the boundaries of the screen
+        sw = Gdk.Screen.width() + style.GRID_CELL_SIZE - 1
+        sh = Gdk.Screen.height() + style.GRID_CELL_SIZE - 1
 
         if self._edge_delay == _MAX_DELAY:
             self._hide(_EDGES)


quozl:
Nak.  You don't seem to have explored the problem fully.  The behaviour you describe is expected and normal.  What is your value of the trigger-size setting set by the frame control panel in my settings?  Reduce this size to 1 and try again.

erilyth:
@Quozl It was set to 1 and I did not change it

quozl:
Try setting it to 2, and reproduce the symptom.

erilyth:
@quozl I tried it, it still gives the same symptoms

quozl:
Well, it doesn't happen for me.  Let's find out more about what you are testing.  What version of Sugar, and how are you starting Sugar?  What are the values in gsettings for org.sugarlabs.frame edge-delay, corner-delay, and trigger-size?

```bash
$ gsettings list-recursively | grep frame
```

erilyth:
I am starting sugar using ./osbuild shell and then using the 'run' command.
The version of sugar I am using is 0.108

This is what I get when I run the command,
org.cinnamon.recorder framerate 15
org.gnome.gedit.plugins.filebrowser virtual-root 'file:///home/vishalapr/OpenSource/sugar-build/sugar/src/jarabe/frame'
org.cinnamon.recorder framerate 15
org.sugarlabs.frame corner-delay 0
org.sugarlabs.frame edge-delay 1000
org.sugarlabs.frame corner-delay 0
org.sugarlabs.frame edge-delay 1000

I have set the trigger size to 2.

quozl:
Since there was no trigger-size in gsettings command output, it means the schema has not been applied, and trigger-size is not being set.  You may find shell.log to contain errors setting or fetching trigger-size, perhaps resulting in src/jarabe/frame/eventarea.py using GRID_CELL_SIZE instead.

Please edit eventarea.py to add logging so that you can find what value is being stored in object *ts*:
```python
      ts = min(settings.get_int('trigger-size'), style.GRID_CELL_SIZE)
```

sugar-build is a very bad method for testing Sugar.  Please in future test without using an emulator.  Instead, use native packaging for your distribution.  Thanks.

erilyth:
@quozl you were right, its using the GRID_CELL_SIZE always, but I don't see any errors in shell.log regarding the trigger-size.

Regarding the sugar-build method, in the native packaging for my distribution (ubuntu) will it contain the latest updates that are present on this repository? I thought that only releases would be present there so I never used them before.

quozl:
On Fedora 18 errors are seen if the schema from Sugar data/org.sugarlabs.gschema.xml is not installed.  No errors for you may mean later version gsettings is silent, and that settings.get_int('trigger-size') is returning a number larger than GRID_CELL_SIZE.  To fix your system you must figure out why gsettings org.sugarlabs.frame trigger-size does not exist.

To use latest updates in native packages will require that you prepare native packages from the repository.  This is a well-known process and is complaint with GNU coding standards; first step is to generate the .tar.xz;

```bash
./autogen.sh
make distclean
```

Next steps depend on your distribution packaging methods; which for Ubuntu are the Debian methods and tools.  The result is a .deb file that can be installed.

erilyth:
@quozl I will try build using native packages from next time. 

Regarding the frame trigger, on default does your frame get triggered only when you go to the boundaries of the screen or at a certain distance from the boundaries of the screen (this is what was happening to me)?

I have sugar built and setup on Fedora 23 as well, do you think I should test it there once as well or would it make no difference?

quozl:
When I test, the trigger-size is obeyed strictly; when it is size 1, the frame does not trigger until the cursor reaches the last pixel on the screen.  You might try to find why trigger-size is not being set properly; it probably has something to do with deficiencies in sugar-build emulator.

My testing is with Fedora 18, Fedora 20, Ubuntu 14.04, and Ubuntu 15.10.  I don't really understand your second question; asking me to speculate if an unexplained problem will persist seems pointless.  You should try it yourself if you think it is important to do so.

erilyth:
I just tested it with Fedora 23 and Ubuntu 15.10 and the problem persists. I'll try find out why the trigger-size is not being set properly as you suggested.

quozl:
Please confirm your test method was sugar-build?  My test method is native packages.  Check each of your sugar-build subordinate git repositories to make sure they are tracking master branch.  You might also use "git checkout" to pin them at v0.108.0 release and test again, that way you'll find out if a patch since 0.108 is the cause.  None of the patches since 0.108 seem relevant in a review just now.

erilyth:
Yes the test method was sugar-build. I used 'git checkout' on all the git repos within the sugar-build repo but the issue still exists, so I guess you are right, the issue must have been introduced before 0.108

PR #661, opened 2016-02-29
Fix a few typos and abbreviations

erilyth:


quozl:
These are historical documents rather than current.  They are out of date in other ways.  The other problems should be fixed as well.  You wouldn't want to give the appearance that they are up to date by editing them so briefly.

erilyth:
@quozl I updated the PR as you mentioned

quozl:
I'm still unwilling to accept these changes; while your changes are valid in English language, they are mostly useless because these documents are generally not used.  These are historical documents, and you're only fixing trivial aspects, without fixing more fundamental problems.  The file controls.txt is mostly relevant to sugar-toolkit-gtk3 repository, not sugar repository, and we have since changed the method by which we write the documentation.  If there is anything in controls.txt that adds to what we have in sugar-toolkit-gtk3, then it should be merged with the source in the other repository.  To fix the more fundamental problems with docs/ will require review of all patches since each file was last changed.  You've shown no evidence of doing that, so I don't see any value yet in what you've done.  Hope that helps!

erilyth:
@quozl , I can understand. Thanks for the review.

Any fundamental changes to the docs, I think should belong to their own PR and not as typos/abbreviations so I'll close this PR.

PR #662, opened 2016-03-02
Fixed #4947: Buddy palette appears in wrong position

AbrahmAB:
Fixed [#4947](https://bugs.sugarlabs.org/ticket/4947)  . The palette was created when the cursor hovered in the buddy column even though there was no buddy icon at cursor position!

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -97,11 +97,23 @@ class TreeView(Gtk.TreeView):
             palette.connect('volume-error', self.__volume_error_cb)
 
         elif column in self.buddies_columns:
-            buddycellrenderer = column.get_cells()[0]
-            if buddycellrenderer.nick is not None:
-                palette = BuddyPalette(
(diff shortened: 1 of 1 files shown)


quozl:
Reproduced reported problem on Fedora 18 and Ubuntu 15.10 with Sugar 0.108.0.

Tested on Fedora 18 and Ubuntu 15.10; your patch does fix the problem.

Reviewed patch;
* the commit message doesn't explain the problem or the solution,
* the commented assignment of *column_index* is puzzling, as it isn't used or explained,
* the use of an index variable *i* is unfortunate, but is used twice inside the for loop, makes me wonder if there is an alternative solution that changes data structures to make it easier.

AbrahmAB:
@quozl 
Changes made in commit
- Amended the commit message.

- Removed unnecessary comment

- Implemented the loop in better way by replacing the variable _i_ with _column_index_

quozl:
Reviewed, thanks.

PR #663, opened 2016-03-07
Fixed issue: Journal Entries can be renamed to blank via Detail View

devAbnull:
Journal Entry can be renamed to blank when tried to rename via the Detail View. 
Renaming directly from the list view was solved in #655.

### DIFF
--- a/src/jarabe/journal/expandedentry.py
+++ b/src/jarabe/journal/expandedentry.py
@@ -491,7 +491,19 @@ class ExpandedEntry(Gtk.EventBox):
 
         old_title = self._metadata.get('title', None)
         new_title = self._title.get_text()
-        if old_title != new_title:
+        if new_title == '' or new_title.isspace():
+            alert = Alert()
+            alert.props.title = _('Empty title')
+            alert.props.msg = _('The title cannot be empty')
+            ok_icon = Icon(icon_name='dialog-ok')
+            alert.add_button(Gtk.ResponseType.OK, _('Ok'), ok_icon)
+            ok_icon.show()
+            alert.connect('response', self._title_alert_response_cb)
+            journalwindow.get_journal_window().add_alert(alert)
+            alert.show()
+            new_title = self._title.set_text(old_title)
+
+        elif old_title != new_title:
             self._icon.palette.props.primary_text = new_title
             self._metadata['title'] = new_title
             self._metadata['title_set_by_user'] = '1'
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Hum, this changes worries me.  While it prevents the user from changing the title to be nothing, it doesn't give them any feedback that the change has been prevented.  Maybe that is something to look into if you wish to make this change?

AbrahmAB:
@samdroid-apps showing an alert message when the 'focus-out-event' signal is emitted would work.

samdroid-apps:
Yeah, you could do that

AbrahmAB:
An alert message added when the 'focus-out-event' signal is emitted and title box is empty.

src/jarabe/journal/expandedentry.py
```diff
@@ -491,7 +491,20 @@ def _update_entry(self, needs_update=False):
 
         old_title = self._metadata.get('title', None)
         new_title = self._title.get_text()
-        if old_title != new_title:
+        if new_title == '' or new_title.isspace():
+            alert = Alert()
+            alert.props.title = _('Empty title')
+            alert.props.msg = 'The title cannot be empty'
```
samdroid-apps: Please put this string through gettext (`_('...')`)

AbrahmAB:
@samdroid-apps Done!

davelab6:
Looks good to me :)

quozl:
0.110.0 can still rename entries to blank in the detail view, when can we have this merged?

i5o:
Just tested. It kinda works and It kinda doesn't.
Sometimes the alert is gone :/ (I mean: set the name to nothing, and press enter) the first time the alert is shown, the 2nd one it isn't.

Maybe I did something wrong!

btw, https://github.com/sugarlabs/sugar/pull/663/commits/a24970174249e00a80175f8a01cddf1f08637261#diff-3f4b091ed74fd4c7f130a59dddba9bebR504

please check pep8 (add space after #)

i5o:
tested in sugar-build btw.

i5o:
ping x2!

quozl:
If @AbrahmAB is not available, feel free to diagnose and fix this yourself, and then make a new pull request with both their work and yours.  On the other hand I'm still of the opinion that blank names should be allowed; see [#4940](https://bugs.sugarlabs.org/ticket/4940).

iamutkarshtiwari:
@devAbnull Ping

quozl:
See also https://github.com/sugarlabs/sugar/pull/752

devAbnull:
@i5o I couldn't replicate the issue :(
Could you elaborate on replicating the issue.

chimosky:
Closed in favour of #949.

PR #665, opened 2016-03-09
Add manual proxy settings manager in network settings

ManashRaja:
Add GUI based proxy manager in the Network section of 'My Settings' to
enable management of manual proxy servers.

The proxy settings are stored in /etc/profile.d/proxy.sh as exports to
environment variables and requires re-login to let the changes take effect.

/extensions/cpsection/network/set_proxy.py file writes the proxy settings,
while two polkit policy files org.sugar.proxy.policy and org.sugar.logout.policy
located at /usr/share/polkit-1/actions provide root permissions required
for file handling and logout command. These two polkit files are placed at
data/ directory of this repository.

### DIFF
--- a/data/org.sugarlabs.gschema.xml
+++ b/data/org.sugarlabs.gschema.xml
@@ -390,23 +390,8 @@
             (something like 192.168.0.0/24).
             </description>
         </key>
-        <key name="use-same-proxy" type="b">
-            <default>true</default>
-            <summary>Unused; ignore</summary>
-            <description>
-            This key is not used, and should not be read or modified.
-            </description>
-        </key>
     </schema>
     <schema id="org.sugarlabs.system.proxy.http" path="/org/sugarlabs/system/proxy/http/">
-        <key name="enabled" type="b">
-            <default>false</default>
-            <summary>Unused; ignore</summary>
-            <description>
-            This key is not used; HTTP proxying is enabled when the host
-            key is non-empty and the port is non-0.
-            </description>
-        </key>
         <key name="host" type="s">
             <default>''</default>
             <summary>HTTP proxy host name</summary>
--- a/data/org.sugarlabs.gschema.xml
+++ b/data/org.sugarlabs.gschema.xml
@@ -333,4 +333,177 @@
             <description>This string is displayed in the control panel, about computer section.</description>
         </key>
     </schema>
+    <enum id='org.sugarlabs.desktop.GDesktopProxyMode'>
+        <value nick='none' value='0'/>
+        <value nick='manual' value='1'/>
+        <value nick='auto' value='2'/>
+        <value nick='system' value='3'/>
+    </enum>
+    <schema id="org.sugarlabs.system.proxy" path="/org/sugarlabs/system/proxy/">
+        <child name="http" schema="org.sugarlabs.system.proxy.http"/>
+        <child name="https" schema="org.sugarlabs.system.proxy.https"/>
+        <child name="ftp" schema="org.sugarlabs.system.proxy.ftp"/>
+        <child name="socks" schema="org.sugarlabs.system.proxy.socks"/>
+        <key name="mode" enum="org.sugarlabs.desktop.GDesktopProxyMode">
+            <default>'none'</default>
+            <summary>Proxy configuration mode</summary>
+            <description>
+            Select the proxy configuration mode. Supported values are "none", 
+            "manual", "auto".
+
+            If this is "none", then proxies are not used.
+
+            If it is "auto", the autoconfiguration URL described by the
+            "autoconfig-url" key is used.
+
+            If it is "manual", then the proxies described by
+            "/system/proxy/http", "/system/proxy/https",
+            "/system/proxy/ftp" and "/system/proxy/socks" will be used.
+            Each of the 4 proxy types is enabled if its "host" key is
+            non-empty and its "port" key is non-0.
+
+            If an http proxy is configured, but an https proxy is not,
+            then the http proxy is also used for https.
+
+            If a SOCKS proxy is configured, it is used for all protocols,
+            except that the http, https, and ftp proxy settings override
+            it for those protocols only.
+            </description>
+        </key>
+        <key name="autoconfig-url" type="s">
+            <default>''</default>
+            <summary>Automatic proxy configuration URL</summary>
+            <description>
+            URL that provides proxy configuration values. When mode is
+            "auto", this URL is used to look up proxy information for all
+            protocols.
+            </description>
+        </key>
+        <key name="ignore-hosts" type="as">
+            <default>[ 'localhost', '127.0.0.0/8', '::1' ]</default>
+            <summary>Non-proxy hosts</summary>
+            <description>
+            This key contains a list of hosts which are connected to directly, 
+            rather than via the proxy (if it is active). The values can be 
+            hostnames, domains (using an initial wildcard like *.foo.com), IP host 
+            addresses (both IPv4 and IPv6) and network addresses with a netmask 
+            (something like 192.168.0.0/24).
+            </description>
+        </key>
+        <key name="use-same-proxy" type="b">
+            <default>true</default>
+            <summary>Unused; ignore</summary>
+            <description>
+            This key is not used, and should not be read or modified.
+            </description>
+        </key>
+    </schema>
+    <schema id="org.sugarlabs.system.proxy.http" path="/org/sugarlabs/system/proxy/http/">
+        <key name="enabled" type="b">
+            <default>false</default>
+            <summary>Unused; ignore</summary>
+            <description>
+            This key is not used; HTTP proxying is enabled when the host
+            key is non-empty and the port is non-0.
+            </description>
+        </key>
+        <key name="host" type="s">
+            <default>''</default>
+            <summary>HTTP proxy host name</summary>
+            <description>
+            The machine name to proxy HTTP through.
+            </description>
+        </key>
+        <key name="port" type="i">
+            <range min="0" max="65535"/>
+            <default>8080</default>
+            <summary>HTTP proxy port</summary>
+            <description>
+            The port on the machine defined by "/system/proxy/http/host" that you 
+            proxy through.
+            </description>
+        </key>
+        <key name="use-authentication" type="b">
+            <default>false</default>
+            <summary>Authenticate proxy server connections</summary>
+            <description>
+            If true, then connections to the proxy server require authentication. 
+            The username/password combo is defined by 
+            "/system/proxy/http/authentication-user" and 
+            "/system/proxy/http/authentication-password".
+
+            This applies only to the http proxy; when using a separate
+            https proxy, there is currently no way to specify that it
+            should use authentication.
+            </description>
+        </key>
+        <key name="authentication-user" type="s">
+            <default>''</default>
+            <summary>HTTP proxy username</summary>
+            <description>
+            User name to pass as authentication when doing HTTP proxying.
+            </description>
+        </key>
+        <key name="authentication-password" type="s">
+            <default>''</default>
+            <summary>HTTP proxy password</summary>
+            <description>
+            Password to pass as authentication when doing HTTP proxying.
+            </description>
+        </key>
+    </schema>
+    <schema id="org.sugarlabs.system.proxy.https" path="/org/sugarlabs/system/proxy/https/">
+        <key name="host" type="s">
+            <default>''</default>
+            <summary>Secure HTTP proxy host name</summary>
+            <description>
+            The machine name to proxy secure HTTP through.
+            </description>
+        </key>
+        <key name="port" type="i">
+            <range min="0" max="65535"/>
+            <default>0</default>
+            <summary>Secure HTTP proxy port</summary>
+            <description>
+            The port on the machine defined by "/system/proxy/https/host" that you 
+            proxy through.
+            </description>
+        </key>
+    </schema>
+    <schema id="org.sugarlabs.system.proxy.ftp" path="/org/sugarlabs/system/proxy/ftp/">
+        <key name="host" type="s">
+            <default>''</default>
+            <summary>FTP proxy host name</summary>
+            <description>
+            The machine name to proxy FTP through.
+            </description>
+        </key>
+        <key name="port" type="i">
+            <range min="0" max="65535"/>
+            <default>0</default>
+            <summary>FTP proxy port</summary>
+            <description>
+            The port on the machine defined by "/system/proxy/ftp/host" that you 
+            proxy through.
+            </description>
+        </key>
+    </schema>
+    <schema id="org.sugarlabs.system.proxy.socks" path="/org/sugarlabs/system/proxy/socks/">
+        <key name="host" type="s">
+            <default>''</default>
+            <summary>SOCKS proxy host name</summary>
+            <description>
+            The machine name to use as a SOCKS proxy.
+            </description>
+        </key>
+        <key name="port" type="i">
+            <range min="0" max="65535"/>
+            <default>0</default>
+            <summary>SOCKS proxy port</summary>
+            <description>
+            The port on the machine defined by "/system/proxy/socks/host" that you 
+            proxy through.
+            </description>
+        </key>
+    </schema>    
 </schemalist>
--- a/extensions/cpsection/network/view.py
+++ b/extensions/cpsection/network/view.py
@@ -18,16 +18,267 @@
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gettext import gettext as _
+from gi.repository import Gio
+from gi.repository import Pango
+from gi.repository import GLib
 
+from sugar3.graphics.icon import Icon
 from sugar3.graphics import style
+from sugar3.graphics.alert import Alert
 
 from jarabe.controlpanel.sectionview import SectionView
 from jarabe.controlpanel.inlinealert import InlineAlert
 
+import os
 
 CLASS = 'Network'
 ICON = 'module-network'
 TITLE = _('Network')
+
+_APPLY_TIMEOUT = 3000
+
+
+def __setitem__(self, key, value):
+    # set_value() aborts the program on an unknown key
+    if key not in self:
+        raise KeyError('unknown key: %r' % (key,))
+
+    # determine type string of this key
+    range = self.get_range(key)
+    type_ = range.get_child_value(0).get_string()
+    v = range.get_child_value(1)
+    if type_ == 'type':
+        # v is boxed empty array, type of its elements is the allowed value type
+        assert v.get_child_value(0).get_type_string().startswith('a')
+        type_str = v.get_child_value(0).get_type_string()[1:]
+    elif type_ == 'enum':
+        # v is an array with the allowed values
+        assert v.get_child_value(0).get_type_string().startswith('a')
+        type_str = v.get_child_value(0).get_child_value(0).get_type_string()
+    elif type_ == 'flags':
+        # v is an array with the allowed values
+        assert v.get_child_value(0).get_type_string().startswith('a')
+        type_str = v.get_child_value(0).get_type_string()
+    elif type_ == 'range':
+        # type_str is a tuple giving the range
+        assert v.get_child_value(0).get_type_string().startswith('(')
+        type_str = v.get_child_value(0).get_type_string()[1]
+
+    if not self.set_value(key, GLib.Variant(type_str, value)):
+        raise ValueError("value '%s' for key '%s' is outside of valid range" % (value, key))
+
+
+def bind_with_convert(self, key, widget, prop, flags, key_to_prop, prop_to_key):
+    self._ignore_key_changed = False
+
+    def key_changed(settings, key):
+        if self._ignore_key_changed:
+            return
+        self._ignore_prop_changed = True
+        widget.set_property(prop, key_to_prop(self[key]))
+        self._ignore_prop_changed = False
+
+    def prop_changed(widget, param):
+        if self._ignore_prop_changed:
+            return
+        self._ignore_key_changed = True
+        self[key] = prop_to_key(widget.get_property(prop))
+        self._ignore_key_changed = False
+
+    if not (flags & (Gio.SettingsBindFlags.SET | Gio.SettingsBindFlags.GET)):
+        # ie Gio.SettingsBindFlags.DEFAULT
+        flags |= Gio.SettingsBindFlags.SET | Gio.SettingsBindFlags.GET
+    if flags & Gio.SettingsBindFlags.GET:
+        key_changed(self, key)
+        if not (flags & Gio.SettingsBindFlags.GET_NO_CHANGES):
+            self.connect('changed::' + key, key_changed)
+    if flags & Gio.SettingsBindFlags.SET:
+        widget.connect('notify::' + prop, prop_changed)
+    if not (flags & Gio.SettingsBindFlags.NO_SENSITIVITY):
+        self.bind_writable(key, widget, "sensitive", False)
+
+Gio.Settings.bind_with_convert = bind_with_convert
+Gio.Settings.__setitem__ = __setitem__
+
+
+def type_as_to_string(value):
+    return ",".join(value)
+
+
+def string_to_type_as(value):
+    return value.split(',')
+
+
+class NumberEntry(Gtk.Entry):
+    def __init__(self):
+        Gtk.Entry.__init__(self)
+        self.connect('changed', self.on_changed)
+
+    def on_changed(self, *args):
+        text = self.get_text().strip()
+        self.set_text(''.join([i for i in text if i in '0123456789']))
+
+
+class SettingBox(Gtk.HBox):
+    """
+    Base class for "lines" on the screen representing configuration
+    settings.
+    """
+    def __init__(self, name, size_group=None):
+        Gtk.HBox.__init__(self, spacing=style.DEFAULT_SPACING)
+        label = Gtk.Label(name)
+        label.modify_fg(Gtk.StateType.NORMAL,
+                        style.COLOR_SELECTION_GREY.get_gdk_color())
+        label.set_alignment(1, 0.5)
+        if size_group is not None:
+            size_group.add_widget(label)
+        self.pack_start(label, False, False, 0)
+        label.show()
+
+
+class ComboSettingBox(Gtk.VBox):
+    """
+    Container for sets of different settings selected by a top-level
+    setting.
+
+    Renders the top level setting as a ComboBox.  Only the currently
+    active set is shown on screen.
+    """
+    def __init__(self, name, setting, setting_key,
+                 option_sets, size_group=None):
+        Gtk.VBox.__init__(self, spacing=style.DEFAULT_SPACING)
+
+        setting_box = SettingBox(name, size_group)
+        self.pack_start(setting_box, False, False, 0)
+        setting_box.show()
+
+        model = Gtk.ListStore(str, str, object)
+        combo_box = Gtk.ComboBox(model=model)
+        combo_box.connect('changed', self.__combo_changed_cb)
+        setting_box.pack_start(combo_box, True, True, 0)
+        combo_box.show()
+
+        cell_renderer = Gtk.CellRendererText()
+        cell_renderer.props.ellipsize = Pango.EllipsizeMode.MIDDLE
+        cell_renderer.props.ellipsize_set = True
+        combo_box.pack_start(cell_renderer, True)
+        combo_box.add_attribute(cell_renderer, 'text', 0)
+        combo_box.props.id_column = 1
+
+        self._settings_box = Gtk.VBox()
+        self._settings_box.show()
+        self.pack_start(self._settings_box, False, False, 0)
+
+        for optset in option_sets:
+            model.append(optset)
+
+        setting.bind(setting_key, combo_box, 'active-id',
+                     Gio.SettingsBindFlags.DEFAULT)
+
+    def __combo_changed_cb(self, combobox):
+        giter = combobox.get_active_iter()
+        new_box = combobox.get_model().get(giter, 2)[0]
+        current_box = self._settings_box.get_children()
+        if current_box:
+            self._settings_box.remove(current_box[0])
+
+        self._settings_box.add(new_box)
+        new_box.show()
+
+
+class OptionalSettingsBox(Gtk.VBox):
+    """
+    Container for settings (de)activated by a top-level setting.
+
+    Renders the top level setting as a CheckButton. The settings are only
+    shown on screen if the top-level setting is enabled.
+    """
+    def __init__(self, name, setting, setting_key, contents_box):
+        Gtk.VBox.__init__(self, spacing=style.DEFAULT_SPACING)
+
+        check_button = Gtk.CheckButton()
+        check_button.props.label = name
+        check_button.connect('toggled', self.__button_toggled_cb, contents_box)
+        check_button.show()
+        self.pack_start(check_button, True, True, 0)
+        self.pack_start(contents_box, False, False, 0)
+
+        setting.bind(setting_key, check_button, 'active',
+                     Gio.SettingsBindFlags.DEFAULT)
+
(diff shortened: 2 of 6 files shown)


samdroid-apps:
Hi @ManashRaja, thanks for working on this, it looks like a very useful feature for schools that require proxies!!  I am yet to review this in depth, but just a few quick comments:

1.  This feature might require discussion with the community, including non-technical stakeholders.  Does this have a feature page attached to it to help foster that discussion?  Or just some screenshots?
2.  What setup did you use to test this?  I don't normally use a proxy to connect to the internet, is there an easy way I can set this up?
2.  Polkit scripts should allow for 1 use and 1 use only.  Some of the policies that you wrote seem a little vauge, and I could exploit them to gain root as an attacker.  Eg, with the "org.sugar.proxy.policy", I can run "python evilscript.py" and run anything as root.  Please look at how to lock these down, eg. making a helper script that only allows for good code to be run.
4.  Please run flake8 against your code

Anyway, looks great.  Hopefully we an work this up to inclusion in the next release!

ManashRaja:
Hello @samdroid-apps, 

1. I found a feature page for this https://wiki.sugarlabs.org/go/Features/Proxy_Settings which targeted release version 0.99. But, since this feature is unavailable in the latest releases of sugar, hence I wonder whether shall I port the 0.99 version to the present versions or go on with the modification of my implementation that I presented in this pr.

2. My college Internet goes though a proxy server and hence I tested it with the required settings. A very quick method to test if it works is to set any random proxy for http and https, and check if the settings are taken by echoing http_proxy and https_proxy environment variables, or by trying to access any webpage though the browser, which is expected to give "webpage not found" error.

3. I will look into the policykit implementation and modify it.

4. And will get back to you with flake8 verified code.

ManashRaja:
Hello @samdroid-apps, 
As mentioned, I have come up with a more secure implementation of polkit permission and have improved my code using flake8.

Thanks

quozl:
Please also review https://github.com/godiard/sugar/commit/ebfd1ed0f941dca982b90d0d08efda7c2ae75a84.

quozl:
* gsettings called many times during startup is inefficient, consider src/jarabe/main.py,
 * you are using your master branch, be careful,
 * please collapse the commits and force push,
 * add screenshots to pull request and/or Wiki.

ManashRaja:
- There is no direct way to export environment variables through python and hence it seems like moving to src/jarabe/main.py will introduce redundant and extra script files. We can use the present /usr/bin/sugar and reduce the number of times gsettings is called by keeping the values at variables in cases where they are needed multiple times.
- Once the above implementation is decided, I will collapse the commits into one with a new message and screenshots.

quozl:
* src/jarabe/main.py does already export environment variables.

ManashRaja:
@quozl : Thanks for the suggestions to use /jarabe/main.py . 
- I have collapsed the multiple commits into one.
- Used /jarabe/main.py to export the proxy settings to environment variables.
- The following is a screenshot of the modified Network Control Panel.
proxy-settings

Can you help me figuring out why is the sugar.in included in this commit even if it is same with the sugarlabs/sugar repository?

quozl:
> Can you help me figuring out why is the sugar.in included in this commit even if it is same with the sugarlabs/sugar repository?

You changed tabs to spaces.  You have an editor which does not show this properly.  You might do this to fix;

```
git checkout origin/master bin/sugar.in
```

Then amend your commit and force push.

Medium term, before commit make sure you understand the "git diff" output fully, and revert any changes you don't want.  I use Magit in Emacs, which shows me tiny changes that I can revert by keyboard shortcut.

Long term, use an editor that shows the difference between tabs and spaces.  My preference is emacs with some custom Lisp code.  The Emacs Wiki has an article on visible tabs.  https://www.emacswiki.org/emacs/ShowWhiteSpace  Also stackoverflow.com has several questions on the same issue.  Do some research.

src/jarabe/main.py
```diff
@@ -347,6 +347,43 @@ def setup_fonts():
...
+        text_to_set = "ftp://" + g_ftp.get_string('host') + \
+                      ":" + str(g_ftp.get_int('port')) + "/"
+        os.environ['ftp_proxy'] = text_to_set
+
+        g_socks = Gio.Settings('org.gnome.system.proxy.socks')
+        text_to_set = "socks://" + g_socks.get_string('host') + \
+                      ":" + str(g_socks.get_int('port')) + "/"
+        os.environ['socks_proxy'] = text_to_set
+
+    elif g_mode == 'none':
```
quozl: Here you erase the environment variables if the setting is set to none.  I'm surprised.  What if the system administrator has set a default proxy in /etc/profile.d or other scripts?

src/jarabe/main.py
```diff
@@ -347,6 +347,43 @@ def setup_fonts():
...
 
 
+def setup_proxy():
+    g_mode = Gio.Settings('org.gnome.system.proxy').get_string('mode')
+    if g_mode == 'manual':
+        g_http = Gio.Settings('org.gnome.system.proxy.http')
+        g_http_host = g_http.get_string('host')
+        g_http_port = str(g_http.get_int('port'))
+        if(g_http.get_boolean('use-authentication')):
+            text_to_set = "http://" + g_http.get_string('authentication-user') + \
```
quozl: This is a very complex assignment to text_to_set, because you use string concatenation operator "+".  Makes security review more difficult.  Simplify using string format operators, like .format or "%".

src/jarabe/main.py
ManashRaja: The final setting in the environment variable will be set by the script that runs last. I just tested setting the proxy at both /etc/environment file and a custom /etc/profile.d/proxy.sh file and I found that the file at /etc/profile.d is getting called after the execution of /jarabe/main.py. In fact the sequence seems to be /etc/environment => jarabe/main.py => profile.d/prosh.sh

quozl: I disagree.  I think you tested with Terminal activity, which is a new shell process, which therefore executes /etc/profile.d again.  Test instead with an activity such as Browse, which is not a shell interpreter.

Besides what you say; /etc/environment may be the file a system administrator sets proxy, and if so your proposed code persists in losing the configuration.

Why do you want to lose the system-wide configuration?

src/jarabe/main.py
ManashRaja: Simplified using string format operator %

src/jarabe/main.py
ManashRaja: I tested with browser and found as you indicated that the proxy settings are as per the /jarabe/main.py. What do you suggest how a system administrator should be able to set default proxy?

ManashRaja:
@quozl How do you expect the proxy settings be set in case of a system administrator settings default proxy settings?

quozl:
@ManashRaja, in the systems I've administered for the past 30 years, there has been no clear standard for where and when to set environment variables.  Some systems use shell initialisation files, all of which depend on which shell the user has selected using chsh.  Some systems use graphical shell initialisation files, like .xsession or xinitrc.  Some set the environment variable in the process that starts the shell, e.g. sshd, lightdm, getty; although this has tended to become unsupported so that environments are pristine and not inherited.

You will see in Firefox, Preferences, Advanced, Network; *use system proxy settings*.  So perhaps you should have that as an option in the UI?

ManashRaja:
Ya, so as you mentioned that a system administrator need to set up default proxy settings sometimes. Also that there haven't been a clear standard for the settings. I wonder after this feature is implemented, why would a system admin not set the proxy through the "Network" control panel itself rather than using any of his own script?

quozl:
```irc
[07:03:03] <ManashRaja> Ya, so as you mentioned that a system administrator need to set up default proxy settings sometimes. Also that there haven't been a clear standard for the settings. I wonder after this feature is implemented, why would a system admin not set the proxy through the "Network" control panel itself rather than using any of his own script?
[07:03:05] <Quozl`_> Your options seem to be either (a) remove the erase of environment variables (elif == 'none'), or (b) add a new option "use system settings", which is not reacted to by the src/jarabe/main.py fragment
[07:03:28] <Quozl`_> Because in a shared system the system administrator must not make decisions for all users unilaterally.
[07:03:52] <Quozl`_> While it is true that most computers are personal, there still exist and will continue to exist many computers with more than one user.
[07:04:18] <Quozl`_> And the decisions that users make include (a) desktop choice, (b) proxy.
[07:04:54] <ManashRaja> Okay so there is something else reffered to as "system settings" I have always used personal computer so the option at firefox that says "system settings" meant the settings I set through Network section itself
[07:05:12] <Quozl`_> Common scenario is LTSP thin client classrooms, with big server.  Each desktop mounts a large NFS share for root filesystem.
[07:05:22] <Quozl`_> Yes, good observation.
[07:05:40] <Quozl`_> ManashRaja: I guess you are familiar with only one desktop environment and firefox?
[07:06:09] <ManashRaja> Quozl`_: ya, that's why I was confused about this implementation
[07:06:33] <ManashRaja> Quozl`_: so I will add another section there to set "system settings"
[07:07:08] <Quozl`_> ManashRaja: okay, great.  That will make Sugar compatible with other packages that might be installed on a large computer.
[07:09:02] <ManashRaja> Quozl`_: But then I want to know, where would the system admin set the proxy settings? In files like /etc/environment or /etc/profile.d . I guess than the jarabe/main.py file doesn't need to export anything while "system settings" is selected. The shared desktop will take the settings from where the sys admin have stored and jarabe/main.py will not overwrite it
[07:11:38] <Quozl`_> ManashRaja: as I said, there are too many places where a system administrator (or custom package) may set these environment variables, so it is pointless to try to list them all; what benefit would be achieved by such a list?  Yes, when "system settings" is selected, Firefox does nothing to the values; or restores it to the value it had at program start.
[07:12:55] <ManashRaja> Oh cool. fine then. So I would update the pr keeping only one commit but with a new commit message to include this option.
[07:14:06] <Quozl`_> Have fun!
```

ManashRaja:
<pre>[03:25] &lt;ManashRaja&gt; Ya, so I am adding a new option "Use system proxy". To implement it, I was thinking of writing a custom schema. Because this option can't be accomodated in "org.gnome.system.proxy mode" as it only accepts "none", "manual", and "auto".
[03:47] &lt;Quozl`_&gt; ManashRaja: my guess is that "none" means not to set the environment variables.
[03:47] &lt;Quozl`_&gt; ... or "auto".  Have you found code that uses org.gnome.system.proxy to be able to check?
[03:50] &lt;ManashRaja&gt; "manual" and "auto' are types of settings. Then I think "none" is for no proxy settings at all and direct connection, where I beleive the proxy settings should be cleared and hence for adding the feature of letting user set system proxy, we may have another option "use system proxy"
[03:51] &lt;Quozl`_&gt; ManashRaja: have you found the GNOME equivalent code for using the setting?
[03:53] &lt;ManashRaja&gt; Quozl`_: I do have the one that uses the standard shema "org.gnome.system.proxy" as I implemented in my PR so far. But to add another option, I am thinking of creating a new custom schema "org.sugarlabs.systemproxy"
[03:54] &lt;Quozl`_&gt; ManashRaja: you mean "GNOME schema", not "standard".  GNOME isn't a standard, it's a desktop environment.
[03:54] &lt;ManashRaja&gt; for that I found /sugar/data/org.sugarlabs.gschema.xml where we can add custom schemas
[03:54] &lt;Quozl`_&gt; (and there are other desktop environments).
[03:57] &lt;ManashRaja&gt; Quozl`_: Oh, sorry for that. I used the gnome based schemas as they were used in https://github.com/godiard/sugar/commit/ebfd1ed0f941dca982b90d0d08efda7c2ae75a84 you sent to me to review
[03:58] &lt;Quozl`_&gt; ManashRaja: I've no idea why those were used, I was not part of the development.
[03:58] &lt;ManashRaja&gt; We might instead use sugarlabs based schemas instead "org.sugarlabs.system.proxy" which I guess will be more apt and good for customization as well as logical in terms of sugar as a desktop environment having its own schemas
[03:59] &lt;Quozl`_&gt; Yes, we can alsop make the options mean what we mean instead of being unable to explain what they mean.
[04:00] &lt;ManashRaja&gt; Quozl`_: Yes, customizability will be an advantage. So, I will create new schemas then by modifying /sugar/data/org.sugarlabs.gschema.xml file
[04:00] &lt;Quozl`_&gt; Ok.
[04:01] &lt;ManashRaja&gt; and sugar/data/sugar-schemas.convert file also.
[04:01] &lt;ManashRaja&gt; Cool</pre>

ManashRaja:
@quozl The option has been included with custom schemas :)
proxy

data/org.sugarlabs.gschema.xml
```diff
@@ -333,4 +333,177 @@
             <description>This string is displayed in the control panel, about computer section.</description>
         </key>
     </schema>
+    <enum id='org.sugarlabs.desktop.GDesktopProxyMode'>
```
samdroid-apps: This is a very nicely documented schema, but I think that something lost the indentation up the top here.

src/jarabe/main.py
```diff
@@ -347,6 +347,50 @@ def setup_fonts():
...
 
 
+def setup_proxy():
+    g_mode = Gio.Settings('org.sugarlabs.system.proxy').get_string('mode')
+    if g_mode == 'manual':
+        g_http = Gio.Settings('org.sugarlabs.system.proxy.http')
+        g_http_host = g_http.get_string('host')
+        g_http_port = str(g_http.get_int('port'))
+        if(g_http.get_boolean('use-authentication')):
+            text_to_set = "http://%s:%s@%s:%s/" % (
```
samdroid-apps: Wouldn't it make sense to also have the authentication with https proxies?

Disclaimer:  never used a proxy in my life.

src/jarabe/main.py
```diff
@@ -347,6 +347,50 @@ def setup_fonts():
...
+        text_to_set = "ftp://%s:%s/" % (
+            g_ftp.get_string('host'),
+            str(g_ftp.get_int('port')))
+        os.environ['ftp_proxy'] = text_to_set
+
+        g_socks = Gio.Settings('org.sugarlabs.system.proxy.socks')
+        text_to_set = "socks://%s:%s/" % (
+            g_socks.get_string('host'),
+            str(g_socks.get_int('port')))
+        os.environ['socks_proxy'] = text_to_set
```
samdroid-apps: There is so much copy-paste code through this section.  I'd wager that it doesn't improve the readability, so maybe use a loop?

samdroid-apps:
You don't seem to do anything with the `auto` proxy settings in main.py, even though you do something with the `manual` proxy settings in main.py.

Have you tested `auto` proxy settings?  What procedure can I use to test it too?

samdroid-apps:
Also, how do I test?  I did:

0.  Select manual proxy
1.  Put "not.sam.today" (bogus subdomain) in the HTTP proxy setting
1.  Tick "use-authentication"
2  Put a bogus username and password in
3.  Apply and restarted as prompted

Expected: Browse could not go to http sites

Actual:  Browse loaded http://example.com just fine

src/jarabe/main.py
ManashRaja: I wrote the settings on the basis of how it has been shown in archlinux [1]
`  export http_proxy="http://$pre$server:$port/" #$pre = "$username:$password@" used only for http
    export https_proxy=$http_proxy
    export ftp_proxy=$http_proxy
    export rsync_proxy=$http_proxy
    export HTTP_PROXY=$http_proxy
    export HTTPS_PROXY=$http_proxy
    export FTP_PROXY=$http_proxy
    export RSYNC_PROXY=$http_proxy`

And proxy setting implementation in Gnome [2]. The following are the settings available in Gnome that are related to proxy server. 
`org.gnome.system.proxy autoconfig-url ''
org.gnome.system.proxy ignore-hosts ['localhost', '127.0.0.0/8']
org.gnome.system.proxy mode 'none'
org.gnome.system.proxy use-same-proxy true
org.gnome.system.proxy.ftp host ''
org.gnome.system.proxy.ftp port 0
org.gnome.system.proxy.http authentication-password ''
org.gnome.system.proxy.http authentication-user ''
org.gnome.system.proxy.http enabled false
org.gnome.system.proxy.http host ''
org.gnome.system.proxy.http port 8080
org.gnome.system.proxy.http use-authentication false
org.gnome.system.proxy.https host ''
org.gnome.system.proxy.https port 0
org.gnome.system.proxy.socks host ''
org.gnome.system.proxy.socks port 0`

[1] https://wiki.archlinux.org/index.php/proxy_settings
[2] https://developer.gnome.org/ProxyConfiguration/

ManashRaja:
@samdroid-apps ,
Thanks for the review.
- data/org.sugarlabs.gschema.xml wrong indentation: **Fixed** in latest commit 
- src/jarabe/main.py: copy paste material converted to loop. **Fixed** in latest commit 
- Auto proxy: The auto proxy setting is to store the location of a *.pac file. This file tells an application, what proxy configuration to use in order to access a particular url. Hence, I understand that the main.py file does not have to do anything as there seems to be no standard environment variable defined for it unlike "manual" proxy settings as seen at Archlinux [1]. By letting the user set auto proxy in the Network Settings is to allow the location of *.pac file to be stored in "org.sugarlabs.system.proxy autoconfig-url" schema as it is stored in "org.gnome.system.proxy autoconfig-url" in gnome. So that any application trying to access internet, (like browser) can query the schema and get the location of the file. Though to increase its accessibility, we can also store the location at "org.gnome.system.proxy autoconfig-url" for application trying to query gnome schemas.  
I believe for such type of proxy setting, an application has its own interface where you can set the url for the file. (Like "Automatic proxy configuration url:" in firefox). We just store the url at schemas if any application might want to use it. Please correct me if I am wrong because this is what I understand. For reference [2]
- Proxy test: I wonder why you were not getting the desired behavior out of the proxy settings. Have you opted for a **restart** after the settings were applied? Also if you have set http proxy alone, then the rules will be applied to only http sites not https. So try something like http://manashpratim.in/ instead of https://www.google.com/ . I followed the same procedure you did except that I need valid settings for my proxy server to get internet access.

[1] https://wiki.archlinux.org/index.php/proxy_settings
[2] http://ubuntuforums.org/showthread.php?t=1541428

ManashRaja:
Shall I add ignore-hosts option? I almost forgot it as the option was not there in Ubuntu proxy setting page.

ManashRaja:
@samdroid-apps I have added ignore hosts option so that user can define hosts for which proxy settings are not required. I added this option to 'manual' type proxy alone as in case of gnome Network Manager. 
I tested the feature by adding www.google.com to the comma separated values in the Ignore hosts: field and get the desired behavior (Did not load as I need proxy for internet connection).

Here is the screen-shot of how it looks now:
ignore-hosts

So, with this, I guess all the work have been done and the feature is fully functional as it is in gnome desktop environment. 
Please review it.

ManashRaja:
Squashed multiple commits to one.

quozl:
What is the risk of misconfiguration?

Many years ago, one of the reasons put for why this feature should not be merged is that it would be too easy for a very early learner to enter bad data into the fields and stop all internet work, and it would take a long time for a teacher or peer to identify the problem.  What happens when garbage text or incorrect data is entered into the new control panel widgets?

ManashRaja:
If garbage text or incorrect data is entered in the settings, then the internet connection will suffer. But, I don't think this feature will bring on serious risks as:
- While troubleshooting for any network error even though internet connection is present, then it is assumed that one would look for the settings in the "Network" section of the control panel to solve it out.
- The feature as is implemented by "jarabe.main" which is run last on boot, overrides all other proxy settings if present on other configuration files that might be hidden. Hence enabling or disabling the proxy at this menu will ensure the correct settings to be taken. In contrast "use system proxy" option can be opted if you don't want to override the settings. So, the menu will prove to be an easy and ultimate way to handle proxy settings and this exclusivity will make it on the top location to search for in case of troubleshooting. 

Hence, my point is that, even though it might be easy for someone to mess with the settings as is the case with all settings, but still it would be very easy for a teacher or peer to identify the problem as they have to look at only one place which is the ultimate place that dictates proxy settings. :)

samdroid-apps:
Can you ping the proxy before you set it?  Eg, make the flow like:

1.  User enters "hglbugrhgwtaiu" as http proxy
2.  User clicks apply
3.  Progress bar shows loading
4.  Alert says, "the proxy is not accessable, setting the settings will break your internet connection" with the buttons [Reset] as default and [Break my internet] hidden somewhere in small text

ManashRaja:
Yes, that can be done and will prevent setting false proxy! Great :+1: 
It would be tested for all the types like http, https, socks, ftp and autoconfig-url right?

quozl:
When devices are shared among users, it would be too easy for a skilled user to set an invalid proxy that an unskilled user could not overcome.  So I'm interested to know what Browse reports when the proxy is invalid; does it refer to the proxy settings in a way that helps the unskilled user?

ManashRaja:
@quozl  Browse does not report that the proxy is invalid. It shows the following error in any case when it is not able to access the internet. It does not even mention about troubleshooting the proxy settings.
screenshot from 2016-04-04 16 10 24

ManashRaja:
@samdroid-apps 
Here is the draft implementation demo for the proxy verification:
out-2

Suggest edits please.
This feature can be more enhanced by:
- using inline alerts that displays which particular setting is invalid
- utilizing the top alert that shows restart option instead of a new alert in the proxy section. As far as I understand, it would involve editing the "jarabe/controlpanel/gui.py" file, hence I included an extra alert.

davelab6:
When the dialog with the 'break my internet' button is shown, is it still
possible to click the "restart now" or "later" buttons at the top?

ManashRaja:
@davelab6 Yes, it is possible to click the buttons on the top as I am not able to figure out a way to hide them or disable the buttons. 

The top alert bar is created by "jarabe/controlpanel/gui.py". I am not able to understand the inheritance and how the different classes (ControlPanel, SectionView, Newtork) are being set up and hence am not able to call the methods present in the above file from the "cpsections/network/view.py" file. What I find is that at present there isn't a framework to modify the flow of how alerts are displayed and methods are called to suit our needs other than the existing restart alert which we can only enable by calling "self.needs_restart = True" and can't disable afterwards.

At present we can only modify the alert title and msg.
But it would be great if we can also modify the buttons, their text and callback response to support custom extended features like presenting "break my internet" type options in the main alert itself.

quozl:
@ManashRaja, thanks for the Browse screenshot; confirms that Browse does not explain the problem, and supports our need for checking the proxy settings in another way.

Yes, jarabe/controlpanel/gui.py is quite complex, but with a careful review you should be able to work out what is needed.  Follow the control flow to see how a cpsection module is imported.  It is unconventional, because we have the flexibility to add cpsection modules from outside Sugar.  Examine other cpsection modules; I seem to recall there may be one that calls back into the gui.py class.

The double alert looks like it might be confusing.  Other sections use a triangle icon to identify problems.

The + and - buttons on the port numbers are really not required; there's almost never a situation where a user will need to use them, and so they could be the *cause* of accidental misconfiguration.

ManashRaja:
@quozl 
I found the following is the flow of how a ControlPanel section is created.
buddymenu.py ==imports==> gui.py ===getattr==> network/view.py
and while fetching and calling view.py, it passes the following arguments:
```self._section_view = view_class(model, self._options[option]['alerts'])```
the `model` is a wrapper for network/model.py while `self._options[option]['alerts']` are lists of additional alert names which do not include the main restart alert in question.(top one). Other cpsection modules use `self._options[option]['alerts']` as a holder for custom inline-alert (small triangle alerts next to each setting feild) names and do not interact with the main restart alert.

Hence I can't find a way to actually modify the alert created by gui.py from view.py due to encapsulation without actually passing the instance of gui.py to view.py through the `view_class` call shown above. But this would require editing the gui.py file and all the view.py files in cpsection folder. Shall I edit them as I guess it might provide even more flexibility in future?

**Comment by quozl** [comment]
Thanks, nice summary.

Yes, include gui.py in your edits and let us see what you suggest.

See if you can make your changes to gui.py in a cautious and re-usable way that provides an API to section instance, rather than only expose data.

**Comment by ManashRaja** [comment]
@quozl , Thanks
So here are a few modifications:
- The + and - buttons from port gtk.spinbuttons have been removed by replacing it with gtk.entry
- Inline Alerts have been added to identify error in individual level of host setting
- As mentioned, I edited the gui.py with as much less modification as it seems possible to allow for the management of the restart top alert and now it can be hidden when not required. (demo below). 
But it requires to pass an extra argument for the instance to the Network class being called. Hence this will break modules other than Network (as I edited it) as those modules won't be expecting that extra argument.
So please check the modifications at https://github.com/sugarlabs/sugar/pull/665/commits/1061ffe4d2c878d7a4b69e2b7ab0baee00e9b9da#diff-31e5dd3b4b74524fa7f048df8fbb99b2 . If we can continue then I will modify the other modules accordingly in a different PR.

proxy-verification-gif

**Comment by quozl** [comment]
You wrote:
> it requires to pass an extra argument for the instance 

You could make that an optional named argument?  That way you will not need to change any of the other modules.  Reference found in a hasty search; http://www.diveintopython.net/power_of_introspection/optional_arguments.html

**Comment by ManashRaja** [comment]
@quozl , no Sir,
optional arguments allow calling a function (say: ```myFunc(arg1, arg2)```) by passing less arguments (2 in this case), while it can accept more arguments which are optional ( ```def myFunc(arg1, arg2, opt_arg1=1, opt_arg2=2):``` ).  

But the function I am dealing with accepts lesser arguments than I am passing it.
For example say: ```myFunc(arg1, arg2, extra_arg1)``` while the function is defined as ```def myFunc(arg1, arg2):```   
So I just need to edit just one line in view.py files of all the module to make it accept one more argument.
Replace ```def __init__(self, model, alerts):``` with ```def __init__(self, model, alerts, caller_instance):```

**Comment by quozl** [comment]
Well, I'm surprised, but you're right, I had things the wrong way around.

So you are proposing to add an argument to each class.

The reason this is bad is that we have external and third-party control panel sections, not just at OLPC, but also at deployments.  You can't edit them.  Breaking the API like this must be avoided, if we can.  I'm looking for an alternative way to implement.

How about a new signal *request-alert* sent by the section view (self.emit), and a connect in ControlPanel.show_section_view to redirect to a callback method?

(Others will also reject this patch because you are using private variables in the ControlPanel class from outside the class.  Private variables have an underscore (_) prefix.)

**Comment by quozl** [comment]
Please take over the feature page; https://wiki.sugarlabs.org/go/Features/Proxy_Settings ... it hasn't changed from 2013.  Delete any old information.

**Comment by ManashRaja** [comment]
Oh cool. :+1:  I will try the signal and callback method and report back. 
And will edit the feature page. What should go in the current status section? also I guess there is no target version as of now right?

**Comment by quozl** [comment]
Sorry, I don't know the process for inclusion in next release, but it looks like you just add a link to the feature page in one of the next release subpages; https://wiki.sugarlabs.org/go/0.110/Feature_List

**Comment by ManashRaja** [comment]
Thanks. I am learning new things everyday as is so exciting. :)
So, instance and extra argument related matter is completely removed. GObject signals have been used to achieve the same as in the previous gif. 

- `request-restart-alert` : displays the alert
- `set-toolbar-display` : which accepts a GObject.TYPE_BOOLEAN sets the sensitivity of the toolbar buttons

**Comment by quozl** [comment]
Quick review:

* signal callbacks in gui.py should always end in _cb,
* `request-restart-alert` might be `show-alert`,
* `set-toolbar-display` might be `set-toolbar-sensitive`,

More review later.

**Comment by ManashRaja** [comment]
Edits done. Will squash all commits to one after final review and edit.

**Comment by ManashRaja** [comment]
Updated feature page. https://wiki.sugarlabs.org/go/Features/Proxy_Settings
Please check.

File: data/org.sugarlabs.gschema.xml [hunk 6]
```diff
@@ -333,4 +333,177 @@
...
+            hostnames, domains (using an initial wildcard like *.foo.com), IP host 
+            addresses (both IPv4 and IPv6) and network addresses with a netmask 
+            (something like 192.168.0.0/24).
+            </description>
+        </key>
+        <key name="use-same-proxy" type="b">
+            <default>true</default>
+            <summary>Unused; ignore</summary>
+            <description>
+            This key is not used, and should not be read or modified.
```
samdroid-apps: can we remove it if it is not used?

data/org.sugarlabs.gschema.xml
```diff
@@ -333,4 +333,177 @@
...
+            </description>
+        </key>
+    </schema>
+    <schema id="org.sugarlabs.system.proxy.http" path="/org/sugarlabs/system/proxy/http/">
+        <key name="enabled" type="b">
+            <default>false</default>
+            <summary>Unused; ignore</summary>
+            <description>
+            This key is not used; HTTP proxying is enabled when the host
+            key is non-empty and the port is non-0.
```
samdroid-apps: can we remove this if not used?

extensions/cpsection/network/view.py
```diff
@@ -18,17 +18,268 @@
...
+
+        setting.bind_with_convert(setting_key,
+                                  entry,
+                                  "text",
+                                  Gio.SettingsBindFlags.GET |
+                                  Gio.SettingsBindFlags.SET |
+                                  Gio.SettingsBindFlags.NO_SENSITIVITY,
+                                  get_method,
+                                  set_method)
+
```
samdroid-apps: All of the code above here seems like helpful code that could be put in a shared library.

After this patch is merged, it could be interesting to go on a refactoring run and put this in a shared library!  We could clean up a lot of control panels using this.

src/jarabe/main.py
```diff
@@ -347,6 +347,51 @@ def setup_fonts():
...
+                     'socks_proxy',
+                     'SOCKS_PROXY']
+    g_mode = Gio.Settings('org.sugarlabs.system.proxy').get_string('mode')
+    if g_mode == 'manual':
+        schemas = ['org.sugarlabs.system.proxy.http',
+                   'org.sugarlabs.system.proxy.https',
+                   'org.sugarlabs.system.proxy.ftp',
+                   'org.sugarlabs.system.proxy.socks']
+        counter = 0
+        for schema in schemas:
```
samdroid-apps: Can we just change this to:

    protos = ['http', 'https', ...]
    for proto in protos:
        name = '{}_proxy'.format(proto)
        schema = 'org.sugarlabs.whatever.{}'.format(proto)
        ...

would that be a little cleaner?

samdroid-apps:
This is amazing.  I love how it warns me when I try to set invalid settings.

It's so close to me merging it.  I just posted 2 small comments about the code.

Also, when I set the proxy to be empty, I feel like it should have no effect.  Eg. I set the FTP proxy to "", I don't get warned about it being an invalid server and it doesn't export the FTP_proxy environment variable.  Is that that is the behaviour that other systems use?  (note, I know nothing about proxy settings, CC @quozl)

But once you fix those little things I will merge.  It's magic and beautiful.

data/org.sugarlabs.gschema.xml
jvonau: use-same-proxy is a short way of saying use the same proxy for all other
protocals like https ftp and should auto populate the remaining proxy boxes
for those services.

ManashRaja: The "use-same-proxy" and "enabled" keys are not used now. But I left them, if it might be required in the future. Should I remove them?

extensions/cpsection/network/view.py
ManashRaja: Ya, I would work on it. I also added a new method bind_with_convert to gsettings which would be quite useful in many applications.

data/org.sugarlabs.gschema.xml
samdroid-apps: We can always add new settings later.  Please remove them now; they are a little confusing having unused settings.

ManashRaja:
Thanks @samdroid-apps ,
- I removed the unused gsetting keys
- Refactored the list part of 'main.py' to make it compact.
- Modified the behavior of alert so that if a user set blank "" host name to some type of proxy in manual settings, then it does not give you the alert. It might happen sometime that proxy settings are required for only some type like "http, https" instead of all. But error is shown when user leaves all the hostnames blank.

samdroid-apps:
Looks great.  I merged it.

It *probably* has some bugs.  But we are so far away from a realease, that we can fix them between now and then (hopefully).  I can make a testing release if anybody is interested?  (CC Fedora people)

PR #668, opened 2016-03-16
Fix custom restart message display in control panel

ManashRaja:
Fixed the issue where custom message for restart options, set through
'self.restart_msg = _('Your custom message')' in case of
'self.needs_restart'  set to True was not getting displayed and instead
the default 'Changes require restart' was displayed.

This was solved by simply passing the 'self.restart_msg' variable in
/jarabe/controlpanel/sectionview.py to 'alert.props.msg' variable in
/jarabe/controlpanel/gui.py where it is finally displayed.

fix-restart_msg


ManashRaja:
@samdroid-apps please review.

PR #670, opened 2016-04-01
Do not set the cursor globally

erilyth:
Issue where the cursor is not being reset even when we exit 'Sugar'.
Discussed on mailing list - '[Sugar-devel] Mouse pointer changes into sugar style pointer in ubuntu'

### DIFF
--- a/src/jarabe/desktop/activitieslist.py
+++ b/src/jarabe/desktop/activitieslist.py
@@ -72,18 +72,20 @@ class ActivitiesTreeView(Gtk.TreeView):
 
         for i in range(desktop.get_number_of_views()):
             column = Gtk.TreeViewColumn()
-            cell = CellRendererFavorite(i)
-            cell.connect('clicked', self.__favorite_clicked_cb)
-            column.pack_start(cell, True)
-            column.set_cell_data_func(cell, self.__favorite_set_data_cb)
+            self.cell_favorite = CellRendererFavorite(i)
+            self.cell_favorite.connect('clicked', self.__favorite_clicked_cb)
+            column.pack_start(self.cell_favorite, True)
+            column.set_cell_data_func(self.cell_favorite,
+                                      self.__favorite_set_data_cb)
             self.append_column(column)
 
-        cell_icon = CellRendererActivityIcon()
-        cell_icon.connect('clicked', self.__icon_clicked_cb)
+        self.cell_icon = CellRendererActivityIcon()
+        self.cell_icon.connect('clicked', self.__icon_clicked_cb)
 
         column = Gtk.TreeViewColumn()
-        column.pack_start(cell_icon, True)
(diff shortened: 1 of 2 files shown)


jvonau:
Some background https://github.com/sugarlabs/sugar/commit/ea803cb709587d506434e09c37fe9f4dfd2128d8

What if the user's starting 'cursor-theme' was not 'default', this would alter that user's choice of themes. Might want to read the current setting, save it and use the observed value to restore the user's previous setting in place of using 'default'.

erilyth:
@jvonau, It now gets reset to the user's original cursor.

jvonau:
should address inadvertent changing of a user setting, currently untested by myself.

quozl:
Would like also to support running /usr/bin/sugar in GNOME or Unity desktop; the cursor should be private to Sugar and activities, not infect all other desktop apps.

erilyth:
@quozl ,how do you think I should proceed to get that done? I am a little new to this part of Sugar, generally worked more with activities within Sugar.

quozl:
@erilyth, look for a more subtle way to set the cursor.  Changing GSettings is overkill, and dates from the time that Sugar was alone on a device.

davelab6:
I filed an issue on bugs.sugarlabs.org about this but it awaits moderation

quozl:
@davelab6, i'm not even sure who does moderation duties on bugs.sugarlabs.org.

davelab6:
> @davelab6, i'm not even sure who does moderation duties on bugs.sugarlabs.org.

Aha! :) 

@walterbender do you know who does moderate bugs.sugarlabs.org? 

When I submit tickets it says,

> Moderation required for new ticket.
>
> Thank you for submitting a new ticket ("Create sugar-event-recorder-daemon package as required part of sugar desktop").
>
> Your submission has been routed to Samuel Cantero for approval. We make every effort to review submissions within one working day.
>
> If you provided a valid e-mail address with your submission, you will receive e-mail notification if the moderator accepts your submission.

quozl:
Ah, okay, thanks.  Not being subject to moderation means I'm ignorant.  Samuel Cantero is regularly responsive on systems@ mailing list.

davelab6:
Okay cool, I'll post there and CC him as I found his email :)

erilyth:
@quozl There are ways to set the cursor using Gdk, but the problem is that the cursor needs to be set individually for every activity. Since the cursor is set on a given window. Would this be okay or should I look for something better?

quozl:
@erilyth, that's great, sounds like what we need.  So you'll be able to put that in sugar-toolkit-gtk3 in the activity common code.  Once you show us that, I can look at whether I can backport it to sugar-toolkit (gtk2).

erilyth:
@quozl , I'll send a PR to sugar-toolkit-gtk3

PR #671, opened 2016-04-04
Feature gsoc screenshot feature

iamutkarshtiwari:
This is just for testing purpose. The code is not formatted yet.

Preview- > http://imgur.com/0e1iq4v


quozl:
Please don't include GsoC metadata in the patches, or the pull request title, because reviewers are to ignore GsoC and evaluate based on normal developer expectations.  If you are required to do so for record keeping, please add any such metadata to the end of the pull request comment.  Ensure that the patch itself describes the problem and the solution in full, without external references.  Thanks!

Reviewed the preview; I like it.  Please consider swapping the two icons on the top right corner of the window, because our control panel (my settings) has them the other way around.

PR #666, opened 2016-03-09
Fix drag_status call to work on fc24/23, fixes #4949

samdroid-apps:
I believe that Gdk.drag_status (new function) has been here forever (tested fc25, fc22), since it doesn't have an "new since" section in the docs [1].

[1]  https://developer.gnome.org/gdk3/unstable/gdk3-Drag-and-Drop.html#gdk-drag-status

PR #673, opened 2016-04-05
Update MAINTAINERS

davelab6:
Based on https://github.com/sugarlabs/sugar/pull/666#issuecomment-205518462 I thought it would be good to update MAINTAINERS


quozl:
Reviewed.

PR #674, opened 2016-04-06
Fix for battery device icon update

quozl:
_(this is the updated commit message from 67ffe98)_

Frame battery device is static on Ubuntu 16.04, with UPower 0.99.4.

The 'changed' signal is not sent.  UPower had deprecated 'changed', replacing with 'notify' for all properties, or on per-property basis.

* switch to UPowerGlib GObject introspection API,

* continue to use 'changed' signals on older UPower, for compatibility with OLPC OS 13.2.7, Fedora 18, UPower 0.9.19,

- switch to 'notify' signals if they are present,

- continue to use D-Bus API to get device path, because the older UPower fails on get_devices method, and doesn't have get_display_device,

- ignore bursts of notify signals; reduces frame icon redraws,

- fixes https://bugs.sugarlabs.org/ticket/4781.

Tested with OLPC OS 13.2.7 on XO-4, and Ubuntu 16.04 on commodity hardware.

Requires gir1.2-upowerglib-1.0 on Ubuntu 16.04.

### DIFF
--- a/extensions/deviceicon/battery.py
+++ b/extensions/deviceicon/battery.py
@@ -20,6 +20,7 @@ from gettext import gettext as _
 from gi.repository import GLib
 from gi.repository import GObject
 from gi.repository import Gtk
+from gi.repository import UPowerGlib
 import dbus
 
 from sugar3 import profile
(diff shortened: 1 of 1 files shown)


quozl:
Rewrote patch;

* avoid checking for get_display_device, always use the D-Bus enumerate devices result,
* split out the connecting of signals into a private method,
* check the signals using GObject.signal_list_names,
* include test results.

extensions/deviceicon/battery.py
```diff
@@ -43,13 +44,9 @@
...
-_UP_STATE_UNKNOWN = 0
-_UP_STATE_CHARGING = 1
-_UP_STATE_DISCHARGING = 2
-_UP_STATE_EMPTY = 3
-_UP_STATE_FULL = 4
-_UP_STATE_CHARGE_PENDING = 5
-_UP_STATE_DISCHARGE_PENDING = 6
+_UP_STATE_UNKNOWN = UPowerGlib.DeviceState.UNKNOWN
+_UP_STATE_CHARGING = UPowerGlib.DeviceState.CHARGING
+_UP_STATE_DISCHARGING = UPowerGlib.DeviceState.DISCHARGING
```
samdroid-apps: Couldn't we just use the UPowerGlib constants directly?  Find and replace is good :)

quozl:
* use `UPowerGlib` constants directly instead of making our own based on them, reported by @samdroid-apps,
* confirmed `_UP_STATE`* were not used outside module.

PR #675, opened 2016-04-06
Screenshot popup alert

iamutkarshtiwari:
This feature add a screenshot alert to Sugar.
On pressing Alt+1 key combination or selecting 'Take a screenshot' option from bottom toolbar
a popup alert appears where user could can the default name of that screenshot.
The popup can be dismissed by clicking 'X' button or by pressing 'Escape' key.
Enter can also be used to save and dismiss the alert.

0e1iq4v


iamutkarshtiwari:
@quozl @davelab6 Please try this patch and share you suggestions. @samdroid-apps You might like it ;)

src/jarabe/screenshotpanel/gui.py
```diff
@@ -0,0 +1,341 @@
...
+    cr.set_operator(cairo.OPERATOR_SOURCE)
+    cr.paint()
+    cr.set_source_surface(screenshot_surface)
+    cr.paint()
+
+    preview_str = StringIO.StringIO()
+    preview_surface.write_to_png(preview_str)
+    
+    return dbus.ByteArray(preview_str.getvalue())
+
```
quozl: Extra blank lines, please remove.

src/jarabe/screenshotpanel/gui.py
```diff
@@ -0,0 +1,341 @@
+# Copyright (C) 2008 One Laptop Per Child
```
quozl: Is this a new file, and if so why are you assigning copyright to One Laptop per Child yet dating it 2008?  I'm fine with you assigning copyright to us, but if so please use the current year.

src/jarabe/screenshotpanel/gui.py
```diff
@@ -0,0 +1,341 @@
...
+    '''
+    Generates a pop papel to allow user to save the
+    screenshot by the name of his choice
+    '''
+
+    __gtype_name__ = 'ScreenshotPanel'
+
+
+    def __init__(self):
+        Gtk.Window.__init__(self)
```
quozl: Most of ```__init__``` seems remarkably familiar; is there an opportunity to refactor other similar classes?

src/jarabe/screenshotpanel/gui.py
quozl: Oh, perhaps you moved this file and didn't say so in the commit message?

quozl:
Looks great.

Worried about the commit message; during review it wasn't clear what you had done and why.

iamutkarshtiwari:
Fixed the commit message. Fixed the blank lines. Updated the copyright.

quozl:
I'm hopeful of receiving feedback from others on whether the dialog should be styled to look like either;
* the wireless password prompt (no X or tick decorators), or
* the control panel, view source, or help modal windows.

quozl:
It is proving difficult to get any feedback, sorry for the delay.

quozl:
While we wait for feedback ... tested on Ubuntu 16.04 with Sugar 0.108.

**appearance issues**
* the "Name" label is unnecessary; we don't have a name label in the Journal detail view, or the activity toolkit text entry for journal object; it is obvious from context that it is a journal entry name,
* if after consensus the "Name" label is kept,
 * it is yellow, it should be white,
 * it is too close to the text entry,
 * it might be to the left of the text entry, like "Server:' in My Settings, Network, Collaboration,
* the preview image is distorted; because the aspect ratio of the preview doesn't match the aspect ratio of the display; you've coded 200:160 (1.5) instead of detecting the display, (you've also used these constants at least twice in the code without using a variable for them),

**interaction issues - keyboard**
* the dialog does not respond to the `Esc` key, it must dismiss the dialog and cancel the screenshot, see [Hotkeys](http://wiki.sugarlabs.org/go/Hotkeys),
* the dialog does not respond to the `Enter` key, it must accept the dialog and save the screenshot,

**interaction issues - other windows**
* while the dialog is visible, the frame can be shown with the `f6` key, but does not react to input,
* while the dialog is visible, the `f1`, `f2`, `f3` keys will hide the dialog and switch view, but the view does not react to input, the user is trapped with no obvious way to fix, but any key press shows the dialog again,
* while in the home view, activating the My Settings control panel (`alt-shift-m`), and then pressing `alt+1` locks the display; the control panel cannot be closed, workaround is to restart display manager `service lightdm restart`,
* while in the home view, activating the Help window (`alt+shift+h`), and then pressing alt+1 locks the display, as above.

Hope that helps!

quozl:
Unwelcome side-effect; when a menu palette is open, such as the central icon on the home view, the `alt+1` key causes the palette to close.

davelab6:
> Unwelcome side-effect; when a menu palette is open, such as the central icon on the home view, the alt+1 key causes the palette to close.

Is that a GTK level issue?

quozl:
> Is that a GTK level issue?

No, it's a Sugar Python code issue.  GTK will do whatever we tell it to do.

davelab6:
> No, it's a Sugar Python code issue. GTK will do whatever we tell it to do.

Should it be solved in this PR?

quozl:
> Should it be solved in this PR?

Yes, because this PR causes the regression.  It doesn't happen without this PR.

samdroid-apps:
Maybe we could move the screenshotpopup module into jarabe.view.  I don't want to clutter the root jarabe namespace.

Also, please add the screenshpotpopup directory to the parent's Makefile.am's SUBDIRS.

samdroid-apps:
Also also add the new directory to configure.ac

iamutkarshtiwari:
@samdroid-apps I'll do as directed but I need your little help with this patch. Wherever I press the <Alt>+<1> key combination more than once, more than one popup gets generated over one another. 

Whereas what I expect is that it should generate only one popup even if the user presses the key combination more than once.

samdroid-apps:
Yes, only one of these popups should be shown at a time.  You might be interested in sharing this logic between the screenshot and settings popups.



Visually, I think it has regressed.  The toolbar buttons were more consistent than using the traditional buttons as you do now.

Additionally, you did not fix any issues that @Quozl posted in:  https://github.com/sugarlabs/sugar/pull/675#issuecomment-209190086

You may be interested in @AbrahmAB's patch to move most of the modal popup logic into a common base class.  You should co-ordinate.  I don't want any new modal popups that do more of this copy paste code reuse.

iamutkarshtiwari:
@samdroid-apps Module rewritten using Abhijit's popup module.

src/jarabe/screenshotpanel/gui.py
```diff
@@ -0,0 +1,342 @@
+#!/usr/bin/env python
```
samdroid-apps: Can you move this to jarabe.gui.screenshotpanel or something?  Having a whole folder for 1 file is unnecessary.

iamutkarshtiwari: I'll do it but before that we need to patch that ```grey``` rectangle bug.

samdroid-apps: It is blocking me from testing :)

iamutkarshtiwari:
@samdroid-apps Issue resolved ;) :+1:

AbrahmAB:
After taking Screenshot of Control Panel and while typing in the text entry of the pop-up, the search_entry of Control Panel toolbar steals the focus!
The user needs to close the Control Panel to write in the text-entry of the pop-up.

samdroid-apps:
Looks so good.  Here is my feedback:

* Sorry for my miscommunication.  I intended to say `jarabe.view.screenshotpopup`, not `jarabe.gui`.
* Instead of having the yes/no buttons, can you just add a tick ("Save Screenshot") button to the toolbar?
* Then entry is very short.  Can you put it above the screenshot, with the label to the left of the entry.  Like:

```

    | Save Screenshot         [Save] [X] |
    |-----------------------------------|
    | Title:  |Screenshot Of XYZ......|  |
    | .________________________________. |
    | |THUMBNAIL                       | |
    | |                                | |
    | |                                | |
    | |________________________________| |
```

* The thumbnail seems static for me.  I took a screenshot of the home view, and got the right thumbnail.  Then I opened Log, took a screenshot and I got the same homeview thumbnail.  It saved the correct screenshot though.
* I can reproduce Abhijit's issue with the settings popup
* The thumbnail looks squashed on my laptop.  My screen is a 16:9 aspect ratio, not 4:3.  It would be interesting to fix it.

src/jarabe/gui.py
```diff
@@ -0,0 +1,338 @@
...
+            jobject.metadata['icon-color'] = color
+            jobject.metadata['mime_type'] = 'image/png'
+            jobject.file_path = self.file_path
+            datastore.write(jobject, transfer_ownership=True)
+        finally:
+            jobject.destroy()
+            del jobject
+
+
+def generate_thumbnail(screenshot_surface):
```
samdroid-apps: This is coppied from somewhere right (which is fine)?

If so, can you delete it in the other place?  I don't want to have 2 identical generate_thumbnail functions lying around the codebase.

src/jarabe/gui.py
```diff
@@ -0,0 +1,338 @@
...
+
+    if content_title is None:
+        title = _('Screenshot')
+    else:
+        title = _('Screenshot of \"%s\"') % content_title
+
+    return preview_image, title
+
+
+def take_screenshot():
```
samdroid-apps: Same comment as for the `generate_thumbnail` function.

src/jarabe/gui.py
```diff
@@ -0,0 +1,338 @@
...
+from gi.repository import Gtk
+from gi.repository import Gdk
+from gi.repository import Gio
+
+from sugar3 import env
+
+from sugar3.datastore import datastore
+from sugar3.graphics.icon import Icon
+from sugar3.graphics import style
+from sugar3.graphics.alert import Alert, TimeoutAlert
```
samdroid-apps: Unused imports.  Can you just run flake8 again?

src/jarabe/gui.py
iamutkarshtiwari: Yes. This function is a modified version of the original ```generate_thumbnail()``` so the code is unique in itself.

iamutkarshtiwari:
@samdroid-apps  Is there any way to fix that deadlock of this popup with Settings windows? I am eager on improving this PR.

src/jarabe/gui.py
iamutkarshtiwari: Will do.

quozl:
@iamutkarshtiwari, any news on this one?

iamutkarshtiwari:
@quozl There were some UI changes proposed by Sam which I haven't looked into yet (been busy with other projects) but will soon get back to it.

i5o:
@iamutkarshtiwari if you need help let me know.

We can also make a GCI Task for this feature.

iamutkarshtiwari:
@i5o I would really love to get this PR fixed. Sure, we can make a GCI task for this feature. 

Things to be done-
1) Changes to be made as suggested by https://github.com/sugarlabs/sugar/pull/675#issuecomment-226890680
2) This popup seems to get deadlocked by 'settings' popup - https://github.com/sugarlabs/sugar/pull/675#issuecomment-226862041. We can get this fixed as well.

Let me know if you need more info.

quozl:
Handing an incomplete pull request to a student for GCI seems unfair to me.  Huge complexity to deal with.  I'm fine with you finishing up your work and merging it.  :smile:

iamutkarshtiwari:
@quozl @tony37 Give me some time, I'll patch this one soon.

quozl:
Thanks.  Also please list what was changed from feedback.

jamessandy:
@iamutkarshtiwari  they is a way of fixing the deadlock

iamutkarshtiwari:
@jamessandy Thanks for your interest in this PR. Your suggestions are welcome :)

jamessandy:
Lets begin

PR #678, opened 2016-04-12

erilyth:
There was a problem with the old PR, for some reason it showed that I wanted to merge from "unknown repository", so my updates weren't being reflected there.

@samdroid-apps Sorry for taking so long, I had exams till day before.
I fixed the problem with the pixbufs. This was happening mainly with the bmp images, but it works now.

Old PR - https://github.com/sugarlabs/sugar/pull/639

### DIFF
--- a/src/jarabe/frame/activitiestray.py
+++ b/src/jarabe/frame/activitiestray.py
@@ -24,6 +24,7 @@ import os
 from gi.repository import GObject
 from gi.repository import Gio
 from gi.repository import Gtk
+from gi.repository import Gdk
 
 from sugar3.graphics import style
 from sugar3.graphics.tray import HTray
@@ -50,6 +51,10 @@ from jarabe.frame.frameinvoker import FrameWidgetInvoker
 from jarabe.frame.notification import NotificationIcon
 from jarabe.frame.notification import NotificationButton
 from jarabe.frame.notification import NotificationPulsingIcon
+from jarabe.frame.clipboardmenu import ClipboardMenu
+from jarabe.frame.clipboardobject import ClipboardObject
+from jarabe.frame.clipboardtray import _ContextMap
+from jarabe.frame.clipboardobject import Format
 import jarabe.frame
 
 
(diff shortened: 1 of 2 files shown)


erilyth:
@samdroid-apps any updates on this? Did you find a bug?

src/jarabe/frame/activitiestray.py
```diff
@@ -90,6 +107,49 @@ def __init__(self, home_activity, group):
...
+        self._context_map.add_context(context, 0, len(context_targets))
+
+        self._cb_object = ClipboardObject(0, "")
+
+        for target in context_targets:
+            if str(target) not in ('TIMESTAMP', 'TARGETS', 'MULTIPLE'):
+                widget.drag_get_data(context, target, time)
+
+        Gdk.drop_finish(context, True, Gtk.get_current_event_time())
+        if hasattr(self, "_cb_object") and self._cb_object is not None:
```
samdroid-apps: do you need the hasattr checks?  You set it to None at the top right?

src/jarabe/frame/activitiestray.py
```diff
@@ -90,6 +107,49 @@ def __init__(self, home_activity, group):
...
+        if selection.get_pixbuf() is not None:
+            data = selection.get_pixbuf()
+        if len(selection.get_uris()) != 0:
+            data = selection.get_uris()[0]
+        if selection.get_text() is not None:
+            data = selection.get_text()
+        if data is None:
+            data = selection.get_data()
+
+        if (data is not None and hasattr(self, "_cb_object") and
```
samdroid-apps: hasattr check again :(

samdroid-apps:
It looks really good.

It seems to have some weird conditions with drags from browse.  If I drag an image from browse to the journal, it fails [1].  But if I go browse->clipboard then clipboard->browse it works.

Also, using BMP is interesting, I'm interested to hear the rational behind this.  Isn't BMP pretty bad on compression?

But wow, I'm so sorry that you're working on this.  Gtk drag n drop looks pretty horrid :(

i5o:
"any updates on this? "

erilyth:
@i5o Sorry I totally forgot about this. I spent a while but I'm kind of stuck on this, I don't really understand why dragging from browse to the journal directly fails but browse->clipboard etc work (as @samdroid-apps mentioned).

chimosky:
Moved to #960.

PR #679, opened 2016-04-16
Connect to multiple XS server easily, fixes #362

ManashRaja:
Issue: As mentioned in the bug [1].
It is difficult to connect a Sugar device to multiple XS servers because
the user is required to perform few manual steps to clear a previous
registration. Also, new registration is performed even  when a laptop
tries to connect to an XS server, where it has already registered
previously. This caused the original backup-url and back-up path to
change on the server causing inefficiency.

Solution: The solution involves a few automated checks and tasks that
are performed to overwrite new registration details, store previous
registration details and restore the details is an old server is connected.

UI: Just the 'Register' and 'Register again' button are replaced with
a single 'Connect to server' button on the FavouriteView desktop. It was
done to provide a non confusing simple and constant button to the user, as
his/her final motive is to get Sugar settings configured with the server
regardless of the fact that it is registering, registering again or
simply restoring settings of previous registration. Also it will mark
the change in the background functionality.

Server: I tried it on XSCE School Server release-6.0 on a CentOS7 machine.
No server side modification was done to achive it apart from enabling
'Services for XO Laptops such as Registration, Backup, and Activity Server'
in 'Configure' section of schoolserver admin. (http://schoolserver/admin/)
I used webpage available at schoolserver:5000 prvided by xs-authserver,
where the list of registered laptops are present.

Here is a related feature page for details of this enhancement [2].

A link to the starting of a discussion thread about this feature. [3].

[1] https://bugs.sugarlabs.org/ticket/362
[2] https://wiki.sugarlabs.org/go/Features/Multi_XS-server_registration
[3] http://lists.sugarlabs.org/archive/sugar-devel/2016-April/051800.html


ManashRaja:
The feature at present:
sugar-server-present

This patch follows the following flow :)
sugar-server-modified

The main advantages of the above flow according to me are:
- Manual clear of registration is not required.
- Each laptop will have same backup url and backup path (in XSCE:/library/users) for a particular XSCE as it was created on first registration with the server. As registration will be done only once. The "Register" button would rather say "Connect to Server". When user connects to an XSCE where it has registered previously (identified by the presence of pubkey), it would simply set its serial number, uuid and backup-url with the pre-registration data. This would enable better and advanced management of backups.
- No extra input is required from the user. The user just has to connect to the XSCE server ssid and specify the jabber-server address as before.

I am using here XSCE, and I haven't tested it for an XS. So please pardon the tittle to be XS as I wasn't clear about the difference. I will try installing an XS for the testing though.

ManashRaja:
Better known_host handling has been added. Please check https://github.com/sugarlabs/sugar/pull/679/commits/930726741744ebdf6d464b84b31210aa0897124a

ManashRaja:
Multiple commits squashed into one.
Please review and comment.

src/jarabe/desktop/schoolserver.py
```diff
@@ -56,7 +61,25 @@ def _generate_serial_number():
     return serial
 
 
-def _store_identifiers(serial_number, uuid_, backup_url):
+def _get_history_for_serial(serial_number):
+    identifier_path = os.path.join(env.get_profile_path(), 'identifiers')
+    if not os.path.exists(identifier_path):
+        os.mkdir(identifier_path)
```
quozl: These three lines are duplicate of code in `_store_identifiers`, please refactor.

src/jarabe/desktop/schoolserver.py
```diff
@@ -56,7 +61,25 @@ def _generate_serial_number():
...
+
+    file_path = os.path.join(identifier_path, 'server_history')
+    if os.path.exists(file_path):
+        json_file = open(file_path, 'r')
+        data = json.loads(json_file.read())
+        json_file.close()
+        if serial_number in data:
```
quozl: These lines (join a path, check exists, open, load, close) are also closely related to the other lines you add below, please refactor into a read from a file in identifiers path, returning data.

src/jarabe/desktop/schoolserver.py
```diff
@@ -80,9 +103,26 @@ def _store_identifiers(serial_number, uuid_, backup_url):
...
 
+    file_path = os.path.join(identifier_path, 'server_history')
+    if os.path.exists(file_path):
+        server_history_file = open(file_path, "r")
+        data = json.load(server_history_file)
+        server_history_file.close()
+        data[serial_number] = {}
+        data[serial_number]["uuid"] = uuid_
+        data[serial_number]["backup_url"] = backup_url
+    else:
```
quozl: These three lines are identical to the last three lines on the `else` branch of the `if`, please move them out of the conditional clauses.

src/jarabe/desktop/schoolserver.py
```diff
@@ -127,39 +170,79 @@ def register_laptop(url=_REGISTER_URL):
...
+        if data['success'] != 'OK':
+            logging.error('Registration: server could not complete request: %s',
+                          data['error'])
+            raise RegisterError(_('The server could not complete the request.'))
+
+        # Registration Successful, hence we can add the identification of this
+        # server to our known_host file.
+        command = "ssh-keyscan -H -t ecdsa " + jabber_server
+        output = subprocess.check_output(command, shell=True)
+        os.system("mkdir -p ~/.ssh")
```
quozl: * there's no need to assume that ~ may be missing (that is, don't use -p),
* python can call `mkdir` directly,
* the mask is not set, which may break later inbound SSH on this system,
* there's no checking of status returned by these steps,
* it isn't clear that an echo to known_hosts is safe; is there a program provided by OpenSSH that provides interlocked access?

quozl:
Several typos in commit message, please spell check.

src/jarabe/desktop/schoolserver.py
ManashRaja: - .ssh directory may be missing. Hence I used -p.
- ya, i used ```os.system("mkdir -p ~/.ssh")``` instead of 
```if not os.path.exists(...):
    os.makedirs(...)```
for its simplicity. Shall i use the later?
- I didn't get the it. How would it break inbound SSH into the system?
- while calling ```ssh-keyscan``` we already know that the server is present and ```jabber_server``` value has correct address due to successful registration, I think assuming and checking for error will be redundant. The output of the command if executed successfully will either be the identity key or blank. that is why I didn't check it before, but now have added a condition to write only if output has value.
- I am not aware of the security of echoing to known_hosts. I replaced it ```with _write_to_file``` with append mode

**Comment by ManashRaja** [comment]
@quozl Thanks for the in depth review and guidance. I have refactored the code and removed typos in commit message.

**Comment by quozl** [comment]
No worries, glad to be of help.  I've not reviewed your latest.  Best of luck.

**Comment by ManashRaja** [comment]
Please review

File: src/jarabe/desktop/schoolserver.py [hunk 5]
```diff
@@ -122,44 +171,85 @@ def register_laptop(url=_REGISTER_URL):
...
-    finally:
-        socket.setdefaulttimeout(None)
-
-    if data['success'] != 'OK':
-        logging.error('Registration: server could not complete request: %s',
-                      data['error'])
-        raise RegisterError(_('The server could not complete the request.'))
-
-    settings.set_string('jabber-server', data['jabberserver'])
+        response = urllib2.urlopen(db_url)
```
samdroid-apps: Please use libsoup for opening urls.  Libsoup properly intergrates with the gobject mainloop; so that the ui is still responsive while we wait on the network.

Lib soup has documentation, but you can probs just copy some code from my reddit client;  https://github.com/samdroid-apps/something-for-reddit/blob/master/src/api.py#L182

ManashRaja:
Replaced urllib2 by libsoup. Shall I squash the commits to one?

src/jarabe/desktop/schoolserver.py
ManashRaja: @samdroid-apps Please review

pro-panda:
I don't have means to test this. Can someone review it again? (@samdroid-apps @quozl)

tony37:
This issue should be closed as not a problem. A lot of design and 
implementation effort has already been expended on a non-issue.

The original issue was raised by someone who apparently did not know how 
to deal with the known-host issue encountered by developers and system 
administrators who need to access more than one server with the same 
host name. The solution on an XO is trivial:

#remove known hosts
sudo rm -rf /home/olpc/.ssh/known_hosts

This problem is totally unrelated to Sugar.

What would be helpful is to eliminate the registration button 
completely. Have Sugar register every time it connects to a server (as 
has been done by OLE Nepal for over a decade). If that registration 
fails, no harm is done. If the laptop is not registered, it won't be 
able to back up the Journal. If registered the data will not be updated 
but will ordinarily be valid.

Tony

quozl:
Thanks Tony.  Was the change to Sugar by OLE Nepal distributed generally to end users?  If so, please point us at the source code?

tony37:
Hi, James

I don't believe it was submitted to SugarLabs specifically. It is part 
of a number of changes which were made to fit the needs there - such as 
using Firefox as the browser, operating Firefox in full-screen mode, and 
so on. However, essentially the only change needed is to execute the 
registration button function as part of the connect procedure. The 
server-side code simply repeats the registration process on its side 
causing no change if the information supplied is the same.

I am sure the OLE Nepal folks will be happy to explain the change.

Tony

PR #681, opened 2016-04-18
Send a file to multiple friends: Fixes #1471

AbrahmAB:
pop-up window with the listview of the friends online is displayed,
multiple friends cn be selected and pressing the send toolbutton sends the file to selected friends.

Note: used the PopWindow class of sugar3.graphics as implemented in commit-> sugar-toolkit-gtk3:3d61134

### DIFF
--- a/src/jarabe/journal/palettes.py
+++ b/src/jarabe/journal/palettes.py
@@ -34,8 +34,9 @@ from sugar3.graphics.xocolor import XoColor
 from sugar3.graphics.alert import Alert
 from sugar3 import mime
 from sugar3 import profile
+from sugar3.graphics.popwindow import PopWindow
+from sugar3.graphics.toolbutton import ToolButton
 
-from jarabe.model import friends
 from jarabe.model import filetransfer
 from jarabe.model import mimeregistry
 from jarabe.journal import misc
@@ -43,6 +44,7 @@ from jarabe.journal import model
 from jarabe.journal import journalwindow
 from jarabe.webservice import accountsmanager
 from jarabe.journal.misc import get_mount_color
+from jarabe.view.friendlistview import FriendListView
 
 
 class ObjectPalette(Palette):
@@ -125,9 +127,7 @@ class ObjectPalette(Palette):
         self.menu.append(menu_item)
         menu_item.show()
 
-        friends_menu = FriendsMenu()
-        friends_menu.connect('friend-selected', self.__friend_selected_cb)
-        menu_item.set_submenu(friends_menu)
+        menu_item.connect('activate', self.__friendlist_popup_cb)
 
         if detail is True:
             menu_item = MenuItem(_('View Details'), 'go-right')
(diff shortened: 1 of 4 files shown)


samdroid-apps:
This is a very good looking patch.  Just a few comments:

* Please put the title of the document that the user is about to send in the title.  Eg "Send <i>Cool Doc</i> to..."

* Please consider the empty state (no friends).  You could display an empty view, like is used in the journal

* Maybe list everybody (not just friends), but put friends first

* Maybe put a search bar at the top of the popup, so that I can search for a friend I want.  If I press enter, it should select the top friend.

* The dialog still stays on top even after I launch another activity.  Maybe it should only be on top when I am in the journal

* Shouldn't the close button and the send button swap places?

* "Send to" in the menu could become "Send to..."

* Maybe clicking anywhere on a row should toggle it's checkbox

* The toolbuttons ("Send" "Close") don't have the roll over palettes (the type of ones that say "Send" or "Close")

i5o:
ping..?

Also, I'm not able to merge, please update the code @AbrahmAB

quozl:
Rebasing to resolve the conflicts doesn't look hard.  Two new files, and the rest of the changes are straightforward.  Then respond to @samdroid-apps comments above.

MrSpideyNihal (commented):
To permit sending mesaages to more than one friends in the context of the furnished code, you may need to make adjustments to the existing functionality associated with buddy choice and report switch initiation. Here's a trendy technique you may follow:
Modifying FriendListView:
Enable Multiple Selection:
Update the FriendListView to permit more than one friend selection. Modify the choice mode in the Gtk.TreeSelection associated with the tree view to allow a couple of alternatives.
Modify Callbacks:

Adjust the callback functions related to toggling friend selection (__cell_select_toggled_cb, and so on.) to address a couple of decided on buddies rather than simply one.
Updating File Transfer Logic:
Revise the File Transfer Function:
Modify the __friend_selected_cb method within the ObjectPalette elegance (or the relevant segment wherein the record transfer is initiated) to iterate through all the selected pals and start document transfers to every of them.
Signal Emission:
Emit Signal with Multiple Friends:
Ensure that after the &quot;Send&quot; motion is brought about, the sign emitted incorporates records approximately all the chosen friends.
Example Changes (Assuming Single Friend Transfer Logic):
python
Copy code
# Inside FriendListView or ObjectPalette elegance:

def __friend_selected_cb(self, menu_item, pals):
    file_name = version.Get_file(self._metadata['uid'])
    title = &quot;File Transfer&quot;
    description = &quot;Sending file&quot;
    mime_type = mime.Get_for_file(file_name)

    for pal in friends:
        filetransfer.Start_transfer(buddy, file_name, identify, description, mime_type)
This code assumes that __friend_selected_cb is wherein the document switch is initiated. Here, friends is believed to be a listing of selected pals.

To enforce this, you need to make certain that the selection handling logic in FriendListView supports a couple of choices and that the essential changes are made inside the strategies triggering the file transfer

quozl:
Off-topic comment hidden.  Next action is to rebase the commits.

PR #683, opened 2016-04-20
Don't contruct ActivityBundle for the journal, fixes #4959

samdroid-apps:
Ticket url:  https://bugs.sugarlabs.org/ticket/4959


ManashRaja:
Reviewed. :)

quozl:
* commit message typo *contruct* should be *construct*,
* the expression is the inverted form of that used in the next `if`, so reformat the code to make this clearer, with the most frequent case executing first.

samdroid-apps:
@quozl force pushed and attempted to address your comments.  Please review again.

PR #686, opened 2016-04-30
Smooth animations try 4

samdroid-apps:
The toolkit part of this feature was merged recently.  This is a very small patch to make use of the additions in jarabe

Feature: http://wiki.sugarlabs.org/go/Features/Tick_based_animation


davelab6:
Looks good

samdroid-apps:
Given that this patch has existed for ~1 year in many different forms, I will just merge it as nobody else is going to.

PR #687, opened 2016-05-23
Erase bundle fixes

quozl:
Add sugar-erase-bundle next to sugar-install-bundle.

Expose system activity after user erase; when a user activity is erased using "rm -rf" or sugar-erase-bundle, any system activity was not exposed.  Handle a FileMonitorEvent.DELETED by also looking for the same activity in the system activity directories.

### DIFF
--- a/src/jarabe/model/bundleregistry.py
+++ b/src/jarabe/model/bundleregistry.py
@@ -140,6 +140,24 @@ class BundleRegistry(GObject.GObject):
             self.add_bundle(one_file.get_path(), set_favorite=True)
         elif event_type == Gio.FileMonitorEvent.DELETED:
             self.remove_bundle(one_file.get_path())
+            for root in GLib.get_system_data_dirs():
+                root = os.path.join(root, 'sugar', 'activities')
+
+                try:
+                    dir_list = os.listdir(root)
+                except OSError:
+                    logging.debug('Can not find GLib system dir %s', root)
+                    continue
+                activity_dir = os.path.basename(one_file.get_path())
+                try:
+                    bundle = bundle_from_dir(os.path.join(root, activity_dir))
+                except MalformedBundleException:
+                    continue
+
+                if bundle is not None:
+                    path = bundle.get_path()
+                    if path is not None:
+                        self.add_bundle(path)
 
     def _load_mime_defaults(self):
         defaults = {}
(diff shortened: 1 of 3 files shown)


src/jarabe/model/bundleregistry.py
```diff
@@ -140,6 +140,23 @@ def __file_monitor_changed_cb(self, monitor, one_file, other_file,
...
+                    continue
+                activity_dir = os.path.basename(one_file.get_path())
+                try:
+                    bundle = bundle_from_dir(os.path.join(root, activity_dir))
+                except MalformedBundleException:
+                    continue
+
+                path = bundle.get_path()
+                if path is not None:
+                    self.add_bundle(path)
```
samdroid-apps: The bundleregistry `uninstall` method does the logic a little differently:

```
        alt_bundles = self.get_system_bundles(act.get_bundle_id())              
        if alt_bundles:                                                         
            alt_bundles.sort(                                                   
                key=lambda b: NormalizedVersion(b.get_activity_version()))      
            alt_bundles.reverse()                                               
            new_bundle = alt_bundles[0]                                         
            self.add_bundle(new_bundle.get_path())                              
```

Can we unify the logic in a shared function?

Other than that, thanks for the patches!  I love the sugar-erase-bundle script.

quozl:
> Can we unify the logic in a shared function?

Yes, go ahead, I'm happy to test what you come up with.

Re: feature label; this pull request also fixes a bug with removal of activities independently of sugar-erase-bundle, so the bug label might also be added.

src/jarabe/model/bundleregistry.py
```diff
@@ -140,6 +140,23 @@ def __file_monitor_changed_cb(self, monitor, one_file, other_file,
...
+                except OSError:
+                    logging.debug('Can not find GLib system dir %s', root)
+                    continue
+                activity_dir = os.path.basename(one_file.get_path())
+                try:
+                    bundle = bundle_from_dir(os.path.join(root, activity_dir))
+                except MalformedBundleException:
+                    continue
+
+                path = bundle.get_path()
```
samdroid-apps: I'm getting errors for when `bundle is None`.  Maybe add a check for that in there?

I also now understand why you didn't keep the logic the same between the 2 functions.  My bad for making that comment before.

quozl:
> I'm getting errors for when bundle is None. Maybe add a check for that in there?

Added.  Force pushed.

PR #688, opened 2016-05-23
Revert "Fixed #3146: Dragging a Journal entry drags the whole white e…

quozl:
…ntry bar, not just the icon"

This reverts commit 89f460b88d84c0887c0b458cf2bcffd0a55c3427.

Breaks on Fedora 18.  Cursor changes to "+" at the start of the drag,
but the drag survives only a very short distance before the cursor
returns to normal, and there is no floating icon following the cursor.

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -23,7 +23,6 @@ from gi.repository import GObject
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import Pango
-from gi.repository import GdkPixbuf
 
 from sugar3.graphics import style
 from sugar3.graphics.icon import Icon, CellRendererIcon
@@ -204,12 +203,10 @@ class BaseListView(Gtk.Bin):
         model.deleted.connect(self.__model_deleted_cb)
 
     def enable_drag_and_copy(self):
-        self.tree_view.drag_source_set(Gdk.ModifierType.BUTTON1_MASK,
-                                       [Gtk.TargetEntry.new(
-                                           'text/uri-list', 0, 0),
-                                        Gtk.TargetEntry.new(
-                                            'journal-object-id', 0, 0)],
-                                       Gdk.DragAction.COPY)
+        self.tree_view.enable_model_drag_source(Gdk.ModifierType.BUTTON1_MASK,
+                                                [('text/uri-list', 0, 0),
+                                                 ('journal-object-id', 0, 0)],
+                                                Gdk.DragAction.COPY)
 
     def disable_drag_and_copy(self):
         self.tree_view.unset_rows_drag_source()
@@ -719,13 +716,6 @@ class ListView(BaseListView):
         return self._is_dragging
 
     def __drag_begin_cb(self, widget, drag_context):
-        path, _column = self.tree_view.get_cursor()
-        if path is None:
-            return
-
-        row = self.tree_view.get_model()[path]
-        _pixbuf = GdkPixbuf.Pixbuf.new_from_file(row[ListModel.COLUMN_ICON])
-        self.tree_view.drag_source_set_icon_pixbuf(_pixbuf)
         self._is_dragging = True
 
     def __drag_data_get_cb(self, widget, context, selection, info, time):


samdroid-apps:
Isn't no icon still better than the previous behaviour?  Isn't it easier to drag a cursor than a screen-width bar?

Did you find the root cause for this issue?

Does this effect any other platforms?  If not, I'd prefer for you to keep this out of the upstream, and keep it as an OLPC patch.

quozl:
> Isn't no icon still better than the previous behaviour?

No.

> Isn't it easier to drag a cursor than a screen-width bar?

No.  At least it gave feedback and completed the drag.

> Did you find the root cause for this issue?

No.

> Does this effect any other platforms? If not, I'd prefer for you to keep this out of the upstream, and keep it as an OLPC patch.

Thanks, I'll let my customers know.  Please advise any other OLPC features you plan to remove?  There are lots of them.

samdroid-apps:
On Fri, May 27, 2016 at 4:52 PM, James Cameron 
<notifications@github.com> wrote:
>> Isn't no icon still better than the previous behaviour?
>> 
> No.
> 
>> Isn't it easier to drag a cursor than a screen-width bar?
>> 
> No. At least it gave feedback and completed the drag.
> 
>> Did you find the root cause for this issue?
>> 
> No.
> 
>> Does this effect any other platforms? If not, I'd prefer for you to 
>> keep this out of the upstream, and keep it as an OLPC patch.
>> 
> Thanks, I'll let my customers know. Please advise any other OLPC 
> features you plan to remove? There are lots of them.
> 

I understand that there are many OLPC features, and I think that it is 
important that we keep them.

This patch seems much more related to the old Gtk+ version that OLPC 
ships.  Currently, Sugar does everything we can to support Gtk+ 3.6.  
Sugar, and some activities, are more complex due to the lack of 
GtkFlowBox, GtkListBox, GtkStack, GtkListBox, GtkPopover, etc support.  
It is suboptimal to support Gtk3.6, but we do it for OLPC.

However, I don't support removing a *bug fix* from upstream Sugar 
because it has a small negative effect on a very specific and outdated 
Gtk+ version.  Correct me if I'm wrong, but this patch still makes the 
experience better for SoaS users, Fedora users, Debian users, and even 
your Ubuntu users?

There is a fine line between supporting OLPCOS and targeting OLPCOS 
exclusively.  Ever since Sugar Labs was separated from OLPC, I was 
under the impression that we aimed to support OLPCOS, in addition to 
other platforms like SoaS or Ubuntu.  Is that true?

godiard:
Would be possible do a new patch solving the issue only if Gtk > that the version shipped in F18?

quozl:
This pull request was a fix for a new problem introduced in Sugar 0.109.0 on the XO-1, that during a drag in the journal there is no drawing that accompanies the cursor during the drag.

The cause has been identified as 89f460b88d84c0887c0b458cf2bcffd0a55c3427 by @AbrahmAB in pull request #656.  Removing 89f460b88d84c0887c0b458cf2bcffd0a55c3427 fixes the problem.  The problem does not occur on XO-1.5 or XO-4 using the same versions of Gtk and Sugar.

The patch didn't explain how it was solving the original problem in [#3146](https://bugs.sugarlabs.org/ticket/3146).  I raised this in #656 but there was no response.  I still don't know how it was solving the problem, though I agree it did solve the problem on Ubuntu and Fedora other than on the XO-1.

The code that was changed was left over from before the Gtk3 port.

It looks like @AbrahmAB was limiting their involvement to rebasing the earlier work by @cristian99garcia in #435.  @AbrahmAB stopped responding on #656 and hasn't responded here.

Thank you for supporting Gtk+ 3.6, however it looks like the root of the problem has something to do with XO-1 not Gtk+.  Should you wish to support XO-1, ensure that pull requests are tested on XO-1 before merging.  Apart from me, I don't know of anybody else willing to do that.

On the assumption that Sugar Labs is dropping support for XO-1, I'll close this pull request.  Thanks for your time!

davelab6:
> On the assumption that Sugar Labs is dropping support for XO-1, I'll close this pull request. Thanks for your time!

Why are you assuming this?

davelab6:
> Should you wish to support XO-1, ensure that pull requests are tested on XO-1 before merging. Apart from me, I don't know of anybody else willing to do that.

I think it is essential that Sugar's stable releases be tested on XO-1s and Sugar Labs support them. 

I support merging this PR.

samdroid-apps:
>Thank you for supporting Gtk+ 3.6, however it looks like the root of the problem has something to do with XO-1 not Gtk+. Should you wish to support XO-1, ensure that pull requests are tested on XO-1 before merging. Apart from me, I don't know of anybody else willing to do that.

I am not an OLPC contractor, so I don't have an incentive to test on xo1.  Additionally, I do not have an XO1 if I wished to test.

Given that this only breaks on an outdated gtk on a specific hardware, maybe it doesn't make sense to hurt XO1.5, XO1.75 or XO4 users by reverting a patch that helps them? 

Also, please don't blame Abhijit.  He tried to help you, which is more than he is required to do.

davelab6:
I'm not blaming Abhijit! :) But I think the incentive is inherent in the fact that there are 100,000s of kids using XO-1s. 

I can arrange you to get hold of an XO-1, I'll email you.

davelab6:
I think it @godiard had a great suggestion:

> Would be possible do a new patch solving the issue only if Gtk > that the version shipped in F18?

I suggest reverting the patch since it is a small one, and filing an issue to rework the patch to include what @godiard suggests

PR #689, opened 2016-05-23
metacity shouldn't use compositing

quozl:
Because of the delay imposed by compositing, on particularly fast hardware, it also causes a breakthrough of home view between launcher and activity window mapping.

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -181,7 +181,7 @@ def _start_window_manager():
     settings = Gio.Settings.new('org.gnome.desktop.interface')
     settings.set_string('cursor-theme', 'sugar')
 
-    _metacity_process = subprocess.Popen(['metacity', '--no-force-fullscreen'])
+    _metacity_process = subprocess.Popen(['metacity', '--no-force-fullscreen', '--no-composite'])
 
     screen = Wnck.Screen.get_default()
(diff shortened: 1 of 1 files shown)


samdroid-apps:
> breakthrough of home view between launcher and activity window mapping.

What does this refer to?

Also, from the last discussion, I remember this resulted in a huge framerate drop when playing videos in WebKit2 browse.  Is that still true?

quozl:
> > breakthrough of home view between launcher and activity window mapping.
> What does this refer to?

The order of redraws is;
* the home view,
* the launcher,
* the home view (for one tenth to one half a second, depending on hardware),
* the activity window.

This presumably happens because the home view composition is already cached, and the system is busy rendering the composition for the activity window.  A redraw race.

The correct order of redraws is;
* the home view,
* the launcher,
* the activity window.

Do you know a way for the launcher to detect completion of window visibility?

Webkit2 Browse was tested and there was no huge framerate drop.  It is difficult to imagine a mechanism whereby removal of compositing should cause video playback to slow, if anything it should be the other way around.

samdroid-apps:
I tested it again and it revealed that the webkit2 video playback issues were not connected to compositing.  It was going at 0.5fps both with and without compositing.  Probably an issue with my setup (is `startx` deprecated?)

Anyway, I will merge this.  Video playback is still silky smooth via "jukebox".

PR #690, opened 2016-05-30
Remove log noise in Journal List View

quozl:
Was introduced in 265353af21f306e5b6316110c8b4d3e0a9e21b07 and does not seem to serve a purpose.


davelab6:
Looks good to me

PR #693, opened 2016-06-14
Password visibility toggle

iamutkarshtiwari:
This fixes the issue - https://bugs.sugarlabs.org/ticket/4963

Now the user can toggle the password visiblility via 'show password' checkButton provided

desktop-animation

### DIFF
--- a/src/jarabe/desktop/keydialog.py
+++ b/src/jarabe/desktop/keydialog.py
@@ -101,11 +101,19 @@ class KeyDialog(Gtk.Dialog):
         self.set_default_response(Gtk.ResponseType.OK)
 
     def add_key_entry(self):
-        self._entry = Gtk.Entry()
+        self._entry = Gtk.Entry(visibility=False)
         self._entry.connect('changed', self._update_response_sensitivity)
         self._entry.connect('activate', self._entry_activate_cb)
         self.vbox.pack_start(self._entry, True, True, 0)
         self.vbox.set_spacing(6)
+
+        self._show_pass_toggle = Gtk.CheckButton("Show Password")
+        self._show_pass_toggle.props.draw_indicator = True
+        self._show_pass_toggle.props.active = False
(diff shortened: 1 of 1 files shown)


src/jarabe/desktop/keydialog.py
```diff
@@ -102,10 +102,19 @@ def __init__(self, ssid, flags, wpa_flags, rsn_flags, dev_caps, response):
 
     def add_key_entry(self):
         self._entry = Gtk.Entry()
+        self._entry.set_visibility(False)
```
samdroid-apps: You can just pass properties as constructor kwargs, eg:

    self._entry  = Gtk.Entry(visibility=False)

src/jarabe/desktop/keydialog.py
```diff
@@ -120,6 +129,12 @@ def create_security(self):
     def get_response_object(self):
         return self._response
 
+    def _toggle_visibility_cb(self, entry):
```
samdroid-apps: Convention is double underscore for callbacks.  Can you fix that?

src/jarabe/desktop/keydialog.py
```diff
@@ -120,6 +129,12 @@ def create_security(self):
...
+        if self._show_pass_toggle.get_active():
```
samdroid-apps: you don't have to use an if statment here... `self._entry.set_visibility(self._show_pass_toggle.get_active())`

samdroid-apps:
Looks good.  I totally agree with the patch.

Maybe we could default to it being visible?  Visible is more usable :) In a school context, is the WiFi password really secret?  I also think that windows 10 defaults to visible wifi passwords, but don't trust me on that.

iamutkarshtiwari:
fixed.

samdroid-apps:
Great.  I'll merge it!  But can you just squash 'em commits?

iamutkarshtiwari:
Squashed.

quozl:
@iamutkarshtiwari, please review #703.

PR #695, opened 2016-07-03
Remove Dari and Guarani languages workaround

quozl:
Sugar should not misrepresent the system language list.

Will be added back as a packaging patch by OLPC OS if required for any future release.

References:
 * http://lists.sugarlabs.org/archive/sugar-devel/2016-July/053245.html by @icarito 
 * https://github.com/sugarlabs/sugar/pull/550#issuecomment-118431549


icarito:
Yes! Thank you. This hack needs to go as per #550 - providing a cleaner Sugar for Debian locale style.
+1 for merging.

davelab6:
+1

PR #698, opened 2016-07-10
Migrate to WebKit2

samdroid-apps:


samdroid-apps:
So as discussed on the sugar-toolkit-gtk3 thread, I have changed the patch to be a simple port to WebKit2.

@quozl, does this look like an easy to revert patch from the OLPC side?

quozl:
No, it looks unnecessarily invasive.  Further changes to the file will be expensive to backport.  You have instead opportunities to;

- hide the version change in the import, e.g. `from gi.repository import WebKit2 as WebKitn`,

- record the API version in some other way, e.g. a global,

- test the version in the other code changes, e.g. suppress use of SoupGNOME.

- split the commit into commits that can be done for both versions of WebKit, and commits that must be done for only one version, e.g. the move to `self._has_local_help`; (although it isn't clear why that was done, the variable is not used outside `__init__`).

API version independent coding is well worth the effort when both versions are expected to be in use.

samdroid-apps:
Thanks @quozl.

> API version independent coding is well worth the effort when both versions are expected to be in use.

It also adds a maintenance burden to the upstream.  As discussed on the sugar-toolkit-gtk3 pull request, the consensus was not to adopt this option.  Running WebKit1 is very risky in 2k16, and not something that SugarLabs should endorse.

I've updated patch as per your recommendations.  Please review.

PR #699, opened 2016-07-10
Resume Activity Gsettings

iamutkarshtiwari:
A new boolean value is now added to gsettings to toggle the default start mode of an activity from home view-
1) If the value is TRUE- Resumed instance opens
2) the value is FALSE - New instance starts

This GSetting value can be toggled from terminal by the following command - 
**$ gsettings set org.sugarlabs.user resume-activity false** (This sets the home view to start the new instance of every activity)

**$ gsettings set org.sugarlabs.user resume-activity true**  (This sets the home view to resume the instance by default)

### DIFF
--- a/data/org.sugarlabs.gschema.xml
+++ b/data/org.sugarlabs.gschema.xml
@@ -87,6 +87,13 @@
             <summary>Group Label</summary>
             <description>Label associated with age, e.g., '2nd Grade'</description>
         </key>
+        <key name="resume-activity" type="b">
+            <default>true</default>
+            <summary>Resume Activity</summary>
+            <description>If TRUE, Sugar will set default launch of activities in home view to 'resumable' mode. 
+                         If FALSE, Sugar will set default launch of activities in home view to 'start-new' mode.
+            </description>
+        </key>
         <child name="background" schema="org.sugarlabs.user.background" />
     </schema>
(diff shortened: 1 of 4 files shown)


iamutkarshtiwari:
@samdroid-apps @icarito Could you please test it and let us know whether is working fine at your end?

data/org.sugarlabs.gschema.xml
```diff
@@ -87,6 +87,11 @@
             <summary>Group Label</summary>
             <description>Label associated with age, e.g., '2nd Grade'</description>
         </key>
+        <key name="resume-activity" type="b">
+            <default>false</default>
```
samdroid-apps: Setting it to false means that "start new" is the default mode.  This is a change from normal sugar.

I'm fine with adding this setting, please set the default to "true" though.

If you want to have the default as "false" that is a design change.  That would need to be discussed.

data/org.sugarlabs.gschema.xml
```diff
@@ -87,6 +87,11 @@
...
+            <summary>Resume Activity</summary>
+            <description>If TRUE, Sugar will set default launch of activities in home view to resumable mode.</description>
```
samdroid-apps: Please document what happens if it is false.

iamutkarshtiwari:
@samdroid-apps Improved and rebased.

samdroid-apps:
Thanks.  Nice change.

PR #703, opened 2016-07-19
Wireless password fixes

quozl:
Sugar 0.109.0.1 and 0.109.0.2 cannot connect to new secured networks.  The wireless password prompt does not appear.  Saved passwords made with 0.109.0.0 and earlier continue to work.

This regression was introduced in #693 commit dca2f17b8dd16adcff0ec98dc0ee63f8c8dce38b, and was apparently not tested in the form it was committed.

Also in this pull request;
- ensure "Show Password" can be translated,
- default to show the password, as per consensus in #693 and the default in 0.108,
- simplify and improve code style,
- remove an unnecessary call to the Gtk.Widget.show() method.

See individual commits for full details.

### DIFF
--- a/src/jarabe/desktop/keydialog.py
+++ b/src/jarabe/desktop/keydialog.py
@@ -110,7 +110,6 @@ class KeyDialog(Gtk.Dialog):
         button.props.draw_indicator = True
         button.props.active = self._entry.get_visibility()
         button.connect("toggled", self._button_toggled_cb)
-        button.show()
         self.vbox.pack_start(button, True, True, 0)
 
         self.vbox.show_all()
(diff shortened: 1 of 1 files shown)


iamutkarshtiwari:
@quozl Looks good to me 👍

src/jarabe/desktop/keydialog.py
```diff
@@ -127,7 +127,7 @@ def create_security(self):
     def get_response_object(self):
         return self._response
 
-    def __toggle_visibility_cb(self, entry):
+    def _toggle_visibility_cb(self, entry):
```
samdroid-apps: It seems to be the convention to use a double underscore for callbacks.  Can you change the caller instead?

quozl:
It was consistently wrong in the file, so I used that style.  Additional patch pushed to change all the callbacks in the file.  Retested.  Plenty more where that came from!  :grinning:

PR #704, opened 2016-07-19
Wireless icon badge consistency

quozl:
The neighbourhood view uses icon badges to mark networks that are either locked or have been connected to.

The locked badge is lost when a network is connected to.  Fix is to add a new badge for locked and connected to.  Requires https://github.com/sugarlabs/sugar-artwork/pull/92.

The connected to badge is not removed promptly when the connections are discarded using the control panel.  Fix is to connect a removed signal and update the icon badge.

p.s. a patch for making discard testing faster by adding it to the buddy menu can be found at http://dev.laptop.org/~quozl/z/1bPOj7.txt

### DIFF
--- a/extensions/cpsection/network/model.py
+++ b/extensions/cpsection/network/model.py
@@ -17,7 +17,6 @@
 
 import logging
 
-import dbus
 from gettext import gettext as _
 from gi.repository import Gio
 from gi.repository import NMClient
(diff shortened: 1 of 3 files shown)


samdroid-apps:
Nice patch, I like the idea.  This might appear like bike-shedding, but I find the icon very hard to comprehend.  Could we space the icons out a little more?  Eg. stack the badges 1 on top of another?

I'd like any suggestions though.

quozl:
I find the locked icon hard to comprehend already.  It is too small.  Even on the XO's diagonal colour pixel arrangement, it is rare to comprehend it as a padlock unless you've been told what it is.  The Help activity resorts to pointing out the icon badge and saying "If the circle shows a lock symbol, expect to enter a key or password."  Take a look at this microscope image of the padlock on an XO-1:

shot0007-crop

Partially close your eyes to see the icon.  :grinning:

See [more microscope images of wireless badges](http://dev.laptop.org/~quozl/2016-07-26-sugar-704-badges/), just uploaded.  Somewhat worse results on another display than XO-1, despite the box of the badge being cleaner as a result of the conventional pixel arrangement.

Perhaps the locked icon should only be used for a network that is secure and not connected, and an unlocked padlock icon for when the network is secure and connected; to express "your device has unlocked the network for itself."  But are we expressing an attribute of the network or an attribute of the connection?

I agree a feature of multiple badges on an `Icon` would be nice, as it would let us use both a favourite star and an locked icon side by side, which would help to solve this comprehension issue.

_Don't let the perfect be the enemy of the good?_  What you point out are useful feature requests, but might not be necessary right now for this pull request.  Is merging the pull request a better situation than not merging it; compare Sugar misleading user about whether a network is secure vs an icon that you find hard to comprehend, but will be easily learned in the same way the locked icon is.

PR #705, opened 2016-07-19
Enhanced favourites: hidden bundles

quozl:
The OLPC library bundles and several activities appear in the activity
ring as favourites; they did not appear previously.

The OLPC library bundles have not changed for several years, and are
effectively a deprecated feature.  They are not included in our Ubuntu
builds, where we use the Wikipedia activity instead.  Adding them to the
list should have no impact on Ubuntu or Fedora SoaS.

ImageViewer, Jukebox, and Read activities were originally hidden.  Sugar
recently exposed them as a result of the enhanced favourites
patches (#4841).  Adding them to the list will hide them.

TamTamEdit, TamTamJam, and TamTamSynthLab were originally hidden.

### DIFF
--- a/data/activities.hidden
+++ b/data/activities.hidden
@@ -1,2 +1,17 @@
 org.laptop.Terminal
 org.laptop.Log
+org.laptop.ImageViewerActivity
+org.laptop.sugar.Jukebox
+org.laptop.sugar.ReadActivity
+org.laptop.TamTamEdit
+org.laptop.TamTamJam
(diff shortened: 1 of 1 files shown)


samdroid-apps:
Oh, there is a Jamendo activity?  That's pretty cool!

This is a great patch though, thanks!

quozl:
> Oh, there is a Jamendo activity? That's pretty cool!

No, it is a library content bundle, not an activity bundle.  Sugar treats the two in a similar way.

PR #707, opened 2016-07-22
Favourites view number unnecessary

quozl:
On the home view, the label of the ring icon is "Favorites view 1", even when there is only one view.  One view is the default.  There is no GUI for changing the number of views.

Regression introduced by c5b0905916519927c35396a857bfe983b4912ea0.

Suppress the suffix when only one view is available.

### DIFF
--- a/src/jarabe/model/desktop.py
+++ b/src/jarabe/model/desktop.py
@@ -71,7 +71,7 @@ class DesktopViewModel(GObject.GObject):
 
         homeviews = self._settings.get_value(_HOMEVIEWS_KEY).unpack()
         if not homeviews:
-            # there should always be atleast one homeview
+            # there will always be at least one homeview
             self._settings.reset(_HOMEVIEWS_KEY)
             homeviews = self._settings.get_value(_HOMEVIEWS_KEY).unpack()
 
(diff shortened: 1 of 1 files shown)


src/jarabe/model/desktop.py
```diff
@@ -81,8 +81,11 @@ def _ensure_view_icons(self, update=False):
         self._view_labels = []
         for view in homeviews:
             if 'view-label' not in view:
-                view['view-label'] = _('Favorites view %d' % (
-                    len(self._view_labels) + 1))
+                default = _('Favorites view %d')
+                if len(homeviews) > 1:
+                    view['view-label'] = default % (len(self._view_labels) + 1)
+                else:
+                    view['view-label'] = default.replace(' %d', '')
```
samdroid-apps: Can we just add another string here?  That looks really risky.

quozl:
Another string is another translation delay; I've no wish to experience such a delay, and I have a production run to push through.

The primary risk is where the translated string requires %d in order to be sensible to the reader.  One possible way is if a translation adds the semantics of "number" before the %d, e.g. "Favourites view number %d".  Cannot tell on my own.  Perhaps _ibo, ig, sw, tr, vi_?

- "Ọkacha mmasị ele %d" (_ibo, ig_),
- "Zinazoonwa sana %d" (_sw_),
- "Favoriler Görünümü %d" (_tr_),
- "Xem Ưa thích %d" (_vi_).

A secondary risk is where the translated string places %d at the start, but none do.  One translation _ca_ places %d midstring;

- "Vista %d dels preferits"

Many translations haven't been updated with the %d, predating c5b0905916519927c35396a857bfe983b4912ea0 last year, e.g. _aym, cy, kn, na, pa, quz, quy, sd, st, to, tvl, zu_.  Guess these are unmaintained.

One translation _nb_ loses the %d, which may be a mistake in the translation;

- "Favoritvisning"

I'll defer to @leonardcj on this.

samdroid-apps:
Ping @leonardcj

i5o:
Tested and works.

Merging

PR #719, opened 2016-10-08
Fix for journal batch operations

quozl:
In 0.108 and earlier, the edit toolbar is shown when journal items are selected.

In 0.110 the edit toolbar does not appear, so operations such as batch erase are not available.

Regression was probably introduced by 8be8b2a ('Bind project list view to auto refresh') and other patches since 0.108.

References:

https://wiki.sugarlabs.org/go/Features/Multi_selection (2013)


samdroid-apps:
Thanks for the fix!

PR #720, opened 2016-10-08
Journal batch erase: fix traceback

quozl:
When erasing all items in the journal using multi selection feature, "Your Journal is empty" is correctly displayed, but a traceback is in shell.log;

```
  File "/usr/lib/python2.7/site-packages/jarabe/journal/listview.py", line 515, in __model_ready_cb
    self._show_message(_(show_message_text),
UnboundLocalError: local variable 'show_message_text' referenced before assignment
```

Indentation was wrong; caused by 92f7af4.

Use of gettext was wrong; cf648ef and 92f7af4 added string constants that were not wrapped, instead a variable was wrapped.

@leonardcj, note internationalisation impact, because two new strings are added.

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -511,12 +511,12 @@ class BaseListView(Gtk.Bin):
                     else:
                         self._show_message(_('The device is empty'))
                 else:
-                    show_message_text = 'No matching entries'
+                    show_message_text = _('No matching entries')
                     if self.get_projects_view_active():
-                        show_message_text = 'No Projects'
+                        show_message_text = _('No Projects')
 
-                self._show_message(_(show_message_text),
-                                   show_clear_query=self._can_clear_query())
+                    self._show_message(show_message_text,
+                        show_clear_query=self._can_clear_query())
         else:
             self._clear_message()
 


quozl:
Yep, and there's at least one more to come; POTFILES.in was out of date.

quozl:
@leonardcj, what is a
> POT refresh in github

As far as I can see, the repository `sugar` contains `po/POTFILES.{in,skip}` an a `po/*.po` for existing translations.

leonardcj:
After the propsed string changes, we need to recalculate/recompute.refresh of the POT fileso it can be landed in Poolte.  In the current configuration, Pootle is not taking care of regenerating POT files.  We have a go-forward plan of upgrading to ver2.8 of Pootle which has VCS support (agan) and it may be possible to take that burden off of the repo owner, but for now, we need the POT file refreshed on the repo side when strings are changed. Does that make sense?

quozl:
Yes, that's how it is supposed to work for everything, but there's no POT file in the git repository (like there is with activities) so the process is a bit vague to me.

leonardcj:
I think Sugar i18n involves intltool instead of the usual gettext somehow. I'll do some further background research on past practices, all of which we are revisiting.

i5o:
Merged.
As far as I understand(?), we will have to regenerate the POT files once we push all requests / land the string freeze period, no?

quozl:
Refreshing po/POTFILES.in should be done whenever any new .py files are added that require translation.  That is, if they `import gettext`.  Then, on release, there's some step done to update the generated POT.

PR #725, opened 2016-10-24
Update en.po

ujdhesa:
Adding missing question mark to the string _Do you want to downgrade to version %s_


quozl:
Thanks for noticing this.

You've made the change to a translation file which is generated from source code.

Please instead make the change to https://github.com/sugarlabs/sugar/blob/master/src/jarabe/journal/misc.py#L322 and force push.  As a result, eventually en.po will change.

i5o:
Replaced by 
https://github.com/sugarlabs/sugar/pull/726

PR #727, opened 2016-10-25
Comma missing in two strings

ujdhesa:
Hello.

Compared to:

Error in automatic pm argument, use on/off.

it seems that there are two strings where a comma is missing.

TIA,
Besnik


walterbender:
Thanks for catching this. Two comments:
(1) I think a semi-colon would be better. Or a period (and start a new sentence).
(2) You need to make the change in the source code, not the PO file that is compiled from the source file. (The source file is indicated by the comment line above the msg string.)

ujdhesa:
Thank you, Walter!

PR #729, opened 2016-11-01
View Source - distribution specific paths

quozl:
View Source was unable to find the toolkit.

site-packages is a Fedora specific path.

dist-packages is the Debian and Ubuntu equivalent.


i5o:
Thanks @quozl 

Merged.

PR #733, opened 2016-11-09
Restart metacity if it fails

quozl:
Fixes https://bugs.sugarlabs.org/ticket/4958

Supersedes https://github.com/sugarlabs/sugar/pull/682


quozl:
Rebased to master.

i5o:
Will test following: https://bugs.sugarlabs.org/ticket/4958

PR #734, opened 2016-11-12
Fix inconsistent HomeBox imports

quozl:
FavoritesBox was referenced as favoritesview, yet ActivitiesList was directly imported into the namespace.

Simplifies reading of the code.


i5o:
Testing

i5o:
Works, merged.

PR #738, opened 2016-11-30
Revert "Refactor journal.objectChooser according to popup module in toolkit-gtk"

quozl:
This reverts a series of three commits;

* cadac88c1c880708bd2afe0e82a2c2356c4bd9df
* 0ba3171a3a5c92698a89553bdfa541c83e520694
* 739993b2c4e12faa33e6714a42e500c5d80fbb9f

On both Fedora 18 and Ubuntu 16.04, the series renders classroom laptops useless in minutes or hours, because any activity that opens the object chooser does fail to exit on quit, and processes leak and build-up on the system until memory is depleted. Evidence of the problem is shown by ps(1). Workaround is to reboot.

Reverting the series fixes the problem.

Handling the `Wnck.Screen` `window-close` signal in the same way as the original code does not fix the problem.

Commit cadac88 doesn't say why the change was made, and doesn't appear to add any functionality, so reverting seems the safest option.

### DIFF
--- a/src/jarabe/journal/objectchooser.py
+++ b/src/jarabe/journal/objectchooser.py
@@ -24,7 +24,6 @@ from gi.repository import Wnck
 from sugar3.graphics import style
 from sugar3.graphics.toolbutton import ToolButton
 from sugar3.graphics.objectchooser import FILTER_TYPE_MIME_BY_ACTIVITY
-from sugar3.graphics.popwindow import PopWindow
 
 from jarabe.journal.listview import BaseListView
 from jarabe.journal.listmodel import ListModel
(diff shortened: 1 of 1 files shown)


i5o:
Just saw that you have few pull requests waiting for testing, or smh.

I'll test ASAP and let you know :) (or just merge)

Thanks :D

quozl:
Thanks.  I'm guessing this problem doesn't yet affect SoaS (because the problem is in 0.110), but should reproduce within sugar-build.

i5o:
Tested in f25:
ps axfww (only sugar-build opened): https://paste.fedoraproject.org/496763/80745449

-> sugar build -> Open Jukebox -> Open Object Chooser -> Close the activity
https://paste.fedoraproject.org/496764/07454891

jukebox process is still there

i5o:
Merged because the bug is affecting ubuntu users. 

We have to look a fix to fedora25 or so

PR #739, opened 2016-12-03
favoritesview: enable all home view layouts

ortegaps:
This is a GCI task and requires to enable all the home views
that we already have coded.

### DIFF
--- a/src/jarabe/desktop/favoritesview.py
+++ b/src/jarabe/desktop/favoritesview.py
@@ -60,11 +60,11 @@ _logger = logging.getLogger('FavoritesView')
 _ICON_DND_TARGET = ('activity-icon', Gtk.TargetFlags.SAME_WIDGET, 0)
 
 LAYOUT_MAP = {favoriteslayout.RingLayout.key: favoriteslayout.RingLayout,
-              # favoriteslayout.BoxLayout.key: favoriteslayout.BoxLayout,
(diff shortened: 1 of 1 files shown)


quozl:
Tested.

* in square view, corner icons overlap; perhaps the square view code has not been adjusted for a change in icon size, you might compare other changes to the same file since the square view was disabled or added,

Reviewed.

* include reasons why the home views are being enabled, and why they were not enabled already,
* in your commit message don't undermine patch review by mentioning that your work is part of a GCI task; patch review is no different whether the work is part of GCI, a regular volunteer or a paid third-party contractor.

square

chimosky:
@quozl I tested on a wider screen, 3440x1440 and don't notice the icon overlap.
Screenshot from 2022-06-27 01-37-30
 
Could you confirm that the issue still persists?

quozl:
That's great.  My guess is something changed since my original test,
which was at about 1355 x 759 pixels.  An odd resolution.

Sorry, don't currently have time to test again.  If it's working now
with current release of Sugar and dependencies, then close this.

chimosky:
> Sorry, don't currently have time to test again.  If it's working now
with current release of Sugar and dependencies, then close this.

Seeing as the aim of the PR is still valid, I think we should close by merging.

chimosky:
@quozl if you have no objections I'll be merging this.

quozl:
Okay.

PR #740, opened 2016-12-05
Added Corner Layout

cvgarciarea:
GCI Task: https://codein.withgoogle.com/dashboard/task-instances/5119622560874496/

To know how the icons are arranged:
https://s14.postimg.org/zcr55soep/corner_layout.png

### DIFF
--- a/src/jarabe/desktop/favoriteslayout.py
+++ b/src/jarabe/desktop/favoriteslayout.py
@@ -631,3 +631,105 @@ class TriangleLayout(RingLayout):
         return RingLayout._calculate_position(self, radius, icon_size, index,
                                               children_count, width, height,
                                               sin=sin, cos=cos)
+
+
+class CornerLayout(ViewLayout):
+    """Lay out icons in the upper left corner."""
+
+    __gtype_name__ = 'CornerLayout'
+
+    icon_name = 'view-corner'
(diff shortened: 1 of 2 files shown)


quozl:
Tested an earlier draft.

Reviewed.  Comments;
 * `setup` does not offset the owner icon by `allocation.y`, result may be collision,
 * the activity icon is not excluded from child allocation area, result may be collision,
 * `CornerLayout.setup` is mostly duplicated from `ViewLayout.setup`, we would prefer calling method instead of copy and paste, as that would reduce the number of code lines to be maintained,
 * might use `self._grid` for collision detection instead of pixel coordinates,
 * might use `RandomLayout` as parent class, in similar fashion to how previous layouts inherit from `RingLayout`.

Pro-Panda:
@cristian99garcia, will you be updating the patch or may I do the same, Thanks

mk1299:
Hi ! Is there anything that still needs to be done?Both the links on the parent comment are dead.

pro-panda:
@mk1299, see [review comments](https://github.com/sugarlabs/sugar/pull/740#issuecomment-265047779) for what needs to be done in the patch

quozl:
Yeah, please everybody in future don't use links that expire.

PR #741, opened 2016-12-14
Load background images from journal

cristian99garcia:
GCI Task: https://codein.withgoogle.com/dashboard/task-instances/5616648155299840/

I used datastore.find for add images in the view from the journal, and when the user select a journal image, copy that to "/home/user/.sugar/background.[extension]".
Why? Because sugar remove all images journal directory when it's restarted, and aren't yet created when it is necessary to load them as a background, for that, I copy the selected image to the above mentioned address.

### DIFF
--- a/extensions/cpsection/background/view.py
+++ b/extensions/cpsection/background/view.py
@@ -16,13 +16,17 @@
 
 
 import os
+import shutil
 
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import GObject
 from gi.repository import GdkPixbuf
 
+from sugar3 import env
+from sugar3 import mime
 from sugar3.graphics import style
+from sugar3.datastore import datastore
 from sugar3.graphics.radiotoolbutton import RadioToolButton
 from jarabe.controlpanel.sectionview import SectionView
 
@@ -37,6 +41,7 @@ class Background(SectionView):
         self._model = model
         self._images_loaded = False
         self._append_to_store_sid = None
+        self._journal_images = []
 
         self.connect('realize', self.__realize_cb)
         self.connect('unrealize', self.__unrealize_cb)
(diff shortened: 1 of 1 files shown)


i5o:
Tested and seems good to me; anyone else is welcome to discuss about it

extensions/cpsection/background/view.py
```diff
@@ -16,13 +16,17 @@
...
 import os
+import shutil
 
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import GObject
 from gi.repository import GdkPixbuf
 
+from sugar3 import env
+from sugar3 import mime
```
AbrahmAB: sugar3.mime imported but not used.

AbrahmAB:
Tested. Looks good :)

quozl:
Reviewed;
- [ ] how is the file eventually deleted; some systems have very little disk space, and every file is important,
- [ ] why did you avoid model.py?  We have a [Model-View design pattern](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) yet your change is to View only,
- [ ] in the commit message, please explain what the problem was and how the patch fixes it, (in particular because without a Google account the GCI task link does not work, and besides it won't be useful for future regression analysis; only the commit message is preserved in the git repository, we do not preserve pull request comments),
- [ ] in the commit message, please explain how to test the patch; expected result vs previous result,

I'd like to see your commit message grow.  Use the first line as summary, then press enter twice before going into detail.

ohnx:
@quozl GCI task link:

> ### Load background from Journal in "Background" section of control panel
> Add an option in background control panel section to let the user load a background (image) from journal, showing all images from Journal in the list is another possible way to do it.
>
> Right now the only option is by copying the image to the Documents folder, and then setting it on the control panel.

quozl:
A deleted comment received by e-mail quoted from the GCI task;

> Load background from Journal in "Background" section of control panel
>
> Add an option in background control panel section to let the user load a background (image) from journal, showing all images from Journal in the list is another possible way to do it.
>
> Right now the only option is by copying the image to the Documents folder, and then setting it on the control panel.

This is a good basis for writing the commit message, though it would be conventional to describe the problem first, then the solution.  See [Making commits](https://developer.sugarlabs.org/contributing.md.html#making%20commits) in the guide to contributing.

PR #745, opened 2017-03-18
Removes continue button while erasing multiple journal entries. Fixes #4971

DarkPrince304:
According to this issue: https://bugs.sugarlabs.org/ticket/4971 when erasing multiple journal entries, clicking on "Continue" while the erase is in process stalls it.

This removes the "Continue" button while the process is being performed. Please do let me know if this is fine or if this causes any repercussions. Thanks :)

### DIFF
--- a/src/jarabe/journal/palettes.py
+++ b/src/jarabe/journal/palettes.py
@@ -622,16 +622,9 @@ class BatchOperator(GObject.GObject):
             # this is only in the case the operation already started
             # and the user want stop it.
             self._stop_batch_execution()
-        else:
-            GObject.idle_add(self._prepare_batch_execution)
-
-    def _prepare_batch_execution(self):
-        self._object_index = 0
-        # Next, proceed with the objects
-        self._operate_by_uid()
-
-    def _operate_by_uid(self):
-        GObject.idle_add(self._operate_by_uid_internal)
+        elif hasattr(self, '_object_index') == False:
+            self._object_index = 0
+            GObject.idle_add(self._operate_by_uid_internal)
 
     def _operate_by_uid_internal(self):
         # If there is still some uid left, proceed with the operation.
@@ -665,6 +658,7 @@ class BatchOperator(GObject.GObject):
         self._object_index = len(self._uid_list)
 
     def _finish_batch_execution(self):
+        del self._object_index
         self._journalactivity.unfreeze_ui()
         self._journalactivity.remove_alert(self._confirmation_alert)
         self._journalactivity.update_selected_items_ui()


quozl:
Thanks, looks good.

I've not tested yet, travelling, but does hiding the button cause a position change of the other button?  If so, changing it to insensitive may be better.

Note the PR links to 4972, but the correct bug in the commit is 4971.  Had me confused for a while.  :grin:

DarkPrince304:
Hi,

In terms of UI/UX I guess "Continue" should not be there while erasing is
in process to begin with. So yeah Continue button is not there in this fix
and the stop button takes it place.

DarkPrince304:
Here is a preview of what it looks like.

quozl:
I've tested.  Yes, as a result of `remove_button` (Continue) the other button (Stop) changes position to be where the cursor or finger is.  This is unfortunate; visually disruptive.

As I said before, it would be better to `set_sensitive` the button, but the toolkit does not expose the buttons of an Alert, so this would require a toolkit change as well.

Instead, why not let the Continue button be pressed, but ignore it?  In the `else` clause of `__confirmation_response_cb` add a test for batch execution in progress, and if so do not call `idle_add`.

DarkPrince304:
Does the GObject provide any in built method for checking if the batch execution is in progress? What approach for checking did you have in mind?

quozl:
Thanks.  Reviewed.  Please rebase to one patch and force push.

quozl:
Thanks.  Tested.  Added to OLPC branch.

quozl:
Your commit message does not follow our conventions, so I did change it during cherry-pick for OLPC branch.  You should edit the commit message.  See [Making commits](https://developer.sugarlabs.org/contributing.md.html#making%20commits).  Here is my interim commit message.

```
Journal - fix batch continue button stall

A click on continue button will stall a batch operation.

Reproducer: create a reasonable number of journal objects so that erase
can be interrupted, select one then all journal objects, click on erase
and a message will show, click on continue and objects will begin to be
erased, click on continue again and progress will stall.

Hiding the button causes the Stop button to move unnecessarily, making
it harder to select.  Setting the button insensitive is difficult
because the toolkit does not expose it.

Ignore the button click if a batch operation is in progress.

Fixes #4971.  https://bugs.sugarlabs.org/ticket/4971
```

DarkPrince304:
Done. Hope this is the correct way :)

quozl:
Yes, well done.

Next step is for a maintainer to merge your pull request.

I've tested your change in new operating system builds, and it works fine.

PR #750, opened 2017-04-24
GSettings for SaveAs functionality

iamutkarshtiwari:
A new boolean value is now added to gsettings to toggle the ``Save-As`` functionality on activity close-

1) If the value is TRUE- Save As popup appears
2) the value is FALSE - Default jobjects write to journal (without Save As)

This GSetting value can be toggled from terminal by the following command - 
`$ gsettings set org.sugarlabs.journal save-as false`(To enable Save-As)

`$ gsettings set org.sugarlabs.journal save-as true`  (To disable Save-As)

### DIFF
--- a/data/org.sugarlabs.gschema.xml
+++ b/data/org.sugarlabs.gschema.xml
@@ -114,6 +114,12 @@
             <summary>Mimetype registry</summary>
             <description>Consists of key-value pairs mimetypes and their corresponding activity</description>
         </key>
+        <key name="save-as" type="b">
+            <default>false</default>
+            <summary>Save-As Alert</summary>
+            <description>If TRUE, Sugar will show a Save-As alert on activity close.
+            </description>
+        </key>
     </schema>
     <schema id="org.sugarlabs.sound" path="/org/sugarlabs/sound/">
         <key name="volume" type="i">


tony37:
I think your items 1 and 2 may be reversed,

Tony

tony37:
Sorry, my mistake. I misread the text.

Tony

data/org.sugarlabs.gschema.xml
```diff
@@ -94,6 +94,13 @@
                          If FALSE, Sugar will set default launch of activities in home view to 'start-new' mode.
             </description>
         </key>
+        <key name="save-as" type="b">
+            <default>false</default>
+            <summary>Resume Activity</summary>
```
quozl: Copy and paste error; text not changed.

data/org.sugarlabs.gschema.xml
```diff
@@ -94,6 +94,13 @@
...
+            <description>If TRUE, on every activity close, a Save-As popup will appear.
```
quozl: Too verbose; there's no need to define what happens with both true and false, and it can be implied the option must be true before the behaviour will happen.

iamutkarshtiwari:
@quozl Please review -
- [ ] fix summary,
- [ ] fix description,
- [ ] reference the new feature page in the commit message.

quozl:
Thanks.

quozl:
Holding off on merge until https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/327 is ready.

quozl:
Rebased and merged as f8b310490c1e6f388acaf87705a573d7e60f39bf.

PR #751, opened 2017-04-25
GConf value for SaveAs functionality

iamutkarshtiwari:
A new boolean value is now added to `GConf` value to toggle the ``Save-As`` functionality on activity close-

1) If the value is TRUE- Save As popup appears
2) the value is FALSE - Default jobjects write to journal (without Save As)

This GSetting value can be toggled from terminal by the following command - 
`$ gconftool-2 --type Boolean --set /desktop/sugar/journal/save_as  true` (To enable Save-As)

`$ gconftool-2 --type Boolean --set /desktop/sugar/journal/save_as  true`  (To disable Save-As)

Feature page - https://wiki.sugarlabs.org/go/Features/Save-As

PR #752, opened 2017-04-26
Revert "Fixed Issue: #4940 Journal entries name can be renamed to blank"

quozl:
This reverts commit e05452a571ef9366b2e44d3c9c28d1b6182c4b50 as the current behaviour is inconsistent across Sugar.

An journal entry title does not need to be non-blank; the learner should be allowed to use a blank or empty title.  The title can always be changed again.

The default title is never blank.  The title does not need to be unique; it is not a file name.

Reference:
https://bugs.sugarlabs.org/ticket/4940

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -802,9 +802,7 @@ class ListView(BaseListView):
 
     def __cell_title_edited_cb(self, cell, path, new_text):
         iterator = self._model.get_iter(path)
-        if not new_text.isspace() and not new_text == '':
-            self._model[iterator][ListModel.COLUMN_TITLE] = new_text
-
+        self._model[iterator][ListModel.COLUMN_TITLE] = new_text
         self.emit('title-edit-finished')
 
     def __editing_canceled_cb(self, cell):


tony37:
The issue #4940 was well taken. The fix should have been applied.

Back in the day, a customer wanted to show off his hospital database. 
His query was for patients older that 100. He was amazed to see huge 
list. Turned out that date of birth had been left blank and so the epoch 
date zero was used, January 1, 1601. Not one person suggested that since 
that was the way SHAS worked, that was the way it should be. The IT 
folks put a verification check in within hours. It took a little longer 
to collect the correct dates of birth.

The title does need to be non-blank so the user looking at the Journal 
View can tell which object is which.  Why should a blank or empty title 
not be considered a mistake?

Tony

quozl:
How do you propose a blank or empty title be set when the learner desires to set it?  Why take agency from the learner?  This isn't a health system.

tony37:
The user also has the agency to:

cd /
rm -rf *

although there is inconsistency in that this works in some deployments 
not others.

In the current scenario, the user gives a non-blank title during save 
but is not prevented from going to the Journal View and erasing the 
title. About the same degree of difficulty as saving an object with a 
blank title and going to the Journal View to set it. However, my guess 
is that in most cases a blank title is a mistake not a wish.

Tony

quozl:
We're not here to force the learner into our mode of thinking; it might be a mistake, and if so let them fix it, or it might be intentional.

quozl:
Some Journal source code citations for empty title support in Sugar;
* https://github.com/sugarlabs/sugar/blob/master/src/jarabe/journal/model.py#L796
* https://github.com/sugarlabs/sugar/blob/master/src/jarabe/journal/palettes.py#L72
* https://github.com/sugarlabs/sugar/blob/master/src/jarabe/journal/palettes.py#L638
* https://github.com/sugarlabs/sugar/blob/master/src/jarabe/journal/volumestoolbar.py#L69
* https://github.com/sugarlabs/sugar/blob/master/src/jarabe/journal/volumestoolbar.py#L132

An interesting previous ticket;
* http://dev.laptop.org/ticket/10533

tony37:
That is apparently exactly what you are doing, forcing the learner to 
think like you do.

What about the learner who would like an alert when entering a blank 
title? Does that thinking count for anything?

My mode of thinking is that #327 is what it is and should be merged 
without any more ticky-tack. You have the setting so it will never be 
inflicted on you.
That should be enough.

Tony

quozl:
Well, we're hearing a lot from you, but I'm hoping other people will say what they think.  I'll hold off on #327 some more now that I can see there is not general agreement on the purpose of the feature; you saying it is to avoid blank or empty titles, and me saying it is to force reflection and choice of a title.

tony37:
The purpose of the feature has been described over and over again. It is 
not about files with blank titles. Utkarsh chose to alert the user who 
clicked on ok with a blank title. Perfectly reasonable programming decision.

The technical issues you raised were fixed by Utkarsh quickly and 
apparently to your satisfaction. Your refusal to merge the PR is now 
based sole on your personal prejudices. There is a setting so that the 
feature by default is not enabled, so it is really difficult to 
understand your concerns.

You do not agree with it. Walter said go ahead. Sebastian says go ahead. 
Utkarsh says go ahead. I say go ahead.

Tony

quozl:
I don't think they have appreciated the finer points that are discovered on careful review of the patch, and it is those finer points I've asked Utkarsh to respond to.  No need to list them again, as they are over on https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/327

tony37:
What 'finer points' do you feel are outstanding?

Here is the checklist -

  * Flake8 in alerts.py fixed
    Please note - There are already existing flake8 errors before this
    PR. I would suggest fixing them in a separate PR rather than
    including in this one.
  * Branch rebased to the latest commit.
  * Variable name use for gsettings is |'_save_as_enabled'| since
    another variable with |_save_as| name has already been used in lot
    of places. And there are other existing gsettings variables in Sugar
    which go by this naming style - |gsetting_name_enabled| so should
    not be a problem.
  * Commit message updated with reference page
  * Check settings everytime rather than save it at startup
  * Add new |GConf| value to sugar schema for Save-As
    PR - sugarlabs/sugar#751 <https://github.com/sugarlabs/sugar/pull/751>
  * Add save-as functionality to |sugar-toolkit|
    PR - sugarlabs/sugar-toolkit#4
    <https://github.com/sugarlabs/sugar-toolkit/pull/4>

Utkarsh shows that he has addressed all of the issues.

Tony

PR #754, opened 2017-05-16
New eo, hr translations from Pootle server

leonardcj:


leonardcj:
Updating eo and hr which have had recent changes.

Issue #770, opened 2017-09-13
Make gender selection optional


icarito:
Off by default. Let deployments who want it activate it.

jamessandy:
Yeah I think that  gender selection should be based on decision because some people may see it from a different view of racism

quozl:
Was asked for more detail about this issue; here's a checklist.

- [ ] review the feature implementation, both in [requirements](https://wiki.sugarlabs.org/go/Features/About_Me) and code,
- [ ] find if the feature is already optional, and if so change the default,
- [ ] if the feature is not yet optional, add GSettings schema, conditional code, and set the default off.

quozl:
https://github.com/sugarlabs/sugar/commit/70011a37b4d0c6ec0c5585355c76afd14a41a5b8

PR #777, opened 2018-01-07
Don't ask for gender on default

kipply:
This will disable asking for gender on startup. Selecting gender will still be available in the About Me section of the settings, however it won't be required. I also set the default age selection icons used (when a gender isn't set) to use the 'male' icons since the clothing there is more generic. 

Fixs #770

### DIFF
--- a/data/org.sugarlabs.gschema.xml
+++ b/data/org.sugarlabs.gschema.xml
@@ -77,6 +77,11 @@
             <summary>User Gender</summary>
             <description>Gender of the Sugar user, either male, female, or unassigned</description>
         </key>
+        <key name="default-gender" type="s">
+            <default>'disabled'</default>
+            <summary>Ask for Gender</summary>
+            <description>Whether or not gender should be asked for when starting sugar. By default, gender will be disabled, and can be set to be male or female on default. Setting it to an empty string is reccomended to enable it without a default gender. </description>
+        </key>
         <key name="birth-timestamp" type="i">
             <default>0</default>
             <summary>User Birth Timestamp</summary>
--- a/src/jarabe/intro/window.py
+++ b/src/jarabe/intro/window.py
@@ -262,19 +262,16 @@ class _AgePage(_Page):
 class _IntroBox(Gtk.VBox):
     done_signal = GObject.Signal('done', arg_types=([object]))
 
-    PAGE_NAME = 0
-    PAGE_COLOR = 1
-    PAGE_GENDER = 2
-    PAGE_AGE = 3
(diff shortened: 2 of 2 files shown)


quozl:
Thanks.  Reviewed.  Please look at the `default-nick` value handling.  Requesting changes;

- [ ] move the new GSettings key from `org.sugarlabs` to `org.sugarlabs.user`,
- [ ] rename the new GSettings key from `gender-onstart` to `default-gender`, to fit with the naming and use of existing `default-nick`, with default value _disabled_ (which will skip the gender page), and possible other values _male_, _female_, and empty string, (which will show the gender page with the default),
- [ ] split your one patch into at least three; the schema trailing whitespace changes, the default age selection icons, and the not asking for gender on startup; it will simplify future `git bisect`,
- [ ] your `_IntroBox` class changes are difficult to read; it seems a complicated way to suppress a page, and it may be better to have `back` and `next` iterate through a list of pages now instead of use page numbers; this class is very old (2008) and did not anticipate optional pages.

I did not yet test your code, but I did test logout and login after clearing `nick`;
```
gsettings set org.sugarlabs.user nick ''
```
So that will be my test case, combined with other GSettings values.

kipply:
Thanks for the review! My branch has been amended with the requests.

quozl:
Thanks.  Nearly there.

Tested 07ac0e1.  Fails if the GSettings schema is not updated.

```
(main.py:2759): GLib-GIO-ERROR **: Settings schema 'org.sugarlabs.user' does not contain a key named 'default-gender'
```

- [ ] please place the `default-gender` settings read inside an exception handler; because we sometimes have people use parts of commits and not others,

Does not show a default gender if `default-gender` is set to `female`.  Didn't test `male`.

- [ ] when `default-gender` is set to `male` or `female`, show that as the default in the gender intro page.

- [ ] expand the schema description for the key; it does not describe the values that can be used,

Merged two of three patches as correct and uncontentious, you may wish to rebase.

kipply:
Updated! I'd avoid a bare except, but it doesn't actually seem to throw an exception?

quozl:
You're right.  Sorry, I was wrong.  We don't have much code that handles missing GSettings schema update.  What we do have uses `Gio.Settings.list_keys()` instead of an exception handler.

I've changed my mind; feel free to not handle a missing schema update.

kipply:
No problem :) 

Updated, and it includes a check to see if `default-gender` is in the keys.

quozl:
Thanks, great work.  Reviewed and tested.

PR #667, opened 2016-03-13
Add accessibility keys to Journal #4891

iamutkarshtiwari:
This PR add this feature request -> https://bugs.sugarlabs.org/ticket/4891

Now the journal can be accessed via keyboard. Following functions are implemeted ->
1. Entries can be scrolled with `'Up'/'Down'` arrow keys.
2. Activity can be started/resumed by pressing `'Enter'` key.
3. `Detail View` can be opened with `'Right'` arrow key.
4. Switching b/w `search entry` and `tree view` can done with `'Tab'` key.
5. `Escape` key can be use to switch back to list from detail view.
6. Activity can be `renamed` by typing anything.

Watch the below gif closely(Mouse is not used). :dancers: 
desktop-animation2

PR #791, opened 2018-03-13
[Feature]: Add accessibility keys to Journal

rahul-bothra:
**Feature:** https://bugs.sugarlabs.org/ticket/4891

This is an update to the existing PR #667

**Updates done (requested in #667):**
 - [X] Pressing `Return` in detail view resumes the activity
 - [X] Use `ctrl + f` to focus on search
 - [X] Search does not lose focus after a timeout

**Other Updates:**
 - [X] Press `ctrl + F2` to rename an entry
 - [X] Pressing `enter` key in title editing in detail view removes the focus from it.
 - [X] 22382d6 (Unrelated to the PR - fixes errors shown in logs)

**Could not be fixed:**
 - [x] Rename an entry, press `enter`, arrow keys need to be pressed twice to start scrolling.
        This issue ocurrs only the first time we try to rename. Further renaming works as expected
 - [ ] `Ctrl + 1`, `Ctrl + 2` ... to switch between Journal, Documents and mounts:
      **Attempts:**
      - emitted `toggled` signal. The connected function is called but no change is observed. Logs didn't show anything unexpected
      - emitted `clicked` signal. Same output as with `toggled`
      - called `__volume_changed_cb` directly with appropriate checks and arguments. Same output as above

@quozl, Kindly review

### DIFF
--- a/src/jarabe/journal/expandedentry.py
+++ b/src/jarabe/journal/expandedentry.py
@@ -252,6 +252,7 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
         self._vbox = Gtk.VBox()
         self.add(self._vbox)
 
+        self.in_focus = False
         self._metadata = None
         self._update_title_sid = None
 
@@ -262,8 +263,11 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
                               style.DEFAULT_SPACING * 2)
         self._keep_sid = self._keep_icon.connect(
             'toggled', self._keep_icon_toggled_cb)
+        self._title.connect('activate', self._title_entered)
         self._title.connect(
             'focus-out-event', self._title_focus_out_event_cb)
+        self._title.connect(
+            'focus-in-event', self._focus_in_cb)
 
         if Gtk.Widget.get_default_direction() == Gtk.TextDirection.RTL:
             # Reverse header children.
@@ -475,12 +479,14 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
 
     def _create_description(self):
         widget = TextView()
+        widget.connect('focus-in-event', self._focus_in_cb)
         widget.connect('focus-out-event',
                        self._description_tags_focus_out_event_cb)
         return self._create_scrollable(widget, label=_('Description:')), widget
 
     def _create_tags(self):
         widget = TextView()
+        widget.connect('focus-in-event', self._focus_in_cb)
         widget.connect('focus-out-event',
                        self._description_tags_focus_out_event_cb)
         return self._create_scrollable(widget, label=_('Tags:')), widget
@@ -488,7 +494,20 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
     def _create_comments(self):
         widget = CommentsView()
         widget.connect('comments-changed', self._comments_changed_cb)
+        widget.connect('focus-in-event', self._focus_in_cb)
+        widget.connect('focus-out-event', self._focus_out_cb)
         return self._create_scrollable(widget, label=_('Comments:')), widget
+
+    def _focus_in_cb(self, widget, event):
+        self.in_focus = True
+
+    def _focus_out_cb(self, widget, event):
+        self.in_focus = False
+
+    def _title_entered(self, widget):
+        self._update_entry()
+        self._title.hide()
+        self._title.show()
 
     def _title_notify_text_cb(self, entry, pspec):
         if not self._update_title_sid:
@@ -507,6 +526,7 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
         self._write_entry()
 
     def _update_entry(self, needs_update=False):
+        self.in_focus = False
         if not model.is_editable(self._metadata):
             return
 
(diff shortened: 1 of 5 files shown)


Pro-Panda:
**Update on the PR:**
 - [X] Fix sound alert while returning from `detail` view to `main` view
 - [X] Fix the need to use arrow keys twice after renaming entries for the first time

Pro-Panda:
Re: **Update:**
 - The sound alert is not yet fixed. It still ocurrs and random instances. A better solution could be to give an option of no sound alert on horizontal scrolling in a `sugar3.widgets.treeview`.

Pro-Panda:
Also requesting others to review and suggest more improvements

Pro-Panda:
**Update:** Commit message updated to detect co-author @iamutkarshtiwari in 2670767 

@quozl, is this change approproate

src/jarabe/journal/expandedentry.py
```diff
@@ -251,7 +251,7 @@ def __init__(self, journalactivity):
         Gtk.EventBox.__init__(self)
         self._vbox = Gtk.VBox()
         self.add(self._vbox)
-
```
quozl: Deleted empty line did help with grouping of code lines.

src/jarabe/journal/expandedentry.py
```diff
@@ -475,21 +478,36 @@ def _create_scrollable(self, widget, label=None):
...
         return self._create_scrollable(widget, label=_('Tags:')), widget
 
     def _create_comments(self):
         widget = CommentsView()
         widget.connect('comments-changed', self._comments_changed_cb)
+        widget.connect('focus-in-event', self._focus_event_cb)
+        widget.connect('focus-out-event', self._focus_out_cb)
         return self._create_scrollable(widget, label=_('Comments:')), widget
 
+    def _focus_event_cb(self, widget, event):
```
quozl: Should method name be `_focus_in_cb`?

src/jarabe/journal/expandedentry.py
```diff
@@ -251,7 +251,7 @@ def __init__(self, journalactivity):
         Gtk.EventBox.__init__(self)
         self._vbox = Gtk.VBox()
         self.add(self._vbox)
-
+        self._has_focus = False
```
quozl: `self._has_focus` is set to True or False but is never used in this class.  It is used by another class, therefore it should be a public attribute of the class, which suggests `self.has_focus`, but that name has specific meaning in Gtk.Widget scope, so either the name should be stolen or another name used.

src/jarabe/journal/listview.py
```diff
@@ -743,6 +744,38 @@ def __init__(self, journalactivity, enable_multi_operations=False):
         column.pack_start(cell_detail, True)
         self.tree_view.append_column(column)
 
+    def __key_detect_cb(self, tree_view, event):
```
quozl: Method name should be `_key_press_event_cb` (occurs nine times) or `_key_press_cb` (occurs two times).

quozl:
Tested e429e95 on Ubuntu 16.04.

Curious behaviour; can you reproduce this;
* produce less than a full screen of journal entries, and switch to it with f5 key,
* position mouse cursor in whitespace below last entry,
* press down arrow key repeatedly until last entry is highlighted,
* briefly move the mouse cursor up into the last entry and then back to the whitespace below,
* press `escape` key.

Expected result: nothing.

Observed result: area below list is shaded grey.
screenshot_ubuntu-xenial_2018-03-15_15 08 48

Pro-Panda:
I am able to reproduce the effect. I remember seeing this effect in one of the activities as well, not sure which.

Another way to reproduce the same effect is:
 - Keep scrolling until the last activity using `down` arrow key.
 - Press `right` arrow key to switch to detail view
 - Press `left` arrow key to switch back to main view
 - The area gets highlighted

I will try to find the cause and a fix for this

Pro-Panda:
**Update:**
 The shading of the bottom empty area:
 - Fixed for `Escape` key-press
 - Ocurrs on going to the last entry using `down` key and pressing
    - `ctrl`+`f2` and then `escape`
    - `ctrl`+`f` and then `escape`
    - `right` and then `left`

I am unable to find a cause or a fix for this. The area gets shaded when the last entry is in focus and then loses focus.
Requesting help from developers on the same.

quozl:
> Requesting help from developers on the same.

Developers of GTK+ may be helpful.  Write a stand-alone program that reproduces the problem, then iteratively remove as much as possible from the program until the program is as small as possible yet still reproduces.  Then post a pointer to the program to a GTK+ mailing list or issue tracker and ask for help.

Pro-Panda:
@quozl, 
Thank you, doing this.

chimosky:
@quozl could you share the branch this work is being done on if you still have it? The branch has been deleted by the author.

jcameron-sso:
You can still do this;
```
git fetch origin pull/791/head
git checkout abc62125c
```
therefore the branch being deleted is of no consequence.

chimosky:
> You can still do this;
> 
> ```
> git fetch origin pull/791/head
> git checkout abc62125c
> ```
> 
> therefore the branch being deleted is of no consequence.

Thanks.

chimosky:
Moved to #959.

PR #792, opened 2018-03-26
jarabe, extensions - remove GConf compatibility

quozl:
GConf compatibility support in Sugar allowed activities to read or be notified of changes to age, gender, nick, user colour, collaboration server, backup server, or automatic power management.  Redundant code updated GConf values at the same time as updates to Gio.Settings values.

Remove GConf from Sugar;

- remove the autoconf rules,

- remove the Makefile rules,

- remove the schema conversion for upgrades from GTK+ 2 Sugar,

- remove the gsettings-migrated setting key,

- remove the GConf schema,

- remove redundant code for GConf value updates,

- remove migration from GConf to Gio.Settings.

Warning: may break some activities.  May cause default values to be used, traceback, or fail to start for any activity using the removed GConf schema to read the removed configuration values for; age, gender, nick, user colour, collaboration server, backup server, and automatic power management.  Fix GTK+ 2 activities by porting to GTK+ 3 and Gio.Settings.  Fix GTK+ 3 activities by porting from GConf to Gio.Settings.

Tested on Ubuntu 16.04.


tony37:
Another make-work project for developers. The best solution would be to 
return to the config file in /home/olpc/.sugar/defaults. That has no 
dependency on upstream and is therefore immune from these diktats.

Tony

quozl:
We could also avoid GTK+ and use something more stable like X or Qt, but such a port would be a huge make-work.  With the withdrawal of GTK+ 2 we lose the use of many abandoned activities.

PR #802, opened 2018-07-20
Fix harmless traceback of sugar-control-panel

### LINKED ISSUE #793: sugar-control-panel harmless traceback
Using `sugar-control-panel` command generates a harmless but emotionally disruptive traceback;

```
guest@guest:~$ sugar-control-panel -g timezone
/usr/lib/python2.7/dist-packages/jarabe/journal/model.py:33: PyGIWarning: SugarExt was imported without specifying a version first. Use gi.require_version('SugarExt', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import SugarExt
/usr/lib/python2.7/dist-packages/jarabe/frame/eventarea.py:19: PyGIWarning: Wnck was imported without specifying a version first. Use gi.require_version('Wnck', '3.0') before import to ensure that the right version gets loaded.
  from gi.repository import Wnck
ERROR:root:Exception while loading extension:
  File "/usr/lib/python2.7/dist-packages/jarabe/controlpanel/cmd.py", line 68, in load_modules
    globals(), locals(), ['model'])
  File "/usr/share/sugar/extensions/cpsection/frame/model.py", line 17, in <module>
    from jarabe.frame import get_view
  File "/usr/lib/python2.7/dist-packages/jarabe/frame/__init__.py", line 16, in <module>
    from jarabe.frame.frame import Frame
  File "/usr/lib/python2.7/dist-packages/jarabe/frame/frame.py", line 31, in <module>
    from jarabe.frame.friendstray import FriendsTray
  File "/usr/lib/python2.7/dist-packages/jarabe/frame/friendstray.py", line 20, in <module>
    from jarabe.view.buddymenu import BuddyMenu
  File "/usr/lib/python2.7/dist-packages/jarabe/view/buddymenu.py", line 37, in <module>
    import jarabe.desktop.homewindow
  File "/usr/lib/python2.7/dist-packages/jarabe/desktop/homewindow.py", line 28, in <module>
    from jarabe.desktop.meshbox import MeshBox
  File "/usr/lib/python2.7/dist-packages/jarabe/desktop/meshbox.py", line 35, in <module>
    from jarabe.view.buddyicon import BuddyIcon
  File "/usr/lib/python2.7/dist-packages/jarabe/view/buddyicon.py", line 19, in <module>
    from jarabe.view.buddymenu import BuddyMenu
ImportError: cannot import name BuddyMenu
/usr/share/sugar/extensions/cpsection/keyboard/model.py:19: PyGIWarning: Xkl was imported without specifying a version first. Use gi.require_version('Xkl', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import Xkl
  
guest@guest:~$ █
```

Return status is zero, which indicates success.  Actual output is the blank line before the shell prompt.

pro-panda:
Fixes #793

Tested on: Ubuntu 16.04, Sugar 0.112

### DIFF
--- a/bin/sugar-control-panel
+++ b/bin/sugar-control-panel
@@ -15,6 +15,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import sys
+import gi
+gi.require_version('Xkl', '1.0')
+gi.require_version('SugarExt', '1.0')
+gi.require_version('Wnck', '3.0')
+gi.require_version('Gtk', '3.0')
 
 from jarabe import config
 
--- a/src/jarabe/desktop/favoritesview.py
+++ b/src/jarabe/desktop/favoritesview.py
@@ -42,7 +42,6 @@ from jarabe.view.palettes import JournalPalette
 from jarabe.view.palettes import CurrentActivityPalette
 from jarabe.view.palettes import ActivityPalette
 from jarabe.view.buddyicon import BuddyIcon
-from jarabe.view.buddymenu import BuddyMenu
 from jarabe.model.buddy import get_owner_instance
 from jarabe.model import shell
 from jarabe.model import bundleregistry
@@ -677,6 +676,9 @@ class OwnerIcon(BuddyIcon):
                                               __enter_notify_event_cb)
 
     def create_palette(self):
+        # import BuddyMenu here to avoid recursive import
+        # See ticket https://github.com/sugarlabs/sugar/issues/793
+        from jarabe.view.buddymenu import BuddyMenu
         palette = BuddyMenu(get_owner_instance())
 
         settings = Gio.Settings('org.sugarlabs')
--- a/src/jarabe/view/buddyicon.py
+++ b/src/jarabe/view/buddyicon.py
@@ -16,7 +16,6 @@
 from sugar3.graphics import style
 from sugar3.graphics.icon import CanvasIcon
 
-from jarabe.view.buddymenu import BuddyMenu
 from jarabe.util.normalize import normalize_string
 
 
@@ -40,6 +39,9 @@ class BuddyIcon(CanvasIcon):
         self._update_color()
 
     def create_palette(self):
+        # import BuddyMenu here to avoid recursive import
+        # See ticket https://github.com/sugarlabs/sugar/issues/793
+        from jarabe.view.buddymenu import BuddyMenu
         palette = BuddyMenu(self._buddy)
         self.connect_to_palette_pop_events(palette)
         return palette


quozl:
Thanks.  Reviewed to d153654.  I don't like the solution, because it does not explain the problem.  A hack.  Mysteries should not be left unsolved, because they often come back later.  Why is `from jarabe.view.buddymenu import BuddyMenu` failing, yet the import of `BuddyIcon` succeeds?  Is it because of a recursive import?  `BuddyMenu` is unique in that it appears twice in the traceback.

Pro-Panda:
> I don't like the solution, because it does not explain the problem.

Thanks for pointing out. I have added comments and updated the commit message

> Is it because of a recursive import ?

Yes,

quozl:
Is there an alternative solution?  Net effect of a326c5f is to defer the import until the palette is created.  At what elapsed time in the user experience?  During shell startup before the activity ring is rendered, or when the icon is hovered over or clicked?  We have some slow systems that run Sugar, and adding delay in the latter case would be unwelcome.

pro-panda:
I don't have access to any slower system, is there an alternative on how I can test for delay in slower systems on my current system

quozl:
You can simulate a slower system by configuring a virtual machine hypervisor to restrict the CPU time available.  You can then add `logger.error` calls to critical points in the shell startup.  For the OLPC XO, the Python files are pre-compiled (to `.pyo` files), so you should run one startup test and discard the results before testing again.

However, you may find `logger.error` calls sufficiently accurate to figure out when your extra delay is incurred.

Remember also that `sugar-control-panel` is very rarely used, yet `favoritesview.py` and `buddyicon.py` are used heavily.

pro-panda:
While I am not getting time to test the delay, I can't see an alternative solution to this

> Remember also that sugar-control-panel is very rarely used, yet favoritesview.py and buddyicon.py are used heavily.

I agree. Should we keep this then ?

quozl:
Perhaps refactor to avoid the recursion?

Look at the traceback again, and consider each import from the top.

Look at `extensions/cpsection/frame/model.py`; it imports `get_view` from `src/jarabe/frame/__init__.py` only in order to reach `Frame.settings` in `frame.py`.  Nothing else.  If `Frame.settings` was in a separate source file instead of in `frame.py`, none of the other parts would be imported.  The import recursion might be avoided.

What alerted me to this was the PyGIWarning for Gtk and the other libraries.  `sugar-control-panel` doesn't have a GUI.  Then no need to add `require_version`.

The reason there was only one `Frame.settings` to begin with is that each call to `Gio.Settings` for the same schema id creates a redundant object, consuming numerous resources, and not delivering signals.  See a5358777bf5ffe40988aeaa0d4a69badb4b80634 for details.

chimosky:
@pro-panda what's the status of this, I hadn't been paying much attention until I recently tried using `sugar-control-panel`

pro-panda:
@chimosky, thanks for bringing my attention to this. I'm not sure what did you mean by "status".

Stale. Closing

PR #803, opened 2018-07-20
embed gwebsockets in jarabe instead of a separate package

Pro-Panda:
Since `python-gwebsockets` is not used by anyone else other than us, other than us, should it be safe to embed it in the code itself ?

Reference Links:
https://www.archlinux.org/packages/community/any/python2-gwebsockets/
http://fedora.portingdb.xyz/pkg/gwebsockets/

Tested on: Ubuntu 16.04, Sugar 0.112

### DIFF
--- a/src/jarabe/apisocket.py
+++ b/src/jarabe/apisocket.py
@@ -22,8 +22,8 @@ import dbus
 from gi.repository import GLib
 from gi.repository import Gtk
 from gi.repository import Gio
-from gwebsockets.server import Server
-from gwebsockets.server import Message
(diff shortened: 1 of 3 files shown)


quozl:
It has usefulness on its own.  We don't know it is not used by others.  It is in Pypi.  There are 202 code references to gwebsockets across all of GitHub.  Downstreams are likely to continue to package it.  Based on these observations, I don't think it should be embedded.

Pro-Panda:
Sorry, I had missed PyPI. 
Thanks.

Issue #769, opened 2017-07-03
Journal placed above Home after startup


quozl:
In previous versions, Sugar startup shows the home view, a busy cursor, asynchronously starts the journal, and then turns off the busy cursor leaving the home view available.

In Ubuntu 17.10 Artful, Sugar startup shows the home view, a busy cursor, but the journal window is created above the home window.

Related to https://github.com/sugarlabs/sugar-toolkit-gtk3/issues/368

satellitee:
also found in fedora soas f27 and f28 rawhide
https://bugzilla.redhat.com/show_bug.cgi?id=1519042

quozl:
Caused by upgrade of metacity from 3.24.0 to 3.25.2, git bisect shows first bad commit is [unminimize windows with initial IconicState if first known on MapRequest](https://github.com/GNOME/metacity/commit/6875256e63fbfd4500ddd27a7bf1aafc8d4c1ca3 ported from [mutter](https://gitlab.gnome.org/GNOME/mutter/commit/be5c2eb).

nullr0ute:
We should probably report it upstream with metacity to get details.

quozl:
Thanks, that would be great.  I could not figure out where to report it upstream.

aperezbios:
@quozl , https://gitlab.gnome.org/GNOME/metacity/issues

quozl:
@aperezbios, thanks.  Banner blindness.  https://gitlab.gnome.org/GNOME/metacity/issues/4

quozl:
Fixed in https://github.com/GNOME/metacity/commit/074af8f87ef89b13ff326fb5d04ee424bbfd4ced

PR #810, opened 2018-11-27
Journal - iconify again if Metacity misbehaves (RHBZ #1519042)

quozl:
Journal covers Home View on startup when Metacity is 3.25.1 or later, isolated to Metacity [6875256](https://github.com/GNOME/metacity/commit/6875256e63fbfd4500ddd27a7bf1aafc8d4c1ca3) ("unminimize windows with initial IconicState if first known on MapRequest").

Detect the window state event where ICONIFIED is cleared without any action by the user.  Repeat the iconify request.

Has side-effect of showing a Metacity iconify animation, but this is better than not fixing it.

Fixes https://github.com/sugarlabs/sugar/issues/769

Fixes https://github.com/sugarlabs/sugar-toolkit-gtk3/issues/368

Tested on Fedora 18 and Ubuntu 16.04; does not trigger.

Tested on Ubuntu 18.04; does trigger.

### DIFF
--- a/src/jarabe/journal/journalactivity.py
+++ b/src/jarabe/journal/journalactivity.py
@@ -226,6 +226,7 @@ class JournalActivity(JournalWindow):
 
         self._dbus_service = JournalActivityDBusService(self)
 
+        self._unwanted = True  # user has not asked to see Journal
         self.iconify()
 
         self._critical_space_alert = None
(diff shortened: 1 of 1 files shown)


quozl:
Underlying problem has been fixed in https://github.com/GNOME/metacity/commit/074af8f87ef89b13ff326fb5d04ee424bbfd4ced.

PR #805, opened 2018-08-11
Port to Python 3

pro-panda:
Fixes #787

PR #817, opened 2019-01-18
Port to Python3

keshavvinayak01:
Updated SimpleHttpServer function of python2 to support python3


quozl:
I'd prefer all Python 3 porting of Sugar's core components to be in one pull request at a time, and we already have https://github.com/sugarlabs/sugar/pull/805 open, so please base your work on that, resolving the conflicts that have arisen since the previous point of merge.

PR #826, opened 2019-03-20
[Feature] Add remove button for a Journal entry

Hrishi1999:
To remove an entry, you need to first select the entry with checkbox and then remove it, or you can open the detail view and delete it. This makes it simple. 
Similar functionality can be found in Gmail, you can select multiple items and delete, or delete a single item by clicking on the "Trash" button.
Added a remove button next to the detail button for each Journal entry item,
The feature is demonstrated below: 
Remove

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -812,11 +812,9 @@ class ListView(BaseListView):
         registry = bundleregistry.get_registry()
         metadata = model.get(row[ListModel.COLUMN_UID])
         bundle = misc.get_bundle(metadata)
-       
         if bundle is not None and registry.is_installed(bundle):
             registry.uninstall(bundle)
         model.delete(metadata['uid'])
-
 
     def __icon_clicked_cb(self, cell, path):
         row = self.tree_view.get_model()[path]
@@ -865,7 +863,7 @@ class CellRendererDetail(CellRendererIcon):
         self.props.fill_color = style.COLOR_BUTTON_GREY.get_svg()
         self.props.prelit_stroke_color = style.COLOR_TRANSPARENT.get_svg()
         self.props.prelit_fill_color = style.COLOR_BLACK.get_svg()
-        
+
 
 class CellRendererRemove(CellRendererIcon):
     __gtype_name__ = 'JournalCellRendererRemove'
@@ -920,4 +918,3 @@ class CellRendererBuddy(CellRendererIcon):
             self.props.xo_color = xo_color
 
     buddy = GObject.Property(type=object, setter=set_buddy)
-    
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -36,6 +36,7 @@ from jarabe.journal.palettes import ObjectPalette, BuddyPalette
 from jarabe.journal import model
 from jarabe.journal import misc
 from jarabe.journal import journalwindow
+from jarabe.model import bundleregistry
 
 
 UPDATE_INTERVAL = 300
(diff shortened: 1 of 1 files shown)


quozl:
Thanks.  I disagree with the feature as you have implemented it.

In your argument, you have missed the right-click, mouse-over or long-touch menu, which has an erase button.

You have made an erase without confirmation.  I disagree with that.

You have also added code to uninstall an activity if a journal object was a bundle file.  I disagree with that.  Perhaps you added this as part of a fix to https://github.com/sugarlabs/browse-activity/issues/81, but that issue was for deleting the bundle file without uninstalling the activity.  Also, you only added it for the delete using your new button, not the other delete methods, which is inconsistent.

The comparison with Gmail is improper.  Gmail has a deleted items folder, and the Journal does not.  And this is not mail, it is children's work.  You seem to have devalued journal objects.  Perhaps you should be enabling the Save-As feature instead, if you don't want so many things in a journal?  I'd like the Save-As feature selectable from My Settings, but that's a different issue.

Also, in future please ensure your commit messages contain all relevant information; your commit message was too brief, and your pull request contained information that could have been in the commit message.  See our [commit message checklist](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).

Also, the order of widget creation does not match the order of display, left to right.  You fixed this in the `append_column` calls, but the lines above remain out of order.

Hope that helps.

Hrishi1999:
Would keep these points in minds. But is there any scope of improvement? Should I discard it?

quozl:
I've given my opinion.  Let's see what other people say.  If you think not enough people have answered your user experience issue, remember that few people watch pull requests in repositories, and you may have more people answer you if you use another place to ask user experience questions.

tony37:
We need to keep the interface simple and stable to help teachers and 
students get familiar with it and confident in its use. The pop-up menu 
for the Journal is already incredibly complex. Currently 8 options with 
three having sub-menus. In this case, apply the rule:If it is not 
broken, don't fix it. Instead, look for ways to make Sugar easier - more 
'iconic' and discoverable.

As a user, I would rather see the switch to the Journal not change the 
current activity (F4). This can simplify the process of renaming a 
Journal entry while working in an activity (e.g. screenshot). The 
Journal has its own key (F5). If that doesn't work - the frame can be 
used as a workaround. Designers should keep in mind that for every Sugar 
instance on a laptop, there are from 10-100 on the XO.

In the Journal popup, eliminate the 'Start/Resume' menu option - let 
that be done as now by clicking on the icon.

Where multiple activities can handle the resume - keep the current 
scheme. However pick 'resume' or 'start' and not both as at present.

In the Journal the button to switch to the detail view is at the far 
right of the screen. It should be moved to the left (possibly swapping 
position with the box). Currently users don't even see that button 
because their focus is on the icon and the title to the left. If moved, 
it will need a more appropriate icon than the small arrow.

Note also - F11 to go to full screen is now masked. Thus the addition of 
the mysterious full-screen icon which sometimes appears in the view 
secondary toolbar and sometimes in the top right corner (where an 
edge-enabled frame can make it difficult to reach). Perhaps the button 
on the XO to the left of the Frame key could be re-purposed as 
full-screen. The location of f11 is difficult to find by counting 
because of the insertion of the Journal key as F5 makes it the twelfth 
button.

Today it is ironic to reflect that one of the original goals of Sugar 
was to break from the 'Desktop' metaphor and specifically to avoid menus 
like 'File'.
Now we have two levels of toolbar; fortunately still based on icons. 
However,  each icon has a popup that must be translated to hundreds of 
languages.
Just to add to the confusion, James Cameron has taken to calling Sugar a 
'desktop'. The educators in Rwanda have a much better solution: 
interface. The XO has two interfaces: Sugar and Gnome.

If this developer is looking for a worthy issue - 'my setting' is 
available. On an XO, wthout scrolling, it shows the big toe of the bare 
foot icon needed to switch to gnome. The power management button shows 
only its top border. A simple resize of the tiles could solve this 
problem - all of them should be sized to be fully visible on an XO.

The locale selection UI is badly broken. It suggests a menu of languages 
but in fact all changes must be made in the window for the current 
locale - choosing in that language. Since Arabic is early in the 
alphabet, students frequently 'discover' how to switch to Arabic. It is 
not nearly so clear to them how to return to English. Also, I would like 
to see the switch to gnome setting as one of the menu items on the XO 
(same level as 'My Setting'). Clicking there on the Barefoot icon should 
initiate the switch. Only a simple alert is needed for confirmation.  In 
Rwanda, learning how to switch interfaces is a part of the curriculum! 
So much for discovery.

Tony
.

quozl:
I'll delete responses that directly attack me by putting words in my mouth.  You had much to say of value, but I won't take attacks any more.

Hrishi1999:
###

Issue #827, opened 2019-03-25
Restore org.gnome.desktop.interface cursor-theme


quozl:
The Sugar cursor affects the GNOME desktop environment.

This is caused by https://github.com/sugarlabs/sugar/blob/master/src/jarabe/main.py#L205 where the GNOME desktop cursor theme is changed to the Sugar theme.

Workaround is to use `gsettings` to change _cursor-theme_ in _org.gnome.desktop.interface_ back to what it was.  Or you can use `dconf` to reset all settings, but that has many side-effects.

* [ ] read the value of _cursor-theme_ before changing it in the `_start_window_manager` function,
* [ ] keep the _Gio.Settings_ instance instead of deleting at end of the `_start_window_manager` function,
* [ ] restore the old value in the `_stop_window_manager function`.

This should fix the problem for future users.

kushagra98:
#828 Checked on my machine and works perfectly :smiley:

PR #828, opened 2019-03-25
updated_main.py

kushagra98:
When you log out from your Ubuntu and login with Sugar, GNOME desktop cursor theme is changed to the Sugar theme. Hence when you log back to Ubuntu your cursor theme is of Sugar. 
However now it will come back to the default theme which you have set earlier in your Ubuntu.

To change cursor : **sudo update-alternatives --config x-cursor-theme**

To see the changes : **/usr/share/icons/default** 

Hence getting back to default before stopping the window manager helps!! :)

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -213,6 +213,8 @@ def _start_window_manager():
 
 
 def _stop_window_manager():
+    settings = Gio.Settings.new('org.gnome.desktop.interface')
+    settings.set_string('cursor-theme', 'default')
     _metacity_process.terminate()
 
 


quozl:
Thanks.  Reviewed.  It does not restore to previous setting, so next problem report will be "When I log in to Sugar, my cursor theme changes in GNOME are reset to default".  See https://github.com/sugarlabs/sugar/issues/827 for suggested changes.

kushagra98:
No it does not resets it.  

Here is what happens when you change you cursor theme on GNOME : 

**sudo update-alternatives --config x-cursor-theme**

cursor

Here I have changed my cursor to Redglass. Changes in the machine due to this : 

- **/usr/share/icons/default/index.theme**

- **/etc/alternatives/x-cursor-theme**

Both the directories get changed, containing information about my current cursor theme (**inherits='Redglass'**).

However when you login to Sugar the only changes that take place is :

- **/etc/alternatives/x-cursor-theme**

Due to this the cursor theme of Sugar in GNOME is displayed. However changing back to default (which still contains the value of the current GNOME cursor, in my case Redglass) in the _stop_window_manager function helps and works fine.

I have already tried this on my machine by changing my Cursor theme into different ones and logging into Sugar and getting back to Ubuntu. It works totally fine and does not resets my cursor theme which I have set.

P.S. 

- Do not mistake 'Default' as the default cursor theme folder. It is just a folder which contains the information about the current cursor. 
- If you run **dconf** to reset all the things, the default folder contains the information about DMZ-Black which is also a folder under the icons.

quozl:
Thanks for the research.  I've tested 827e2a7 on Ubuntu 18.04, by changing the GNOME cursor to _whiteglass_ using this command;
```
gsettings set org.gnome.desktop.interface cursor-theme 'whiteglass'
```
then I logged out of GNOME session and logged in to Sugar session, then logged out of Sugar session and logged in to GNOME session.  The default cursor had returned.  So I do think Sugar should save the current value and restore it.

srevinsaju:
Oops, forgot to mention closes this PR, on PR #897 
@quozl , This PR should be closed by #897

PR #832, opened 2019-05-23
Replacing "guest"

Aniket21mathur:
A simple guest might be misleading in place of username and password.

### DIFF
--- a/docs/development-environment.md
+++ b/docs/development-environment.md
@@ -58,7 +58,7 @@ Or, install `xrdp` and `rdesktop` then log in to Sugar in a window;
     sudo apt install xrdp rdesktop
     sudo adduser guest
     echo sugar | sudo tee -a /home/guest/.xsession
-    rdesktop -g 1200x900 -u guest -p guest 127.0.0.1
+    rdesktop -g 1200x900 -u username -p password 127.0.0.1
 
 Once you have a local Sugar desktop, or a remote Sugar desktop, development of activities can begin.
 
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -40,4 +40,4 @@ Create a user for Sugar and set a default desktop environment:
 
 Start a session:
 
-    rdesktop -g 1200x900 -u guest -p password 0
+    rdesktop -g 1200x900 -u username -p password 0


quozl:
There's an `adduser guest`, and the password isn't set.

Aniket21mathur:
No, it is necessary to have a username and a password on making a new user. So these flags are necessary.

quozl:
Thanks for the suggestion.  On reading your answer, I don't quite understand, sorry.  The `adduser` has to be consistent with the `rdesktop` command, if the optional `-u` is used to `rdesktop`.

I'm worried I'm going around in circles.  There's a balance between full detail and enough detail for the average developer.  5a1e02a900ff508f641077746a4228909197ac31 was done earlier this year because someone tested by not reading documentation for Xsession, adduser and rdesktop.  It makes me _not_ want to document commands in brief, and choose either descriptions of the task to be done or a complete script.  If I'm going for a complete script, I'm inclined to suggest the documentation should be replaced by an installation script.  But that's not appropriate for the expert section of setting up a development environment.  Perhaps shell variables should be used instead;

```sh
    sudo apt install xrdp rdesktop
    USERNAME=guest
    PASSWORD=guest
    sudo adduser $USERNAME
    echo -e "$PASSWORD\n$PASSWORD" | \
        sudo passwd $USERNAME 2>/dev/null && echo password changed
    echo sugar | sudo tee -a /home/$USERNAME/.xsession
    rdesktop -g 1200x900 -u $USERNAME -p $PASSWORD 127.0.0.1
```

However, it looks very complicated, and now the reader has to be guided to;
* choose another username if _guest_ is not appropriate, or already in use,
* choose a better password, because if the system is exposed to untrusted users (e.g. via SSH or locally), guest is a common guess.

Aniket21mathur:
Thanks. What I wanted to say in my previous comment was that after running
`sudo adduser guest`, the process asks for a password which is a necessary field and cannot be left empty.
So to start an instance using rdesktop it is mandatory to use `-u` for username and `-p` for password.
So the answer to your question

> There's an `adduser guest`, and the password isn't set. Does it work without `-u` and `-p` anyway? If so, they could be removed.

is No. It will not work without `-u` and `-p` .

quozl:
Thanks.  I've tested on Ubuntu 19.04;
* `adduser` does prompt for password,
* `rdesktop` does work without `-u` and `-p`, a login window is shown,

Screenshot_udd_2019-05-29_09:10:20

* default user is the current user,

However, rdesktop has some difficulties, such as a broken cursor, an inability to put the mouse on the edge of the display to activate the frame, and on some systems a black screen.  There's also errors shown by rdesktop suggesting a protocol violation.  The packages have not been recently updated.

So I've removed the instructions on how to use rdesktop until we can see a better way.  c6620493f

Aniket21mathur:
Thanks.

PR #834, opened 2019-05-24
Adding code for enabling deb-src.

Aniket21mathur:
A simple code snippet will make it easy for the new developers to enable deb
sources, as compared to the most general approach of manually doing it with a
text editor, in which there are chances of missing a line or two.

### DIFF
--- a/docs/development-environment.md
+++ b/docs/development-environment.md
@@ -77,9 +77,14 @@ Install the build dependencies. There are many, and their package
 names vary by distribution. A first start is in the Debian or Fedora
 packaging files. From 0.113, add Six.
 
-On Debian or Ubuntu, ensure `deb-src` lines are enabled in `/etc/apt/sources.list`, and then;
+On Debian or Ubuntu, ensure `deb-src` lines are enabled in `/etc/apt/sources.list`,or simply do;
 
-    sudo apt update
+```
+sudo sed -Ei 's/^# deb-src /deb-src /' /etc/apt/sources.list
+sudo apt-get update
+```
+and then;
+
     for module in sugar{-datastore,-artwork,-toolkit,-toolkit-gtk3,}; do
         sudo apt build-dep $module
     done


quozl:
Thanks.  Reviewed.  Looks familiar, perhaps you got this from askubuntu.com [E: You must put some 'source' URIs in your sources.list](https://askubuntu.com/questions/496549/error-you-must-put-some-source-uris-in-your-sources-list)?

My comments;
* the `-E` option isn't needed,
* `apt` is easier to type and is more correct than `apt-get` in recent versions of Ubuntu, and is consistent with the other uses of `apt` in the file,
* some systems may not have `# deb-src` lines; although they are a default left behind by ubiquity after installation, the user could have removed them and substituted a site-specific replacement; which is what I've got.

[SourcesList](https://wiki.debian.org/SourcesList) on the Debian Wiki is a good explanation.

Is there a better way to enable source lines?  @nswarup14, did you hit this too?

nswarup14:
@quozl
Thanks.
You're correct. 
1) -E isn't needed.
2) apt is definitely easier, safer and the recommended method.
3) As a solution, we can perhaps write a bash script/python script that would

Aniket21mathur:
> Looks familiar, perhaps you got this from askubuntu.com [E: You must put some 'source' URIs in your sources.list](https://askubuntu.com/questions/496549/error-you-must-put-some-source-uris-in-your-sources-list)?

Yes, I got it from this.

@quozl ,Thanks. I agree with your comments and the fact that my changes would not be appropriate for all cases.


> 3\. As a solution, we can perhaps write a bash script/python script that would update/enable the #deb-src lines from /etc/apt/sources.list( assuming it hasn't been tampered with already, otherwise we'll have to check and the right packages).

@nswarup14 . Thanks. But I don't think writing a bash script just for enabling deb-src is a good idea.
I made these changes just to make things easier for new developers who even don't know what deb-src
are, I also agree that without my changes the doc is self-sufficient(with a bit of googling) to guide through the installation process. :-)

quozl:
Isn't this simpler?

```
sudo apt-add-repository --enable-source --update universe
```

It almost works for me, does it work for you?

nswarup14:
I don't think
http://dev.laptop.org/~quozl/.us/dists/bionic/main/binary-amd64/Packages will
be added to to the /etc/apt/sources.list, (which contains the
rsvg.so(python2 static bindings for rsvg lib)and hence may not work on some
systems ( especially the later versions of Ubuntu where the rsvg packages
were being unmaintained/deprecated).

Thanks

quozl:
It adds a line;
```
deb http://archive.ubuntu.com/ubuntu bionic universe
```
It's a Python script, `/usr/bin/add-apt-repository`, so I had a look, but it does pass the enable source boolean to another module.  So whatever reason it is that it doesn't work will likely be down in that module.  Oh well, it was a nice idea.

quozl:
I've tested.  Debian has `deb-src` lines enabled by default.  Having disabled `deb-src` lines is specific to Ubuntu.  Debian users may disable `deb-src` themselves.  I can see the point; as Debian uses pdiffs to fetch updates, but Ubuntu does not.

quozl:
7aea5d1fb pushed.

Aniket21mathur:
Thanks.

PR #847, opened 2019-07-29
Dash in signal names

quozl:
GLib accepts signal names with dash or underscore, and documents the
signals with a dash.

Use of GLib signal names in Sugar was mixed.

Change to using the dash.

No functional change.

Not tested.


quozl:
@chimosky, please review.

chimosky:
Reviewed, tested. Thanks.

PR #883, opened 2019-12-03
Update rpi-soas.md

pupp3tc0m:
I found that It says that wireless wasn't working, but I tested it on my RPi and It works. I'm using version 31.


quozl:
Thanks for testing.  Yes, the comment on the wireless probably got out of date.

PR #886, opened 2019-12-03
Readme.md updated to add links to dependencies.

srevinsaju:
Unwanted Dependencies was mentioned. Most of it automatically installed. Links to dependencies was added;

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ Building
 Sugar follows the [GNU Coding
 Standards](https://www.gnu.org/prep/standards/).
 
-Install all dependencies, especially `sugar-artwork`, `sugar-datastore`,
-and `sugar-toolkit-gtk3`.
+Install all dependencies, especially [`sugar-artwork`](https://github.com/sugarlabs/sugar-artwork), `sugar-datastore` (automatically installed) ,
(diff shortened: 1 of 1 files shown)


Aniket21mathur:
@srevinsaju `sugar-datastore` is not installed automatically if you build through the native method(manual installation) on the other hand all the dependencies are automatically installed if you install through the `sucrose` pacakge.

Thanks for the linking part. Please also do it for datastore.

srevinsaju:
@Aniket21mathur updated, but I have not been able to use the sugar DE yet. It stops working at 
```
make: *** No targets specified and no makefile found. Stop.
make: *** No rule to make target 'install'. Stop.

```

Aniket21mathur:
Please follow [this](https://github.com/sugarlabs/sugar/blob/master/docs/development-environment.md#native-sugar) method and let us know where you get stuck. Thanks

srevinsaju:
> Autogen, configure, make, and install modules for Python 3;
 ```
for module in sugar{-artwork,-toolkit-gtk3,-datastore,}; do
    cd $module
    ./autogen.sh --with-python3
    make
    sudo make install
    cd ..
done
```
Here, also see #890

PR #897, opened 2020-01-08
Restore GNOME cursor theme on sugar DE exit

### LINKED ISSUE #827: Restore org.gnome.desktop.interface cursor-theme

### LINKED ISSUE #872: Cursor changed automatically
Issue: When I installed sugar desktop environment, the cursor automatically changed on ubuntu even though I did not do it explicitly
The cursor is same as that of sugar DE.

WhatsApp Image 2019-09-04 at 9 10 28 AM

srevinsaju:
Fixes #872 & Fixes #827 
@walterbender @quozl  Please review

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -212,8 +212,7 @@ def _start_window_manager():
 
 def _stop_window_manager():
     settings = Gio.Settings.new('org.gnome.desktop.interface')
-    if _saved_cursor_theme:
-        settings.set_string('cursor-theme', _saved_cursor_theme)
+    settings.set_string('cursor-theme', _saved_cursor_theme)
     _metacity_process.terminate()
 
 
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -195,12 +195,9 @@ def _restart_window_manager():
     return False
 
 
-def _restore_gnome_cursor():
-
-
 def _start_window_manager():
-    settings = Gio.Settings.new('org.gnome.desktop.interface')
     global _saved_cursor_theme
+    settings = Gio.Settings.new('org.gnome.desktop.interface')
     _saved_cursor_theme = settings.get_string('cursor-theme')
 
     settings.set_string('cursor-theme', 'sugar')
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -87,8 +87,6 @@ from jarabe.model import brightness
 _metacity_process = None
 _window_manager_started = False
 _starting_desktop = False
-
-# Set cursor as None on init
 _saved_cursor_theme = None
 
 
@@ -217,7 +215,6 @@ def _start_window_manager():
 
 def _stop_window_manager():
     settings = Gio.Settings.new('org.gnome.desktop.interface')
-    # Restore the GNOME cursor
     if _saved_cursor_theme:
         settings.set_string('cursor-theme', _saved_cursor_theme)
     _metacity_process.terminate()
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -27,7 +27,6 @@ import os
 import sys
 import subprocess
 import shutil
-import json
 
 # Disable overlay scrolling before GTK is loaded
 os.environ['GTK_OVERLAY_SCROLLING'] = '0'
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -27,6 +27,7 @@ import os
 import sys
 import subprocess
 import shutil
+import json
 
 # Disable overlay scrolling before GTK is loaded
 os.environ['GTK_OVERLAY_SCROLLING'] = '0'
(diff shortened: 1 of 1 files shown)


quozl:
Thanks.

* [ ] store the old value in memory, not in a file; there's no need to store it in a file,
* [ ] don't instantiate the same Gio.Settings so many times, cache it instead.

srevinsaju:
@quozl  Fixed it,. Please review

quozl:
Thanks.  Reviewed.

* [ ] new global variable _settings_  is specific to one settings, yet others are used in the source, see all references to _Gio.Settings_, this can be confusing,
* [ ] why a dictionary for the cursor?  isn't a single object enough?
* [ ] what is _e_ in the warning, and why is a warning shown?
* [ ] the number of lines added seems excessive.

srevinsaju:
@quozl 

> * [ ]  new global variable _settings_  is specific to one settings, yet others are used in the source, 

see all references to _Gio.Settings_, this can be confusing,
Ok, I had to declare `Gio.Settings` for mouse cursor on `stop_window_manager`. 

> * [ ]  why a dictionary for the cursor?  isn't a single object enough?

I expected to extend the cursors to other DE's which uses GTK or other forks of GNOME like CInnamon, Budgie which  doesn't inherit from Gio.settings-customtheme but uses some other methodology. I don't use CInnamon or Budgie, so well, can't test it.

> * [ ]  what is _e_ in the warning, and why is a warning shown?

Removed it. Used to instantiate user with message, custom cursor theme could not be received from Gio

> * [ ]  the number of lines added seems excessive.

Thats because I created two new functions to handle the events. As mentioned above, I thought it would be useful to extend the existing functions.

srevinsaju:
@quozl made changes on the basis of `IRC` messages: https://github.com/sugarlabs/sugar/pull/897/commits/a358d333f434348d0d6a0235bf16c92b079af2f2

quozl:
Thanks.  New flake8 warnings added;

```log
src/jarabe/main.py:199:1: E302 expected 2 blank lines, found 1
src/jarabe/main.py:202:1: E112 expected an indented block
src/jarabe/main.py:202:3: E999 IndentationError: expected an indented block
src/jarabe/main.py:206:1: W293 blank line contains whitespace
```

* [ ] prefix the global variable with `_` like the others, sorry didn't notice that before,
* [ ] put the global declaration after the function declaration,
* [ ] don't really need those comments,
* [ ] still not sure what the point is of testing the variable; it should never be None, (if the GNOME schema is missing, Sugar won't start),

srevinsaju:
> * [x]  prefix the global variable with `_` like the others, sorry didn't notice that before,
> * [x]  put the global declaration after the function declaration,
> * [x]  don't really need those comments,
> * [ ]  still not sure what the point is of testing the variable; it should never be None, (if the GNOME schema is missing, Sugar won't start),
Yes, in the unlikely case of getting an empty string, it would be incorrect to assign the custom_cursor_theme as 'None', if it ever happens. It would be better to terminate sugar with sugar-cursor unchanged.

quozl:
> Yes, in the unlikely case of getting an empty string, it would be incorrect to assign the custom_cursor_theme as 'None', if it ever happens. It would be better to terminate sugar with sugar-cursor unchanged.

An empty string is fine.  If we get an empty string on entry, we should set an empty string on exit, not _default_.

srevinsaju:
@quozl, thanks for that info. fixed it on https://github.com/sugarlabs/sugar/pull/897/commits/6be7c69e8754be0b878411e56cfeddfe1a29e2a0

quozl:
Thanks.  Reviewed.  Commits added to keep style and minimise change.

quozl:
Probably the last remaining thing;

* [ ] per the issue, and my first review, _keep the Gio.Settings instance instead of deleting at end of the _start_window_manager function_, this is so that time is not wasted setting up and tearing down the Gio.Settings instance; faster logout or shutdown, important if the shutdown is being driven by depleted battery.

srevinsaju:
@quozl have fixed the final correction on https://github.com/sugarlabs/sugar/pull/897/commits/0b89bb2577c84001b5a029447ea6650b91440cd7 .

quozl:
Thanks.  Reviewed.  Tested on Ubuntu 19.04 with Sugar 0.116.

srevinsaju:
Thanks @quozl for merging it. I was away for a while :)

PR #906, opened 2020-02-10
Depend on empy python module

### LINKED ISSUE #882: Configure check for empy3 with 0.116 fails with Fedora 32
@quozl , back in August, you committed a fix for configure.ac (d47694d) which now fails under Fedora 32.

When running configure, the empy test fails, even when the python3-empy package is installed.

You can successfully complete a build by simply commenting out the offending AC_PATH_PROGS([EMPY], [empy empy3]) test condition, or, by using the following hack:

AC_PATH_PROGS([EMPY], [empy empy3], [/usr/lib/python3.8/site_packages/em.py])

...but we should probably be using the AX_PYTHON_MODULE M4 macro instead: https://www.gnu.org/software/autoconf-archive/ax_python_module.html 

This needs to be corrected in the configure.ac file for all python release tarballs.

### LINKED ISSUE #896: configure: error: empy not detected despite being installed
# empy is required
Although I have installed empy `pip install empy` and `pip3 install empy`.
When I tried running `autogen.sh`, empy is required error keeps on showing.
```bash
checking for empy... no
checking for empy3... no
configure: error: empy is required
```

srevinsaju:
Fixes #882 
Fixes #896
Fix makefile dependency on /usr/bin/empy for Linux distributions which do not package a empy binary executable.

In comparison to @aperezbios 's patch on #896, I have tried to call 
`python3 -m em` so that the
* the `em` wherever located should be called
* Fixes the dependency on Linux distros that doesn't have the `x` flag set on the python files.

I am not sure if what I have attempted to fix is with regards in the Coding guidelines. The semantics seems correct, I have tested it. Works satisfactorily. 
Please review @quozl @aperezbios @tchx84

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,10 +15,7 @@ AM_MAINTAINER_MODE
 PYTHON=python3
 AM_PATH_PYTHON
 
-AC_PATH_PROGS([EMPY], [empy empy3])
-if test -z "$EMPY"; then
-	AC_MSG_ERROR([empy is required])
-fi
+AS_IF([$PYTHON -m em >/dev/null 2>&1], AC_MSG_RESULT([empy found]), AC_MSG_ERROR([empy not found]))
 
 PKG_CHECK_MODULES(SHELL, gtk+-3.0)
 
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -1,12 +1,12 @@
 SUBDIRS = icons
 
 sugar-72.gtkrc: gtkrc.em
-	$(PYTHON) - -D scaling=\'72\' $(srcdir)/gtkrc.em > \
-		$(top_builddir)/data/sugar-72.gtkrc < $(EMPY)
+	$(PYTHON) -m em - -D scaling=\'72\' $(srcdir)/gtkrc.em > \
+		$(top_builddir)/data/sugar-72.gtkrc
 
 sugar-100.gtkrc: gtkrc.em
-	$(PYTHON) - -D scaling=\'100\' $(srcdir)/gtkrc.em > \
-		$(top_builddir)/data/sugar-100.gtkrc < $(EMPY)
+	$(PYTHON) -m em - -D scaling=\'100\' $(srcdir)/gtkrc.em > \
+		$(top_builddir)/data/sugar-100.gtkrc
 
 sugardir = $(pkgdatadir)/data
 sugar_DATA =			\


quozl:
Thanks.  Reviewed.

* [ ] EMPY_VERSION is not a version, but an exit status,
* [ ] the three lines of the test could be reduced to one by putting the running of the module inside the `AS_IF` expression,

Next steps;

* [ ] needs testing on Fedora,
* [ ] needs to be done also in sugar-artwork.

srevinsaju:
@quozl I attempted to fix that thing which you have told. Please review

quozl:
Thanks, but that's not quite what I meant.  The three lines of the test were these;

```bash
$PYTHON -m em -V
EMPY_VERSION=$?
AS_IF([test "${EMPY_VERSION}" = "0" ], 
```

And a quick search of `configure.ac` files on a few hundred projects shows it can also be done like this;

```bash
AS_IF([$PYTHON -m em], 
```

A good example is the Telepathy Salut `configure.ac` file which has this;

```bash
AS_IF([$PYTHON -c "import twisted.words.xish.domish, twisted.words.protocols.jabber, twisted.internet.reactor, avahi" >/dev/null 2>&1],
```

Note the output redirection to quash noise.

Another is from systemd;

```bash
AS_IF([$($PKG_CONFIG --exists bash-completion)],
```

Also, your change was incorrect, the `AC` is missing from `AC_MSG_RESULT`.

Please take the time to do it right, and test.  If you don't have the time, then don't do it.  :grin:

srevinsaju:
I apologize for the missing AC, I remember testing the build, but yea. 
I attempted to create this PP with particularly no knowledge about make build systems, I just wanted to learn how to do it, and the best way to learn is probably by doing.

I am not sure what `dev/null` in `AS_IF([$PYTHON -c "import twisted.words.xish.domish, twisted.words.protocols.jabber, twisted.internet.reactor, avahi" >/dev/null 2>&1],` does, but it looks cool, and works quite efficiently too.

Tested on Arch Linux, Sugar 0.116 latest source.

chimosky:
>I apologize for the missing AC, I remember testing the build, but yea.
I attempted to create this PR with particularly no knowledge about make build systems, I just wanted to learn how to do it, and the best way to learn is probably by doing.

You're doing a good job, I doubt @quozl meant any harm.

> I am not sure what dev/null in AS_IF([$PYTHON -c "import twisted.words.xish.domish, twisted.words.protocols.jabber, twisted.internet.reactor, avahi" >/dev/null 2>&1], does, but it looks cool, and works quite efficiently too.

`>/dev/nul 2>&1` redirects any output to `/dev/null` and any output redirected there is discarded, like @quozl said earlier it's use to "quash noise".
You can read about `/dev/null` [here](https://www.maketecheasier.com/dev-null-in-linux/).

quozl:
Thanks.

Just one final typo; missed space after comma and before `AC_MSG_RESULT`.  Has no effect other than making the code readable.

Next steps;

* [ ] needs testing on Fedora,
* [ ] needs to be done also in sugar-artwork.

Answering from the comments above;

When testing, you should test both paths of the conditional statement; i.e. with and without empy installed.  With the `AC` missing before, that path of the statement should not have worked properly, and so your test should have failed.

The `/dev/null` is a Linux kernel device which when opened and read immediately behaves like an empty file, and when opened and written discards everything.  You'll often see it used for that purpose.  In this scenario, it is enough for the configure script to say that empy is not found, and so the failure to import empy need not be reported.  It saves confusion on the part of the user, which we've certainly seen from our new developers, who will spend all their time looking at a part of an error and miss the actual cause. :grin:

The `2>&1` is a shell redirection binding; it connects the standard error stream to the standard output stream. Python writes tracebacks to the standard error stream.  With both streams redirected to `/dev/null` the traceback is intentionally lost.

(You may also be interested in `/dev/zero`, a Linux kernel device that when opened and read returns zero bytes without ever reporting end of file.)

srevinsaju:
yup, thanks for the info, this time I tested both cases. I will test on Fedora Rawhide tonight

srevinsaju:
Seems like I would not be able to test on Fedora (low on bandwidth)
@chimosky if you use Fedora, could you help me test this 🥺

chimosky:
Tested on fedora, @srevinsaju apply patch to `sugar-artwork` so I can test.

quozl:
@chimosky, thanks for testing.

Thanks also to @srevinsaju for [extending this to sugar-artwork](https://github.com/sugarlabs/sugar-artwork/pull/114).

Issue #909, opened 2020-03-06
Port to Python 3 - apisocket.py


quozl:
* [ ] complete the Port to Python 3 for apisocket.py and related files.

apisocket.py implements a communication channel between a Sugar Web activity and the Sugar shell and Journal.  It is an important part of Sugar.

apisocket.py was automatically ported to Python 3 using `2to3`.  Testing revealed regressions.  The regressions revealed the porting was not done by a human.  Rather than fix each regression, the port to Python 3 should be repeated using our [Python 3 Porting Guide](https://github.com/sugarlabs/sugar-docs/blob/master/src/python-porting-guide.md) and other guides.

Commits will need to explain the changes and tests so that we can reach consensus.

Reproducer: start and stop a web activity on Sugar 0.116 with gwebsockets 0.7

shell.log contains

```log
1583480569.218756 ERROR dbus.connection: Unable to set arguments (None,) according to signature 's': <class 'TypeError'>: Expected a string or unicode object
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 401, in _introspect_reply_handler
    self._introspect_execute_queue()
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 387, in _introspect_execute_queue
    proxy_method(*args, **keywords)
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 137, in __call__
    **keywords)
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 584, in call_async
    message.append(signature=signature, *args)
TypeError: Expected a string or unicode object
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/jarabe/apisocket.py", line 323, in _message_received_cb
    stream_id = ord(message.data[0])
TypeError: ord() expected string of length 1, but int found
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/jarabe/apisocket.py", line 323, in _message_received_cb
    stream_id = ord(message.data[0])
TypeError: ord() expected string of length 1, but int found
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 607, in msg_reply_handler
    *message.get_args_list()))
[...]

PR #911, opened 2020-03-14
Fixed Python3 TypeError

shaansubbaiah:
For TypeError: ord() expected string of length 1, but int found
In Python3 indexing a `bytes` object returns an integer, `ord()` is not required.
`stream_id = ord(message.data[0])` -> `stream_id = message.data[0]`

For TypeError: must be str, not bytes
`file_object.write(data[1:])` -> `file_object.write(data[1:].decode("utf-8"))`

### DIFF
--- a/src/jarabe/apisocket.py
+++ b/src/jarabe/apisocket.py
@@ -211,7 +211,7 @@ class DatastoreAPI(API):
             self._client.send_error(info["close_request"], error)
 
         def on_data(data):
-            file_object.write(data[1:])
+            file_object.write(data[1:].decode("utf-8"))
 
         def on_close(close_request):
             file_object.close()
@@ -320,7 +320,7 @@ class APIServer(object):
 
     def _message_received_cb(self, session, message, client):
         if message.message_type == Message.TYPE_BINARY:
-            stream_id = ord(message.data[0])
+            stream_id = message.data[0]
             stream_monitor = client.stream_monitors[stream_id]
             stream_monitor.on_data(message.data)
             return
--- a/src/jarabe/apisocket.py
+++ b/src/jarabe/apisocket.py
@@ -121,7 +121,7 @@ class DatastoreAPI(API):
         instance_path = os.path.join(activity_root, "instance")
 
         file_path = os.path.join(instance_path, "%i" % time.time())
-        file_object = open(file_path, "w")
+        file_object = open(file_path, 'wb')
 
         return file_path, file_object
 
@@ -152,7 +152,7 @@ class DatastoreAPI(API):
 
     def load(self, request):
         def get_filename_reply_handler(file_name):
-            file_object = open(file_name)
+            file_object = open(file_name, 'rb')
             info["file_object"] = file_object
 
             if "requested_size" in info:
@@ -168,7 +168,7 @@ class DatastoreAPI(API):
             self._client.send_error(request, error)
 
         def send_binary(data):
-            self._client.send_binary(chr(stream_id) + data)
+            self._client.send_binary(bytes([stream_id]) + data)
 
         def on_data(data):
             size = struct.unpack("ii", data)[1]
@@ -211,7 +211,7 @@ class DatastoreAPI(API):
             self._client.send_error(info["close_request"], error)
 
         def on_data(data):
-            file_object.write(data[1:].decode("utf-8"))
+            file_object.write(data[1:])
 
         def on_close(close_request):
             file_object.close()
--- a/src/jarabe/apisocket.py
+++ b/src/jarabe/apisocket.py
@@ -261,6 +261,10 @@ class APIClient(object):
         self._session.send_message(json.dumps(response))
 
     def send_error(self, request, error):
+        if(type(error) == dbus.exceptions.DBusException):
+            dbus_error = error.get_dbus_name() + " " + error.get_dbus_message()
+            error = dbus_error
+
         response = {"result": None,
                     "error": error,
                     "id": request["id"]}


srevinsaju:
Seems like I worked on a different fix
@shaansubbaiah What is your test plan you have followed

shaansubbaiah:
Hi Srevin, this was supposed to fix  #909, issue pops up when opening and closing a web activity; have tried it on Gears Activity and the error `TypeError: ord() expected string of length 1, but int found` seems to have gone.
Edit: On the moon activity
```
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 607, in msg_reply_handler
    *message.get_args_list()))
  File "/usr/lib/python3.7/dist-packages/jarabe/apisocket.py", line 211, in error_handler
    self._client.send_error(info["close_request"], error)
  File "/usr/lib/python3.7/dist-packages/jarabe/apisocket.py", line 270, in send_error
    self._session.send_message(json.dumps(response))
  File "/usr/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type DBusException is not JSON serializable
```
Issue seems to arise from encoder.py?

srevinsaju:
@shaansubbaiah Thanks, I will test it with Gears Activity and let you know soon

srevinsaju:
@shaansubbaiah tested, I couldn't reproduce the error mentioned by @quozl . Could you give me a detailed help on how to test / reproduce this error (without your PR)
Thanks a lot

shaansubbaiah:
Did you check if gwebsockets is at v0.7 (the live build I had came with 0.6) and [Enabled debugging](https://github.com/sugarlabs/sugar/blob/master/docs/development-environment.md#change-debugging-level)? Simply opening and closing the Gears Activity should reproduce the error; I haven't done anything else.

srevinsaju:
@shaansubbaiah I am testing it on the latest 0.116 Build on Manjaro Kyria 19.0
This is my logs
image
I have set the logger levels and gwebsockets is 0.7

shaansubbaiah:
Please check `shell.log` :). Even my Gears Activity log doesn't show any error.

srevinsaju:
Thanks , that helped, I will get back to your PR. I found it in `shell.log`. Thanks

srevinsaju:
Thanks @shaansubbaiah.
This PR worked in removing the `TypeError`.
I was able to remove the other logs. 
Verified the the entry and exit of data before and after changing the ord to be same.

quozl:
Thanks.    I accept that your change has stopped the error, but I don't yet fully understand why your change worked, and how to test the code that was changed,.  Your commit message on 95675f2 didn't help.

Please rewrite the commit message.  See [making commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits) and [guide for reviewers](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#guide-for-reviewers).  In particular some or all of the narrative you placed in the pull request opening comment should be in the commit message, and the commit message should try to explain what the problem was, how it was fixed, and how you tested that the code works as it should.

quozl:
Fixes https://github.com/sugarlabs/sugar/issues/909.

shaansubbaiah:
Sorry for the sloppy commit message, I'll update it shortly. I've always thought there was a limit to the length of the commit messages, turns out the 50 char length is only for the first line.

quozl:
I see how you could have come to that conclusion.  The information out there is somewhat variable.  I don't have a problem with line widths of up to 76 characters, so that I can use "`git log`" in a standard terminal.  Unnecessary words should be removed from the first line, and the detail added in lines following the blank line.  Also check the style used in the repository you are working on.

srevinsaju:
@quozl How did the integer come instead of the character? I still cannot understand what actually happened between the lines / commits

shaansubbaiah:
@quozl I have updated the commit message. (Also, for some reason, editing the commit message using a GUI text editor like Kate didn't work. I had to use Nano CLI editor and make the changes)

@srevinsaju I believe that is due to a difference in how python2 and python3 index byte buffers.

quozl:
Thanks.  Much nicer commit message, and I understand better why you made the changes you did.

For the change to DatastoreAPI.load.on_data, your fix is to decode the bytes into utf-8, but the only reason you gave is that it stops the error.  The error happened when the file is created and written to.  I don't see a corresponding change to when the file is read.  Are you sure the file is read and sent back to the client intact?  How did you test that?  Might it be better to change the file open mode (in the two places) to binary?

For the change to _message_received_cb, again it looks like you're focusing on fixing the error but not on the underlying problem; that the code was not properly ported from Python 2 and tested.  Our [Python 3 Porting Guide](https://github.com/sugarlabs/sugar-docs/blob/master/src/python-porting-guide.md) mentions the need to "_Check for use of binary data, especially in files, pipes, and subprocesses,_", and as it looks very much like this was not done, perhaps it should be done now.

This problem looks deeper than fixing a couple of lines. It goes to the heart of handing the bytes data exchanged with the activity over the web socket.

Another test is whether you can resume a web activity from the journal and have the previously saved data properly loaded, even if it is binary.  You may have to look for a web activity that does that.  I can't remember which ones do.

quozl:
> @quozl How did the integer come instead of the character? I still cannot understand what actually happened between the lines / commits

Sorry, I couldn't figure out what you are referring to.

It might help to look at the changes to this file between 0.112 and 0.116;

```
git diff v0.112..v0.116 -- src/jarabe/apisocket.py
```

No sign of changes to the file mode, so not really ported yet?

srevinsaju:
Thanks @quozl and @shaansubbaiah 
The newer commit message explained everything perfectly to me. I went away to test it on python2 and python3, quite astounded by the results. It looks cool!

quozl:
Let us know if there's something we can fix in the porting guide.

shaansubbaiah:
@quozl As mentioned earlier, with and without the changes in the PR I get this error in `shell.log` . Could this also be related to the Porting from Python2 to 3 ? Seems to be originating from `encoder.py`, but I'm not exactly sure.
```
1584360319.744499 DEBUG gwebsockets: Received binary message, length 16
1584360319.744759 DEBUG gwebsockets: Got data, length 68
1584360319.744881 DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":14,"params":[1],"jsonrpc":"2.0"}'
1584360319.976344 DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 13}
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 607, in msg_reply_handler
    *message.get_args_list()))
  File "/usr/lib/python3.7/dist-packages/jarabe/apisocket.py", line 211, in error_handler
    self._client.send_error(info["close_request"], error)
  File "/usr/lib/python3.7/dist-packages/jarabe/apisocket.py", line 270, in send_error
    self._session.send_message(json.dumps(response))
  File "/usr/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type DBusException is not JSON serializable
1584360319.980641 DEBUG gwebsockets: Got data, length 69
1584360319.980873 DEBUG gwebsockets: Received text message b'{"method":"activity.close","id":15,"params":[],"jsonrpc":"2.0"}'
1584360319.981730 DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 15}
1584360319.992085 DEBUG gwebsockets: Got data, length 0

```

quozl:
> `TypeError: Object of type DBusException is not JSON serializable`

Watch this; in Python 2;

```log
host:~$ python2
Python 2.7.17 (default, Nov  7 2019, 10:07:09) 
[GCC 7.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import dbus
>>> x = dbus.DBusException()
>>> import json
>>> json.dumps(x)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/json/__init__.py", line 244, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: DBusException() is not JSON serializable
>>> █
```

and in Python 3 ...

```log
>>> json.dumps(x)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'DBusException' is not JSON serializable
>>> █
```

From this it would seem that the problem is that apisocket.py is attempting to convert an exception to JSON, which is impossible in both Python 2 and Python 3.  You would need to find out why it is trying to do that.

Looking back through the trace, DatastoreAPI.save.error_handler at your line 211 is the trigger.

Converting that into a narrative; when the datastore object is saved, a file is created, a stream monitor for close is set up, and when the close occurs the datastore is asked to update, which specifies an error handler.  The update fails though we don't know why.  (perhaps the _uid_or _metadata_ is invalid).  The error handler is called and tries to send the exception back to the client.

It is possible that this error handling code path was not tested by @dnarvaez.

You could debug further by finding out what the exception is.  It might be that the exception is also caused by our failure to port the module to Python 3.

quozl:
Overall, a more correct approach is to port the module to Python 3, by careful analysis of every data flow in light of the changes required when porting any code from Python 2 to Python 3.  It is a process of staring at the code for hours if necessary, reverse engineering, documenting for yourself, maybe running a few tests to prove your theories, and then coming up with changes that should have happened in aa18879e9717dfe2d30f249549e9a43d6dd6da4f.

shaansubbaiah:
@quozl could you please explain your workflow in coding/debugging Sugar and its components?
I have followed your advice earlier, SSHing to the VM and editing the files through VSCode on my main computer speeds up things quite a bit. One issue I keep facing is that for every change I have to log out and log in to Sugar again. Is there a workaround or a more efficient way to go about it?

quozl:
Sure.

Only have to restart for changes to Sugar.  Changes to Toolkit may need a restart if you are testing Sugar.  Otherwise changes to Toolkit are picked up by an activity when it starts.

I've two methods I use to restart under IDE control.  Both use a `Makefile`.  One method not used for a while is to drive the VM over SSH or via `xdotool` to do something.  Second method is preferred; add code that will detect a restart request via filesystem and call `sys.exit` or `os.execv` or similar.

https://github.com/sugarlabs/record-activity/commit/2703fc2fc0803f056faedd69fb95d82cc055b63d is an example on the Record activity port to GTK 3 branch.  A `Makefile` does a `touch go` on the VM, and this repeating task in the activity detects it and calls the activity again.

But take care; sometimes the tooling you do can undo your testing.  In this case on the Record activity, on certain laptops (not VMs) the camera could not be opened after being closed so rapidly, or the camera wasn't actually closed before the `os.execv` took effect.  It depended on the kernel driver, which was unique for the camera type.  I had to add an explicit close of GStreamer to finish tooling on that laptop.  Always remember to do a final test without any automation tooling, and try not to leave the tooling behind in the pull request. You can use `git stash` or branches to help with that.

srevinsaju:
@shaansubbaiah, maybe this might benefit you too. In my case where I couldn't use a VM, I used to test sugar by having a native Linux installation (say, Ubuntu, Fedora or Arch Linux) and then running two desktop environments at the same time.

I too like you, used vscode and vim on the native KDE Plasma Desktop environment and then used switch user to create another TTY on TTY2. Launch sugar on TTY2 by selecting in drop down tab.
Make changes in your vscode, save it locally, or do `python3 setup.py dev` for testing purposes to create a symlink on your test files to the `~/Activities` directory.
Then use `Ctrl + Alt + F1` and Ctrl + Alt + F2` to switch between desktop environments realtime

quozl:
Yes, that works too, although collaboration testing does need separate IP addresses and this is hard to do with separate virtual consoles.  You can also use technologies like xvfb, RDP, or VNC in order to avoid VMs.

shaansubbaiah:
Thanks, I'll try them out.

shaansubbaiah:
Just set up the Live Build v117 from the Mailing List, working on porting.

shaansubbaiah:
@quozl is there any documentation for the apisocket file?

Also, I should be able to view the data sent at each step by using the logger?

quozl:
I don't know of any documentation.  Make sure you check gwebsockets, the mailing list archives at the time of the first commits, and our Wiki.  The original author vanished, we never found out what happened to them.

Yes, data can certainly be logged.  If you are not seeing logs, make sure you flush stderr, because it is buffered in Python 3.

Also, it is critical to bring up a Python 2 system for comparison debugging, that way we can be more certain of the changes.  I suggest OLPC OS 18.04 as a working Python 2 release of Sugar that can be edited to add debugging.  Create a new branch from the version tag to demonstrate any debugging added.

shaansubbaiah:
@quozl , printing the error I get:

```
1585224824.453091 DEBUG PY3-VM: ERROR:
1585224824.453414 DEBUG PY3-VM: org.freedesktop.DBus.Python.ValueError: Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/dbus/service.py", line 707, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/usr/lib/python3.7/dist-packages/carquinyol/datastore.py", line 373, in update
    lambda * args: self._update_completion_cb(async_cb,
  File "/usr/lib/python3.7/dist-packages/carquinyol/filestore.py", line 49, in store
    raise ValueError('No file at %r' % file_path)
ValueError: No file at dbus.String('/home/shaan/.sugar/default/com.garycmartin.Moon/instance/1585224824')

1585224824.453572 DEBUG PY3-VM: INFO CLOSE REQUEST:
1585224824.453698 DEBUG PY3-VM: {'method': 'close_stream', 'id': 14, 'params': [1], 'jsonrpc': '2.0'}
```
It is due to apisocket.py not saving the file for the Moon Activity at _/home/shaan/.sugar/default/com.garycmartin.Moon/instance/1585224824_ . This does not happen with the gears activity (perhaps they store files differently, I have to check it out.)

The _../instance/{timestamp}/_ directory remains empty when I try to list files before, after and while the Moon Activity is open. I have used `inotifywatch` to check if files are being created and got this output: (inotifywatch is checking the _~/.sugar/default/com.garycmartin.Moon/_  directory recursively for 10 seconds during which I open and close the activity once)

```
shaan@debian:~/.sugar/default/com.garycmartin.Moon$ inotifywatch -v -t 10 -r /home/shaan/.sugar/default/com.garycmartin.Moon/
Establishing watches...
Setting up watch(es) on /home/shaan/.sugar/default/com.garycmartin.Moon/
OK, /home/shaan/.sugar/default/com.garycmartin.Moon/ is now being watched.
Total of 4 watches.
Finished establishing watches, now collecting statistics.
Will listen for events for 10 seconds.
total  access  modify  close_write  close_nowrite  open  moved_from  create  filename
37     13      0       0            12             12    0           0       /home/shaan/.sugar/default/com.garycmartin.Moon/
13     2       3       2            1              3     1           1       /home/shaan/.sugar/default/com.garycmartin.Moon/instance/
```
This shows that it has created a file? I'm not exactly sure what the issue is, I'll be going through the [Datastore Documentation](http://wiki.sugarlabs.org/go/Development_Team/Almanac/sugar.datastore.datastore) and inspecting the Moon Activity.

Full log [here](https://drive.google.com/file/d/182qWWQdH2d8RxW5mNe5_Bcy40T9zHWxx/view?usp=sharing). (Gdrive link, fpaste expires after a day, wasn't sure when you would have time to go through this comment.) Also is all the data in the ~/.sugar/default/datastore and ~/.sugar/default/data is stored in plaintext? It appears so. 

Sorry it has been awhile, had some college assignments and projects preponed due to the current situation with the virus.

quozl:
Thanks for the update.

shaansubbaiah:
@quozl As mentioned earlier, the issue had occurred only with the Moon activity and not with the Gears activity. My understanding based on what I have read in the DataStore documentation and going through the logs is that 

1. `Stream is created -> Data loaded/ stored -> Stream closed`.
2. The `instance` folder at ~/.sugar/default/{activity name} is used as some sort of temporary storage before it is moved to the DataStore directory at ~/.sugar/default/datastore.

I have observed that the Moon Activity opens the Stream twice. 
`Open 1st stream -> Open 2nd stream -> Save Data -> (instance is emptied) Close 1st stream -> Close 2nd stream (Generates error as the data at the instance folder is missing). `

Log here:
```
1585664098.542321 DEBUG gwebsockets: Received text message b'{"method":"open_stream","id":11,"params":[],"jsonrpc":"2.0"}'
1585664098.542874 DEBUG gwebsockets: Sending text message {"result": [0], "error": null, "id": 11}
1585664098.543675 DEBUG gwebsockets: Got data, length 66
1585664098.544173 DEBUG gwebsockets: Received text message b'{"method":"open_stream","id":12,"params":[],"jsonrpc":"2.0"}'
1585664098.544683 DEBUG gwebsockets: Sending text message {"result": [1], "error": null, "id": 12}
1585664098.545433 DEBUG gwebsockets: Got data, length 415
1585664098.546014 DEBUG gwebsockets: Received text message b'{"method":"datastore.save","id":13,"params":["7750f837-db5b-409d-8d25-c9024a8d497d",{"timestamp":1585151101,"title":"Shaan Moon Activity","uid":"7750f837-db5b-409d-8d25-c9024a8d497d","title_set_by_user":"1","creation_time":1585151101,"activity":"com.garycmartin.Moon","checksum":"1441a7909c087dbbe7ce59881b9df8b9","filesize":"15","activity_id":"fe11c941b3b2337c9d2f395d82fe06364b30d62c"},0],"jsonrpc":"2.0"}'
1585664098.547274 DEBUG PY3-VM: DATA STORE API save()
1585664098.547662 DEBUG PY3-VM: uid
1585664098.548018 DEBUG PY3-VM: 7750f837-db5b-409d-8d25-c9024a8d497d
1585664098.548545 DEBUG PY3-VM:  metadata:
1585664098.548766 DEBUG PY3-VM: {'timestamp': 1585151101, 'title': 'Shaan Moon Activity', 'uid': '7750f837-db5b-409d-8d25-c9024a8d497d', 'title_set_by_user': '1', 'creation_time': 1585151101, 'activity': 'com.garycmartin.Moon', 'checksum': '1441a7909c087dbbe7ce59881b9df8b9', 'filesize': '15', 'activity_id': 'fe11c941b3b2337c9d2f395d82fe06364b30d62c'}
1585664098.549025 DEBUG PY3-VM:  stream_id:
1585664098.549250 DEBUG PY3-VM: 0
1585664098.556527 DEBUG PY3-VM: OPENED A BINARY FILE @
1585664098.556987 DEBUG PY3-VM: /home/shaan/.sugar/default/com.garycmartin.Moon/instance/1585664098
1585664098.557275 DEBUG PY3-VM: <class 'str'>
1585664098.557628 DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 13}
1585664098.558614 DEBUG gwebsockets: Got data, length 415
1585664098.559231 DEBUG gwebsockets: Received text message b'{"method":"datastore.save","id":14,"params":["7750f837-db5b-409d-8d25-c9024a8d497d",{"timestamp":1585151101,"title":"Shaan Moon Activity","uid":"7750f837-db5b-409d-8d25-c9024a8d497d","title_set_by_user":"1","creation_time":1585151101,"activity":"com.garycmartin.Moon","checksum":"1441a7909c087dbbe7ce59881b9df8b9","filesize":"15","activity_id":"fe11c941b3b2337c9d2f395d82fe06364b30d62c"},1],"jsonrpc":"2.0"}'
1585664098.560868 DEBUG PY3-VM: DATA STORE API save()
1585664098.561305 DEBUG PY3-VM: uid
1585664098.562127 DEBUG PY3-VM: 7750f837-db5b-409d-8d25-c9024a8d497d
1585664098.562791 DEBUG PY3-VM:  metadata:
1585664098.563410 DEBUG PY3-VM: {'timestamp': 1585151101, 'title': 'Shaan Moon Activity', 'uid': '7750f837-db5b-409d-8d25-c9024a8d497d', 'title_set_by_user': '1', 'creation_time': 1585151101, 'activity': 'com.garycmartin.Moon', 'checksum': '1441a7909c087dbbe7ce59881b9df8b9', 'filesize': '15', 'activity_id': 'fe11c941b3b2337c9d2f395d82fe06364b30d62c'}
1585664098.563725 DEBUG PY3-VM:  stream_id:
1585664098.564133 DEBUG PY3-VM: 1
1585664098.565575 DEBUG PY3-VM: OPENED A BINARY FILE @
1585664098.565937 DEBUG PY3-VM: /home/shaan/.sugar/default/com.garycmartin.Moon/instance/1585664098
1585664098.566167 DEBUG PY3-VM: <class 'str'>
1585664098.566454 DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 14}
1585664098.568335 DEBUG gwebsockets: Got data, length 90
1585664098.569035 DEBUG gwebsockets: Received binary message, length 16
1585664098.569811 DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":15,"params":[0],"jsonrpc":"2.0"}'
1585664098.571831 DEBUG gwebsockets: Got data, length 90
1585664098.572204 DEBUG gwebsockets: Received binary message, length 16
1585664098.572531 DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":16,"params":[1],"jsonrpc":"2.0"}'
1585664098.857101 DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 15}
1585664098.857396 DEBUG PY3-VM: ERROOR:
1585664098.857453 DEBUG PY3-VM: org.freedesktop.DBus.Python.ValueError: Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/dbus/service.py", line 707, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/usr/lib/python3.7/dist-packages/carquinyol/datastore.py", line 373, in update
    lambda * args: self._update_completion_cb(async_cb,
  File "/usr/lib/python3.7/dist-packages/carquinyol/filestore.py", line 49, in store
    raise ValueError('No file at %r' % file_path)
ValueError: No file at dbus.String('/home/shaan/.sugar/default/com.garycmartin.Moon/instance/1585664098')
```
The files in datastore are saved in plaintext so we don't have to open them as binary files. The previous fixes: Removing the ord()  and  Converting the data to String using .decode('UTF-8') seem to work normally. I am able to resume the gears activity. I had renamed the moon activity and even that wasn't lost when I resumed it. 

I shall try a few more web activities and report back.

quozl:
Thanks for the update.  Earlier I said you must [find out what happens on Python 2](https://github.com/sugarlabs/sugar/pull/911#issuecomment-601826671).  Please do try that and compare.  This is because we have a reasonable expectation that any problem in this code was introduced by incomplete port to Python 3.  It will also give you a baseline to compare against.

shaansubbaiah:
I had tried that earlier, the VM kept freezing and Python 2 kept crashing without any error message. I was prompted to update some packages in Ubuntu while setting up SSH, not sure if that had anything to do with it. I will install it again and test.

quozl:
Okay.  I don't remember what Ubuntu release you have chosen to get a Python 2 environment, but [ubuntu.md](https://github.com/sugarlabs/sugar/blob/master/docs/ubuntu.md) shows what works and what doesn't.  i.e. use 19.04, or 18.04.

OLPC OS 18.04.0 [here](http://wiki.laptop.org/go/18.04.0) is my curated derivative with a later Sugar but still Python 2.

shaansubbaiah:
I had installed the OLPC OS 18.04. Also when I login to Sugar, the display resolution scales to 4K and it becomes really hard to operate. Attempting to resize scales UI from 4K to 1080p making everything 4x smaller. I have not had this issue with the other VMs installed. I have started installing the OS again, I'll avoid the update this time and see if the issue appears again.

quozl:
Ah, yes, that's possible.  OLPC OS 18.04 is intended for laptops with a resolution of about 1366x768 with an external monitor option in clone mode, and contains a patch to `/usr/bin/sugar` that facilitates that.  In turn because Sugar doesn't have multiple display support.

To fix, you might try editing the file to remove these lines;

```bash
# OLPC NL3 external monitor and power button support
if [ -e /sys/devices/virtual/dmi/id/bios_version ]; then
    gsettings set org.gnome.settings-daemon.plugins.media-keys active false
    gsettings set org.gnome.settings-daemon.plugins.xrandr default-monitors-setup clone
    unity-settings-daemon &
fi
```

shaansubbaiah:
@quozl after a fresh install of the OLPC 18.04 OS, installing only openssh-client and enabling the debugger, I get the following errors:

Opening and closing the Moon Activity

Initially 
```
1585833574.988766 ERROR dbus.connection: Unable to set arguments (None,) according to signature u's': <type 'exceptions.TypeError'>: Expected a string or unicode object
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 401, in _introspect_reply_handler
    self._introspect_execute_queue()
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 387, in _introspect_execute_queue
    proxy_method(*args, **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 137, in __call__
    **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 584, in call_async
    message.append(signature=signature, *args)
TypeError: Expected a string or unicode object
```

and then

```
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 607, in msg_reply_handler
    *message.get_args_list()))
  File "/usr/lib/python2.7/dist-packages/jarabe/apisocket.py", line 210, in error_handler
    self._client.send_error(info["close_request"], error)
  File "/usr/lib/python2.7/dist-packages/jarabe/apisocket.py", line 267, in send_error
    self._session.send_message(json.dumps(response))
  File "/usr/lib/python2.7/json/__init__.py", line 244, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: DBusException(dbus.String(u'Traceback (most recent call last):\n  File "/usr/lib/python2.7/dist-packages/dbus/service.py", line 707, in _message_cb\n    retval = candidate_method(self, *args, **keywords)\n  File "/usr/lib/python2.7/dist-packages/carquinyol/datastore.py", line 375, in update\n    lambda * args: self._update_completion_cb(async_cb,\n  File "/usr/lib/python2.7/dist-packages/carquinyol/filestore.py", line 47, in store\n    raise ValueError(\'No file at %r\' % file_path)\nValueError: No file at dbus.String(u\'/home/ss/.sugar/default/com.garycmartin.Moon/instance/1585833595\')\n'),) is not JSON serializable
```

Opening and closing the Gears activity

```
TypeError: DBusException(dbus.String(u'Traceback (most recent call last):\n  File "/usr/lib/python2.7/dist-packages/dbus/service.py", line 707, in _message_cb\n    retval = candidate_method(self, *args, **keywords)\n  File "/usr/lib/python2.7/dist-packages/carquinyol/datastore.py", line 375, in update\n    lambda * args: self._update_completion_cb(async_cb,\n  File "/usr/lib/python2.7/dist-packages/carquinyol/filestore.py", line 47, in store\n    raise ValueError(\'No file at %r\' % file_path)\nValueError: No file at dbus.String(u\'/home/ss/.sugar/default/com.garycmartin.Moon/instance/1585833595\')\n'),) is not JSON serializable
1585833672.466127 ERROR dbus.connection: Unable to set arguments (None,) according to signature u's': <type 'exceptions.TypeError'>: Expected a string or unicode object
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 401, in _introspect_reply_handler
    self._introspect_execute_queue()
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 387, in _introspect_execute_queue
    proxy_method(*args, **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 137, in __call__
    **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 584, in call_async
    message.append(signature=signature, *args)
TypeError: Expected a string or unicode object
```

This issue was present before the port to Python3 ?

chimosky:
> This issue was present before the port to Python3 ?

Yes, `OLPC 18.04` runs python2 activities and not python3.

shaansubbaiah:
> Yes, `OLPC 18.04` runs python2 activities and not python3.
> I'm guessing you're running both at master, you can look at running the gtk2 branch of moon activity or checking out the change before the port to python3 and do same for gears activity.

Thanks! You were right, they both ran Moon v19 from Master. I moved `/usr/share/sugar/activities/Moon.activity` to another directory and ran `setup.py dev` for the `gtk` branch. The issue does not seem to occur.

Is there a list of the activities ported from Python2 to 3? 
The issue seems to occur on the Find-words-activity also ([logs similar to the one at the Live Build v117](https://github.com/sugarlabs/sugar/pull/911#issuecomment-606668667)). This is built with JS.

chimosky:
> Is there a list of the activities ported from Python2 to 3?

Search `sugar-activity` to see python2 activities and `sugar3-activity` for python 3 activities.
You can view any python 3 activity commit history to see the commit for the port, you can also see this [project](https://github.com/orgs/sugarlabs/projects/1) that was created to track porting to python 3. 

> The issue seems to occur on the Find-words-activity also
Is it the same test environment?
Can you share the logs too?

quozl:
The problem that was being discussed is how a [JavaScript activity that writes to the journal twice doesn't succeed in doing so](https://github.com/sugarlabs/sugar/pull/911#issuecomment-606668667).  So I suggested testing to see if [this is a Python 3 porting regression](https://github.com/sugarlabs/sugar/pull/911#issuecomment-607057922).

Moon activity _master_ branch is JavaScript, not Python 3.  So I don't agree with @chimosky's most recent comments.  :grin:

quozl:
> @quozl after a fresh install of the OLPC 18.04 OS, installing only openssh-client and enabling the debugger, I get the following errors:
> [...]
> This issue was present before the port to Python3 ?

Thanks for testing.  I agree.  The traceback shows a functional similarity, though not a precise similarity.  Good to know.  Knowing that it isn't an effect of porting means the Python 2 release of Sugar can't be used as a well-functioning standard on which to look for data changes, and instead the problem will have to be fixed by deep analysis of the data flows between the JavaScript activity and Sugar.

You might also check that Moon activity is correctly using the API, by analysis of the source code and tracing execution or tcpdump data packets.

shaansubbaiah:
> Is it the same test environment?
> Can you share the logs too?

@chimosky Same environment, OLPC 18.04. Logs are similar to the previous one.

---


@quozl It appears that the activities are calling datastore.create and datastore.save multiple times. It appears 
```
1585933641.570952 DEBUG gwebsockets: Received text message {"method":"datastore.create","id":1,"params":[{}],"jsonrpc":"2.0"}
1585933641.581081 DEBUG gwebsockets: Got data, length 144
1585933641.581456 DEBUG gwebsockets: Received text message {"method":"datastore.create","id":2,"params":[{}],"jsonrpc":"2.0"}
1585933641.582134 DEBUG gwebsockets: Received text message {"method":"datastore.create","id":3,"params":[{}],"jsonrpc":"2.0"}
1585933641.795370 DEBUG gwebsockets: Sending text message {"id": 1, "result": ["700a3ede-85f4-4521-b523-a1aff5576e39"], "error": null}
1585933641.807777 DEBUG gwebsockets: Got data, length 65
1585933641.807978 DEBUG gwebsockets: Received text message {"method":"open_stream","id":4,"params":[],"jsonrpc":"2.0"}
1585933641.808159 DEBUG gwebsockets: Sending text message {"id": 4, "result": [0], "error": null}
1585933641.809838 DEBUG gwebsockets: Got data, length 111
1585933641.810202 DEBUG gwebsockets: Received text message {"method":"datastore.save","id":5,"params":["700a3ede-85f4-4521-b523-a1aff5576e39",{},0],"jsonrpc":"2.0"}
1585933641.810851 DEBUG PY2-OLPC: DATA STORE API save()
1585933641.811007 DEBUG PY2-OLPC: uid
1585933641.811099 DEBUG PY2-OLPC: 700a3ede-85f4-4521-b523-a1aff5576e39
1585933641.811201 DEBUG PY2-OLPC:  metadata:
1585933641.811277 DEBUG PY2-OLPC: {}
1585933641.811353 DEBUG PY2-OLPC:  stream_id:
1585933641.811442 DEBUG PY2-OLPC: 0
1585933641.811682 DEBUG PY2-OLPC: OPENED A BINARY FILE @
1585933641.811780 DEBUG PY2-OLPC: /home/ss/.sugar/default/org.sugarlabs.FindWords/instance/1585933641
1585933641.811855 DEBUG PY2-OLPC: <type 'str'>
1585933641.811961 DEBUG gwebsockets: Sending text message {"id": 5, "result": [], "error": null}
1585933641.814559 DEBUG gwebsockets: Got data, length 242
1585933641.814864 DEBUG gwebsockets: Received binary message, length 167
1585933641.815016 DEBUG PY2-OLPC: Type of streamid:
1585933641.815107 DEBUG PY2-OLPC: <type 'str'>
1585933641.815245 DEBUG gwebsockets: Received text message {"method":"close_stream","id":6,"params":[0],"jsonrpc":"2.0"}
1585933642.030567 DEBUG gwebsockets: Sending text message {"id": 2, "result": ["660d37bd-dc29-4e09-b7e0-7e1c7c5e1aed"], "error": null}
1585933642.040888 DEBUG gwebsockets: Sending text message {"id": 3, "result": ["980c6265-d1f4-4fed-93aa-8f5b77ce80a8"], "error": null}
1585933642.041639 DEBUG gwebsockets: Got data, length 65
1585933642.042030 DEBUG gwebsockets: Received text message {"method":"open_stream","id":7,"params":[],"jsonrpc":"2.0"}
1585933642.042278 DEBUG gwebsockets: Sending text message {"id": 7, "result": [0], "error": null}
1585933642.043267 DEBUG gwebsockets: Got data, length 65
1585933642.043493 DEBUG gwebsockets: Received text message {"method":"open_stream","id":8,"params":[],"jsonrpc":"2.0"}
1585933642.043717 DEBUG gwebsockets: Sending text message {"id": 8, "result": [1], "error": null}
1585933642.045229 DEBUG gwebsockets: Got data, length 111
1585933642.045574 DEBUG gwebsockets: Received text message {"method":"datastore.save","id":9,"params":["660d37bd-dc29-4e09-b7e0-7e1c7c5e1aed",{},0],"jsonrpc":"2.0"}
1585933642.046489 DEBUG PY2-OLPC: DATA STORE API save()
1585933642.046629 DEBUG PY2-OLPC: uid
1585933642.046718 DEBUG PY2-OLPC: 660d37bd-dc29-4e09-b7e0-7e1c7c5e1aed
1585933642.046807 DEBUG PY2-OLPC:  metadata:
1585933642.046888 DEBUG PY2-OLPC: {}
1585933642.046963 DEBUG PY2-OLPC:  stream_id:
1585933642.047033 DEBUG PY2-OLPC: 0
1585933642.047356 DEBUG PY2-OLPC: OPENED A BINARY FILE @
1585933642.047769 DEBUG PY2-OLPC: /home/ss/.sugar/default/org.sugarlabs.FindWords/instance/1585933642
1585933642.047858 DEBUG PY2-OLPC: <type 'str'>
1585933642.047970 DEBUG gwebsockets: Sending text message {"id": 9, "result": [], "error": null}
1585933642.048737 DEBUG gwebsockets: Got data, length 112
1585933642.048973 DEBUG gwebsockets: Received text message {"method":"datastore.save","id":10,"params":["980c6265-d1f4-4fed-93aa-8f5b77ce80a8",{},1],"jsonrpc":"2.0"}
1585933642.049973 DEBUG PY2-OLPC: DATA STORE API save()
1585933642.050137 DEBUG PY2-OLPC: uid
1585933642.050227 DEBUG PY2-OLPC: 980c6265-d1f4-4fed-93aa-8f5b77ce80a8
1585933642.050316 DEBUG PY2-OLPC:  metadata:
1585933642.050388 DEBUG PY2-OLPC: {}
1585933642.050469 DEBUG PY2-OLPC:  stream_id:
1585933642.050540 DEBUG PY2-OLPC: 1
1585933642.050787 DEBUG PY2-OLPC: OPENED A BINARY FILE @
1585933642.050882 DEBUG PY2-OLPC: /home/ss/.sugar/default/org.sugarlabs.FindWords/instance/1585933642
1585933642.050962 DEBUG PY2-OLPC: <type 'str'>
1585933642.051066 DEBUG gwebsockets: Sending text message {"id": 10, "result": [], "error": null}
1585933642.051710 DEBUG gwebsockets: Got data, length 243
1585933642.052184 DEBUG gwebsockets: Received binary message, length 167
1585933642.052372 DEBUG PY2-OLPC: Type of streamid:
1585933642.052685 DEBUG PY2-OLPC: <type 'str'>
1585933642.052862 DEBUG gwebsockets: Received text message {"method":"close_stream","id":11,"params":[0],"jsonrpc":"2.0"}
1585933642.053323 DEBUG gwebsockets: Got data, length 243
1585933642.053517 DEBUG gwebsockets: Received binary message, length 167
1585933642.053631 DEBUG PY2-OLPC: Type of streamid:
1585933642.053714 DEBUG PY2-OLPC: <type 'str'>
1585933642.053843 DEBUG gwebsockets: Received text message {"method":"close_stream","id":12,"params":[1],"jsonrpc":"2.0"}
1585933642.170768 DEBUG gwebsockets: Sending text message {"id": 6, "result": [], "error": null}
1585933642.388364 DEBUG gwebsockets: Sending text message {"id": 11, "result": [], "error": null}
1585933642.390156 DEBUG PY2-OLPC: ERROOR:
1585933642.390406 DEBUG PY2-OLPC: org.freedesktop.DBus.Python.ValueError: Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/dbus/service.py", line 707, in _message_cb
    retval = candidate_method(self, *args, **keywords)
  File "/usr/lib/python2.7/dist-packages/carquinyol/datastore.py", line 375, in update
    lambda * args: self._update_completion_cb(async_cb,
  File "/usr/lib/python2.7/dist-packages/carquinyol/filestore.py", line 47, in store
    raise ValueError('No file at %r' % file_path)
ValueError: No file at dbus.String(u'/home/ss/.sugar/default/org.sugarlabs.FindWords/instance/1585933642')

1585933642.390633 DEBUG PY2-OLPC: INFO CLOSE REQUEST:
1585933642.390771 DEBUG PY2-OLPC: {u'jsonrpc': u'2.0', u'params': [1], u'method': u'close_stream', u'id': 12}
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 607, in msg_reply_handler
    *message.get_args_list()))
  File "/usr/lib/python2.7/dist-packages/jarabe/apisocket.py", line 228, in error_handler
    self._client.send_error(info["close_request"], error)
  File "/usr/lib/python2.7/dist-packages/jarabe/apisocket.py", line 292, in send_error
    self._session.send_message(json.dumps(response))
  File "/usr/lib/python2.7/json/__init__.py", line 244, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: DBusException(dbus.String(u'Traceback (most recent call last):\n  File "/usr/lib/python2.7/dist-packages/dbus/service.py", line 707, in _message_cb\n    retval = candidate_method(self, *args, **keywords)\n  File "/usr/lib/python2.7/dist-packages/carquinyol/datastore.py", line 375, in update\n    lambda * args: self._update_completion_cb(async_cb,\n  File "/usr/lib/python2.7/dist-packages/carquinyol/filestore.py", line 47, in store\n    raise ValueError(\'No file at %r\' % file_path)\nValueError: No file at dbus.String(u\'/home/ss/.sugar/default/org.sugarlabs.FindWords/instance/1585933642\')\n'),) is not JSON serializable
1585933647.442990 DEBUG gwebsockets: Got data, length 69
```

I have created a TCP Dump while `Opening Find-words-activity -> Clicking on Random Word Generate Button -> Closing the activity.` [Linked here.](https://drive.google.com/drive/folders/1Oc33X6ONxwzadm0otv_CSn0GqxpEpNxS?usp=sharing) 
The file is very large as I created it without filtering anything. I opened it in Wireshark and was unable to figure anything out, hopefully it makes sense to someone else. I have to spend some time and learn how to analyse these packets.
This happens in the Find-words-activity when I click on the button to generate some random words. I'll go through the activity code and try to figure something out.

shaansubbaiah:
@quozl I followed your advice and attempted to create a TCPdump using the Apisocket port. Upon termination, I get output saying 0 packets captured, file is empty. (There were multiple processes with the Findwords command, I chose the process ID from the one which had active CPU activity. 
**Edit:** checked the other processes as well, had the same port.)

I tried tracing the code executed by clicking the randomize words button in `activity.js` all the way to `{activity-directory}/lib/activity/sugar-web/datastore.js` file, did not find anything obvious that could have caused the issue.

Inspecting the activity from within Sugar here were some errors in the console. Not sure if this is directly related to the issue.
Screenshot_20200404_193121
Screenshot_20200404_193234
Screenshot_20200404_193255

chimosky:
It probably is, always test your changes within sugar as people use sugar.
My javascript isn't great so I can't help you here.

quozl:
`tcpdump` should be configured with the local network interface, usually `lo`.  Configuring it with the external network interface will show no data.

For the port number, if you can't be sure of it even after using `/proc/$PID/environ`, there are two other ways to verify it;

* change apisocket or gwebsockets to get the port number that the Gio.SocketService has bound to, and then write it to a file in `/tmp` for you to look at, (in gwebsockets/server.py, the port number is returned by the `Server.start` method, and in jarabe/apisocket.py),
* use `netstat -anp` and look for the `jarabe` process, and a TCP socket in LISTEN state,

shaansubbaiah:
I'll try this and post a filtered TCPDump. @quozl Could the issue in [env.js here](https://github.com/sugarlabs/sugar/pull/911#issuecomment-609039877) be linked to this?

shaansubbaiah:
TCPdump generated using `sudo tcpdump -i lo -s 0 -w ~/dump.pcap port 41709` [linked here](https://drive.google.com/open?id=1Oc33X6ONxwzadm0otv_CSn0GqxpEpNxS)

Again, the packets were captured while I clicked on the Random words button. 
The request is sent thrice, just like the error generated. `41709` is Apisocket port.
image

The rest of the information is in the dump file.

Also running 
```
$ netstat | grep 47914
tcp        0      0 localhost:47914         localhost:41709         ESTABLISHED
tcp6       0      0 localhost:41709         localhost:47914         ESTABLISHED

$ lsof -i :41709
COMMAND  PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
python2 1524   ss   10u  IPv6  33130      0t0  TCP localhost:41709->localhost:47914 (CLOSE_WAIT)
python2 1524   ss   28u  IPv6  28855      0t0  TCP *:41709 (LISTEN)
```
`pid 1524` corresponds to `python2 -m jarabe.main`

srevinsaju:
@shaansubbaiah it would be nice if you could refrain from uploadin to google drive. You can alternatively use MEGA.nz which is open source, or rather attach it here in github comment assets.
Thanks, maybe this will make it helpful for other mentors to review work too.

quozl:
@shaansubbaiah, I confirm by analysis with `tcpdump` and `wireshark` that you have captured some of the data exchanged between Sugar and the activity over the apisocket/gwebsocket connection.  Thanks.

The packet trace is not complete in the temporal domain; there are no SYN, SYN ACK, FIN, or FIN ACK packets.  You can fix that by starting the `tcpdump` before the activity is started, and keeping `tcpdump` running until the activity has finished.

I agree with @srevinsaju that attaching this tcpdump, which was 5114 bytes, to a GitHub comment would be more efficient for us than using Google Drive.  (Google Drive's supporting HTML and JavaScript exceeds the size of the file by orders of magnitude).  However, I don't foresee you having to do it much, as I expect you to analyse the data yourself and to cite any relevant sections.

Data in the packet trace is certainly recognisable as WebSocket protocol.  It looks like what I saw when I was finishing up the port of gwebsockets.  Wireshark contains a WebSocket protocol dissector, but I've not figured out how to use it yet.  Maybe you will.  Here are the references that I found;
* https://www.wireshark.org/docs/dfref/w/websocket.html
* https://wiki.wireshark.org/WebSocket
* https://tewarid.github.io/2017/04/25/wireshark-dissector-in-lua-for-custom-protocol-over-websockets.html
* https://www.wireshark.org/docs/wsdg_html_chunked/ChapterDissection.html
* https://www.wireshark.org/docs/wsug_html_chunked/ChCustProtocolDissectionSection.html

Getting back to the problem at hand; "how a [JavaScript activity that writes to the journal twice doesn't succeed in doing so](https://github.com/sugarlabs/sugar/pull/911#issuecomment-606668667). So I suggested testing to see if [this is a Python 3 porting regression](https://github.com/sugarlabs/sugar/pull/911#issuecomment-607057922)."

You have proof from logs that the activity writes to the Journal twice.  As far as I know, and unless someone can tell me otherwise, _there's nothing wrong with an activity doing that_, and if it doesn't work then something somewhere should be fixed.  Not just fixed so that we don't get an error, but fixed so that the data is properly stored and can be recalled.

You can use the packet trace to confirm that the activity writes to the Journal twice.

You may already have this proof from the activity source code, I don't know.

You already proved that the Python 2 build of Sugar and gwebsockets didn't properly handle writing to the Journal twice.  So the problem is latent before porting, and so your pull request commits are suggesting a change that may not be correct.

shaansubbaiah:
@srevinsaju thanks for the tip! I wasn't aware that I could attach non image files to GitHub comments.

@quozl there are two issues: 

1] Web activities calling the datastore functions multiple times to load/store activity data.
2] The Type errors due to Python 2 and 3's difference in indexing a bytes buffer.

The 2 issues are independent of each other. I was facing the web activity issue before applying the changes in the PR.
As you said, the web activity issue doesn't break anything, only causes those extra errors when data in the instance directory has already moved.

The Type Errors however prevent activities like the Gears activity from loading data stored in the Journal `Create gears in activity -> Close -> Open activity  -> Data doesn't load, no gears present` Comparing the 'type' of data between the **Live Build v117 running Python3** and **OLPC 18.04 running Python 2**: 

Live Build v117, stream_id type = int
```
1586440656.674570 DEBUG PY3-VM: message.data[0] of type:
1586440656.674726 DEBUG PY3-VM: <class 'int'>
```  

OLPC 18.04, stream_id type = str
```
1586441221.329491 DEBUG PY2-OLPC: Type of streamid:
1586441221.329558 DEBUG PY2-OLPC: <type 'str'>

1586441221.329618 DEBUG PY2-OLPC: Ord(stream_id) type:
1586441221.329673 DEBUG PY2-OLPC: <type 'int'>
```
stream_id is supposed to be an Int and ord() is not required as explained in the commit.

and the data loaded at DATASTOREAPI.save.on_data :
Live Build v117, data of type = bytes
```
1586441759.834844 DEBUG PY3-VM: DATASTOREAPI.save.on_data(data), data of type:
1586441759.834897 DEBUG PY3-VM: <class 'bytes'>

1586441759.834958 DEBUG PY3-VM: data[1:].decode('utf-8') of type:
1586441759.835013 DEBUG PY3-VM: <class 'str'>
```

OLPC 18.04, data of type = str
```
1586441868.168910 DEBUG PY2-OLPC: DATASTOREAPI.save.on_data(data), data of type:
1586441868.168946 DEBUG PY2-OLPC: <type 'str'>
```
write() function requires parameter of type str. In python 2 the bytes type is the same as the str type. So write() had not issues earlier.

The commit fixes these and data from the Journal loads properly.

quozl:
@shaansubbaiah, thanks, great explanation.  Please amend your commit and add some of your explanation there.  Sugar is a critical repository, so see [making commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits) and [https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#guide-for-reviewers](guide for reviewers).

You have confirmed Gears is able to save and load data.

I've gone back through the pull request to check for unresolved questions.

About 25 days ago, I observed;

> For the change to DatastoreAPI.load.on_data, your fix is to decode the bytes into utf-8, [...] . The error happened when the file is created and written to. I don't see a corresponding change to when the file is read. Are you sure the file is read and sent back to the client intact? How did you test that? Might it be better to change the file open mode (in the two places) to binary?

> Another test is whether you can resume a web activity from the journal and have the previously saved data properly loaded, even if it is binary. You may have to look for a web activity that does that. I can't remember which ones do.

Looking at apisocket.py, the file is not opened in binary mode, and this suggests that while binary data might have been supported by Python 2, it is certainly not supported by Python 3.

I speculate that this was not considered in the port to Python 3, instead `2to3` was run followed by some testing without measured coverage.  What do you think?

chimosky:
> I speculate that this was not considered in the port to Python 3, instead 2to3 was run followed by some testing without measured coverage. What do you think?

I agree, I was looking at the changes made to the datastore and they're things we should do differently for compatibility for both python3 and python2 activities.

shaansubbaiah:
@quozl agree, like @chimosky mentioned, in the porting process I feel care must have been taken where `binary` type data was used, as it was used interchangeably with the `str` type in Python2.

> Might it be better to change the file open mode (in the two places) to binary?

As to my knowledge, data in the datastore is stored as text/string, so the current way of opening as text mode is correct. The APIClient passes messages of binary type to the APIServer which calls the Datastore API, Activity API and open_stream, close_stream functions.

```
    def _message_received_cb(self, session, message, client):
        if message.message_type == Message.TYPE_BINARY:
            stream_id = ord(message.data[0]) // ord() removed in PR
```

quozl:
> As to my knowledge, data in the datastore is stored as text/string, 

Since the start, datastore is for binary data.  Interpretation of the data is left to the activities.  A content type is kept for reference and to associate with activities.

@shaansubbaiah, which web activities did you survey to ensure they stored data as text/string?

@llaske, doesn't the Sugarizer datastore support binary too?

shaansubbaiah:
@quozl I might have called the wrong location the datastore. There are files in the 1]`~/.sugar/default/datastore/`, 2]`~/.sugar/default/data`and some sort of temporary storage at 3]`~/.sugar/default/<activity_name>/`

For example the Gears Activity temporarily stores data in the instance folder eg `/home/shaan/.sugar/default/org.sugarlabs.GearsActivity/instance/1586609408` some file is created by DatastoreAPI._create_file() which creates the file in text mode.

Activity data is kept in an object in plaintext with the metadata at `/home/shaan/.sugar/default/datastore/63/634fc302-4b74-4b13-b82e-30ab18e13e08/ containing data, metadata` [634fc....e08] -> activity uid.

There's also Activity data stored with the uid in the same object in plaintext format at `~/.sugar/default/<activity_name>/` with the uid as filename.

Similarly,
the Find Words Activity stores data at 1] , 3]
the Moon Activity stores data at 1] , 2] , 3]

The wiki does state that the instance directory is temp storage but I'm not very clear on the other 2 locations- Data/ and Datastore/

llaske:
@quozl Sugarizer datastore content is stored as string. It's usually serialized Javascript object used to save activity context. It supports binary (images and even LZ compressed content) encoded as base64.
Locally the datastore was initially stored in browser localStorage but it's now in a IndexedDB database of the browser.
On the server, the content is stored in a MongoDB database.

quozl:
Thanks both.

I've reviewed bus.js, datastore.js, and activity.js in sugar-web.  First time.  I don't fully understand them, but I agree that they seem to support text without supporting unencoded binary.

I know that Sugar Python activities can store binary data; such as PNG, JPEG, OGG, and other formats.

I've searched for use of datastoreObject in https://github.com/sugarlabs and found 63 matches, and reviewed use of the datastore API among the Sugar Web activity collection.  None of them seem to store binary data.  I didn't look at Sugarizer activities, as @llaske has answered on that, and GitHub didn't give any matches for forks(!).

Data on the API socket arrives and leaves as Python object type `bytes`.  What I'm still not clear on is whether that data should be converted to `str` for writing to a file opened in text mode (the default, and present state of the code), or whether it should be written unconverted to a file opened in binary mode.  Both would be functionally identical as far as an activity is concerned, but the conversion method would require extra CPU resources and the activity to always submit data in a supported character set.  At the moment, a conversion does occur.  Why do we need to do this conversion only for Python to undo it?

shaansubbaiah:
I agree that data stored as `bytes` prevents the conversions every time the API is called but storing as text makes it very easy to view, modify data directly, especially while debugging. Some activities may want to use string functions on the data (then again it could have been decoded to `str` if required). Currently data is stored using the default Linux encoding-`utf-8`.

Text files are also less prone to corruption and compressed text files are generally smaller than binary files. I don't think the CPU overhead is significant enough to outweigh the simplicity the text files provides.

I would prefer that the data be stored as text, what are the organization's views on this?

quozl:
Thanks for thinking about this, but you're wrong in several respects.

Regardless of whether Python opens the file as text or binary, there's no difference to the bit stream in the file.  There will be no impact on viewing, modifying, or debugging.  What activities do with the data has nothing to do with how it is stored by the API socket class.

I don't believe text files are less prone to corruption.  We don't compress the files, but as the bit stream in the files would be no different, compression would equally be no different.

Demonstration of no difference;

```python
#!/usr/bin/python3

# an original text string
d0 = 'Regardless of whether Python opens the file as text or binary'

# convert to bytes for transmission through a socket
d1 = d0.encode()

# imagine the transmission

# option 1, the long way around

# convert to string
d2 = d1.decode()

# write to a text file
open('/tmp/data-text', 'w').write(d2)

# option 2, the short method

# write to a text file
open('/tmp/data-binary', 'wb').write(d1)

# test that the files are the same
import subprocess

subprocess.call(['/usr/bin/md5sum', '/tmp/data-text', '/tmp/data-binary'])
rc = subprocess.call(['/usr/bin/cmp', '/tmp/data-text', '/tmp/data-binary'])
if rc == 0:
    print("same")
```

shaansubbaiah:
@quozl I was wrong, storing characters in either file mode yields the same result, both can be viewed like normal text files. I had got confused with how it worked in C with .bin files. Turns out after testing in C, it works similarly. Might have stored and viewed non string files earlier that led to this confusion.

Thank you for the correction, I now see that the conversion is unnecessary.

shaansubbaiah:
I've made some progress towards making it read, write as binary files. It needs more testing, have only tested with the Gears activity as of now (Saving and Resuming works properly). I'll try out other activities for proper coverage.

Some changes being:
1] Read, write files in binary mode (rb, wb)
2] Remove the part where data is decoded to UTF format
3] stream_id is an integer as mentioned earlier, function send_binary() used to send a string (concatenating stream_id and data). Now that data is bytes, made stream_id as bytes.

quozl:
Thanks.  Reviewed changes.  Looks okay.

Another review by interpreting the change relative to the Python 2 source;

```log
git diff 69e3a01545eee18ced09c4b1505ecbe8bbebd934..HEAD -- src/jarabe/apisocket.py
```

Looks okay.

Another review using;
* [Python 3 Porting Guide](https://github.com/sugarlabs/sugar-docs/blob/master/src/python-porting-guide.md),
* [dbus-python Porting to Python 3](https://dbus.freedesktop.org/doc/dbus-python/PY3PORT.html), and;
* [Supporting Python 3: An in-depth guide](http://python3porting.com/).

Looks okay.

This extends and seeks to complete the aa18879e9717dfe2d30f249549e9a43d6dd6da4f ("Port to Python 3") for `apisocket.py`, so that's what the commit message should eventually become.

Last thing I've noted as open;

* [ ] error handlers are still coded to send the error back to the client, per [my earlier comment](https://github.com/sugarlabs/sugar/pull/911#issuecomment-599771706), instead of converting the exception into something the client can be given.

shaansubbaiah:
Thanks for reviewing the current progress. I didn't quite catch what you meant by 'instead of converting the exception into something the client can be given.' Could you please explain in detail?

quozl:
Sure.  At the moment in `apisocket.py` when a D-Bus exception occurs, the error handler is called with an argument which contains the exception.  This cannot be translated using json, and so cannot be sent to the client.  However, the client should be informed in some way that the API call failed, and if possible why it failed.  You might need to look at sugar-web JavaScript to figure out what is expected when a call fails.

shaansubbaiah:
Regarding the DBUSException error not being handled properly,

Error generated at `self._client.send_error(info["close_request"], error)`:

TypeError: Object of type DBusException is not JSON serializable at `self._session.send_message(json.dumps(response))` where `response` is an object with `error` as a parameter (should be JSON or string, here it is of type DBusException)

Where:
error.  (type dbus.exceptions.DBusException)
```
        org.freedesktop.DBus.Python.ValueError: Traceback (most recent call last):
        File "/usr/lib/python3/dist-packages/dbus/service.py", line 707, in _message_cb
            retval = candidate_method(self, *args, **keywords)
        File "/usr/lib/python3.7/dist-packages/carquinyol/datastore.py", line 373, in update
            lambda * args: self._update_completion_cb(async_cb,
        File "/usr/lib/python3.7/dist-packages/carquinyol/filestore.py", line 49, in store
            raise ValueError('No file at %r' % file_path)
        ValueError: No file at dbus.String('/home/shaan/.sugar/default/org.sugarlabs.FindWords/instance/1587569354')
```
      

info["close_request"].  (type JSON)
        ```
{'method': 'close_stream', 'id': 21, 'params': [1], 'jsonrpc': '2.0'}```

---
TEMPORARY FIX
Making error a `string` with the Name and Traceback of the Exception (https://dbus.freedesktop.org/doc/dbus-python/dbus.exceptions.html):
```        
        if(type(error) == dbus.exceptions.DBusException):
            dbus_error = error.get_dbus_name() + " " + error.get_dbus_message()
            error = dbus_error
```

---

The error generated before fix (Live Build v117)
```Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 607, in msg_reply_handler
    *message.get_args_list()))
  File "/usr/lib/python3.7/dist-packages/jarabe/apisocket.py", line 252, in error_handler
    self._client.send_error(info["close_request"], error)
  File "/usr/lib/python3.7/dist-packages/jarabe/apisocket.py", line 328, in send_error
    self._session.send_message(json.dumps(response))
  File "/usr/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type DBusException is not JSON serializable
```

Output after (Error only in the second line as the first call was able to access the data at the /instance folder)
```
DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 13}
DEBUG gwebsockets: Sending text message {"result": null, "error": "org.freedesktop.DBus.Python.ValueError Traceback (most recent call last):\n  File \"/usr/lib/python3/dist-packages/dbus/service.py\", line 707, in _message_cb\n    retval = candidate_method(self, *args, **keywords)\n  File \"/usr/lib/python3.7/dist-packages/carquinyol/datastore.py\", line 373, in update\n    lambda * args: self._update_completion_cb(async_cb,\n  File \"/usr/lib/python3.7/dist-packages/carquinyol/filestore.py\", line 49, in store\n    raise ValueError('No file at %r' % file_path)\nValueError: No file at dbus.String('/home/shaan/.sugar/default/org.sugarlabs.FindWords/instance/1587574588')\n", "id": 14}
```

---
The error could have been converted at `DatastoreAPI.save().error_handler(error)` but I wasn't sure if DBUSExceptions could occur elsewhere. Putting it within `APIClient.send_error()` will check every error sent from the client.

Right now the error passed is a string containing the error name and traceback. It's very long and appears in a single line so it is slightly difficult to read.

I'm not sure of the exact format of the errors passed by Sugar. I went through the sugar-web repo and didn't find anything. Only found that the functions pass the error in the callback like https://github.com/sugarlabs/sugar-web/blob/cef6655e7230759aadc23f4c9641b285f845f17c/bus.js#L209 .

quozl:
Thanks, well done.  What does the client do with it?

shaansubbaiah:
Went through sugar-web again, `self._session.send_message(json.dumps(response))` calls gwebsockets send_message() and on the client side upon receiving a message with an error it should throw a new error with the contents of the sent error.

```javascript
client.onMessage(){
.
.
.
                if (parsed.error === null) {
                    callback(null, parsed.result);
                } else {
                    callback(new Error(parsed.error), null);
                } 
.
}
```

Right now the error is sent as a message but it does not throw it in shell.log

quozl:
Is the error shown to the user?

shaansubbaiah:
It is not shown to the user explicitly, just gets logged in `shell.log` as a gwebsockets message (when logging is enabled in debug settings)

FindWords Activity 
`DEBUG gwebsockets: Sending text message {"result": null, "error": "org.freedesktop.DBus.Python.ValueError Traceback (most recent call last):\n  File \"/usr/lib/python3/dist-packages/dbus/service.py\", line 707, in _message_cb\n    retval = candidate_method(self, *args, **keywords)\n  File \"/usr/lib/python3.7/dist-packages/carquinyol/datastore.py\", line 373, in update\n    lambda * args: self._update_completion_cb(async_cb,\n  File \"/usr/lib/python3.7/dist-packages/carquinyol/filestore.py\", line 49, in store\n    raise ValueError('No file at %r' % file_path)\nValueError: No file at dbus.String('/home/shaan/.sugar/default/org.sugarlabs.FindWords/instance/1587574588')\n", "id": 14}`

Moon Activity
```
DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 13}
DEBUG gwebsockets: Sending text message {"result": null, "error": "org.freedesktop.DBus.Python.ValueError Traceback (most recent call last):\n  File \"/usr/lib/python3/dist-packages/dbus/service.py\", line 707, in _message_cb\n    retval = candidate_method(self, *args, **keywords)\n  File \"/usr/lib/python3.7/dist-packages/carquinyol/datastore.py\", line 373, in update\n    lambda * args: self._update_completion_cb(async_cb,\n  File \"/usr/lib/python3.7/dist-packages/carquinyol/filestore.py\", line 49, in store\n    raise ValueError('No file at %r' % file_path)\nValueError: No file at dbus.String('/home/shaan/.sugar/default/com.garycmartin.Moon/instance/1587713909')\n", "id": 14}
```

Would it be better to 
1] Print the error in the log properly (like how other errors display, showing traceback with proper formatting eg.
```log
1587714611.466397 DEBUG root: _Account.__set_current_activity_cb
/usr/lib/python3/dist-packages/gi/overrides/Gtk.py:1641: Warning: g_value_transform: assertion 'G_IS_VALUE (src_value)' failed
  return _Gtk_main(*args, **kwargs)
/usr/lib/python3/dist-packages/gi/overrides/Gtk.py:1641: Warning: unable to set property 'buddy' of type 'PyObject' from value of type '(null)'
  return _Gtk_main(*args, **kwargs)
1587714616.034032 DEBUG root: launch bundle_id=org.sug.......
```

2] Leave it as is, show the error in the message as we are not planning to fix the error?

I had expected the change made in the DBUSException handling commit to function like [1] but unfortunately it doesn't. Perhaps the format of the error is wrong? Is there a way of throwing some known error in apisocket so I could inspect its format? I couldn't find anything helpful in the gwebsockets and sugar-web repos.

quozl:
I'm getting a bit lost, sorry.  My recent replies have been regarding the general handling of errors; how they are logged, or how they are displayed.  Goal is software that does the right thing, or the expected thing, and people expect to be told if something isn't done because of either (a) a flaw in the activity, or (b) some failure relating to the filesystem, such as corruption or read-only remounts.

For problems involving loss of user data, reporting this _directly_ and _instantly_ to the user is needed.  Logs can contain additional detail, but people don't look at logs unless they are told there is a problem.

So assuming a web activity runs, and a datastore exception occurs, how is the user informed?

shaansubbaiah:
There is an issue where multiple datastore (store, load) calls by an activity are made, where some of them fail because data has already moved by a previous call. We were unable to find the cause of the issue and concluded that since it didn't affect the working of the activity it wasn't very important.

However, every time the issue occurred, it generated a DBUSException which could not be passed as an error in the APISocket error handling function. The temporary fix here only converts the DBUSException to string so that is sent to the error handler function and the error can be viewed in the logs.

I do not think this issue needs to be reported to the user as it doesn't affect the activity's working and occurs quite frequently.

What I understood by your earlier message was that there should be a method to report corruption/read-only issues directly to the user. I think this will require creating some sort of popup in the activity displaying the error so the user gets to know when it occurs. Regarding this, I am still not very sure how the errors are handled to come up with such a fix at the moment.

quozl:
Thanks for summarising the problems.  This is exactly the sort of thing that should be in commit messages.

> There is an issue where multiple datastore (store, load) calls by an activity are made, where some of them fail because data has already moved by a previous call. We were unable to find the cause of the issue and concluded that since it didn't affect the working of the activity it wasn't very important.

We need to fix this issue.  I don't see why multiple datastore calls aren't allowed.  Is this because the calls occur within the same clock second?  `"%i" % time.time()` seems to be remarkably non-unique as a filename.  `"%s" % time.time()` is likely to be more unique, but still, on faster and faster systems or where clock values are cached, this method of creating a temporary file name seems unwise.  There's a whole module `tempfile` for the knotty problem of creating unique temporary files.  Or just a sequence number would be fine?

> However, every time the issue occurred, it generated a DBUSException which could not be passed as an error to the APISocket error handling function. The temporary fix here only converts the DBUSException to string so that is sent to the error handler function and the error can be viewed in the logs.
> 
> I do not think this issue needs to be reported to the user as it doesn't affect the activity's working and occurs quite frequently.

It will result in data loss.  If the activity saves to the datastore, then changes the data, then saves again, it is not clear what is saved, and there is no message to the user.

> What I understood by your earlier message was that there should be a method to report corruption/read-only issues directly to the user. I think this will require creating some sort of popup in the activity displaying the error so the user gets to know when it occurs. Regarding this, I am still not very sure how the errors are handled to come up with such a fix at the moment.

We need to be able to focus on everything at once, not one repository at a time.  The investigation into these exceptions has revealed severe deficiencies in the API and sugar-web.  If an activity is asked to save data by the user, but cannot do so, then it must tell the user.  I don't know how yet.

aperezbios:
@shaansubbaiah , how close do you think you are to addressing the remaining changes @quozl has requested? This is an important fix, and I'd love to see it committed.

shaansubbaiah:
@aperezbios Sorry I've been a little busy with tests and submissions this week, I still haven't figured out a way to communicate issues to the user.

@quozl you were right regarding the issue occuring due to the calls at the same clock second. Just switching the `"%i" % time.time()` to a random 10 digit number while testing has removed the DBUS Errors.

Before:
```
DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":20,"params":[0],"jsonrpc":"2.0"}'
DEBUG gwebsockets: Got data, length 174
DEBUG gwebsockets: Received binary message, length 166
DEBUG gwebsockets: Got data, length 68
DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":21,"params":[1],"jsonrpc":"2.0"}'
DEBUG gwebsockets: Got data, length 242
DEBUG gwebsockets: Received binary message, length 166
DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":22,"params":[2],"jsonrpc":"2.0"}'
DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 20}
DEBUG gwebsockets: Sending text message {"result": null, "error": "org.freedesktop.DBus.Python.ValueError", "id": 21}
DEBUG gwebsockets: Sending text message {"result": null, "error": "org.freedesktop.DBus.Python.ValueError", "id": 22}
```

After:
```
DEBUG gwebsockets: Received binary message, length 180
DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":20,"params":[0],"jsonrpc":"2.0"}'
DEBUG gwebsockets: Got data, length 256
DEBUG gwebsockets: Received binary message, length 180
DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":21,"params":[1],"jsonrpc":"2.0"}'
DEBUG gwebsockets: Got data, length 256
DEBUG gwebsockets: Received binary message, length 180
DEBUG gwebsockets: Received text message b'{"method":"close_stream","id":22,"params":[2],"jsonrpc":"2.0"}'
DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 20}
DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 21}
DEBUG gwebsockets: Sending text message {"result": [], "error": null, "id": 22}

```

quozl:
@aperezbios, thanks for your interest.  Port to Python 3 is not yet finished for Sugar, in turn because of too few developers.  Lots of bugs expected and not yet detected.  You are better off using Python 2 if you want good results with children.  We're working on it, and we have a project idea for GSoC to continue the work.  We will learn on 4th May if we have that project approved.

@shaansubbaiah, re 0421cef8759993c51072e19491bf0287d8f7b0f6;
* should use `isinstance`,
* does not need parentheses around conditional expression,
* does not need another local variable,
* rewritten in 9fa8cffb3c50afaeeb7566b14018104b43c7ac59,

Re: using time.time or large random number.  Thanks for testing the alternative.
* should use a sequence number, as there is only one server instance,
* written as 03c464355,

Re: ef26a9f649a962e22012022377b65adcf58019df and 08a46a9ae4dc48035c7881751018da5df6e8f0e7.
* best squashed because they touch the same code,
* rewritten as f1ac68488631ab153dd31b737819fd5cadc3b158.

I've looked through apisocket.py while imagining port to Python 3, and haven't come up with anything else to change.

I've pushed to _master_ branch after testing Sugar with Gears and Moon on Ubuntu 19.04.  Please test _master_ branch.  I'll close this pull request; it seems everything is done with it.

shaansubbaiah:
Thanks, tested on master branch with Moon, Find-Words, Gears Web Activities, found no issue :)

quozl:
Thanks for testing.

PR #919, opened 2020-05-24
PEP8 fixes

Saumya-Mishra9129:


Saumya-Mishra9129:
https://github.com/sugarlabs/sugar/blob/5b0e880692a679aa6c4179e6579b8202e2524c6a/src/jarabe/model/friends.py#L82

The above line or any such line shows a warning "Unexpected argument". I couldn't be able to get why it so.

extensions/cpsection/aboutcomputer/model.py
```diff
@@ -178,6 +178,8 @@ def get_wireless_firmware():
             continue
 
         try:
+            driver = None
+            version = None
```
quozl, line 182: Consider how `None` will be processed in later code; if _driver_ was set to _libertas_, then _version.startswith_ would fail.

extensions/cpsection/backup/view.py
```diff
@@ -224,6 +224,7 @@ def __init__(self, operation, view):
         # check the backend availables, if there are more than one
         # show to the user to select one
         if len(self._view.manager.get_backends()) > 1:
+            message = None
```
quozl, line 227: When _message_ is set to None, then `_ask_options` will fail on the `set_text` call.  What made you choose None?

src/jarabe/frame/activitiestray.py
```diff
@@ -411,7 +411,7 @@ def __init__(self, file_transfer):
         ToolButton.__init__(self)
 
         self.file_transfer = file_transfer
-        file_transfer.connect('notify::state', self.__notify_state_cb)
+        self.file_transfer.connect('notify::state', self.__notify_state_cb)
```
quozl, line 414: This is not correct.  (a) the warning is wrong, as IncomingFileTransfer is a subclass of BaseFileTransfer, which is a subclass of GObject.GObject, which certainly does have _connect_, and (b) the way you change it makes no sense; both variables will point to the same object instance.

src/jarabe/frame/activitiestray.py
```diff
@@ -452,11 +452,11 @@ class IncomingTransferButton(BaseTransferButton):
 
     def __init__(self, file_transfer):
         BaseTransferButton.__init__(self, file_transfer)
-
+        self.file_transfer=file_transfer
```
quozl, line 455: Missing spaces.  And not correct, for same reason as above.

src/jarabe/frame/activitiestray.py
```diff
@@ -452,11 +452,11 @@ class IncomingTransferButton(BaseTransferButton):
...
         self._ds_object = datastore.create()
 
-        file_transfer.connect('notify::state', self.__notify_state_cb)
-        file_transfer.connect('notify::transferred-bytes',
+        self.filetransfer.connect('notify::state', self.__notify_state_cb)
+        self.file_transfer.connect('notify::transferred-bytes',
```
quozl, line 459: Not correct, as mentioned earlier.

src/jarabe/model/filetransfer.py
```diff
@@ -54,7 +54,6 @@
 FT_REASON_LOCAL_STOPPED = 2
 FT_REASON_REMOTE_STOPPED = 3
 FT_REASON_LOCAL_ERROR = 4
-FT_REASON_LOCAL_ERROR = 5
```
quozl, line 57: 5 is valid.  See https://lazka.github.io/pgi-docs/TelepathyGLib-0.12/enums.html#TelepathyGLib.FileTransferStateChangeReason

Also occurs in collabwrapper/collabwrapper.py

src/jarabe/model/neighborhood.py
```diff
@@ -790,19 +790,20 @@ def __error_handler_cb(self, error):
...
+        self.account.connect('buddy-left-activity', self.__buddy_left_activity_cb)
+        self.account.connect('activity-added', self.__activity_added_cb)
+        self.account.connect('activity-updated', self.__activity_updated_cb)
+        self.account.connect('activity-removed', self.__activity_removed_cb)
+        self.account.connect('current-activity-updated',
                         self.__current_activity_updated_cb)
-        account.connect('connected', self.__account_connected_cb)
-        account.connect('disconnected', self.__account_disconnected_cb)
+        self.account.connect('connected', self.__account_connected_cb)
+        self.account.connect('disconnected', self.__account_disconnected_cb)
```
quozl, line 806: All change to this file is incorrect, as stated earlier.

src/jarabe/model/network.py
```diff
@@ -701,7 +701,7 @@ def network_hash(self):
         else:
             fl |= 1 << 6
 
-        hashstr = str(fl) + '@' + self.ssid.decode()
+        hashstr = str(fl) + '@' + self.ssid
```
quozl, line 704: aa18879e9717dfe2d30f249549e9a43d6dd6da4f added this,   What makes you think self.ssid is a string?

src/jarabe/model/filetransfer.py
Saumya-Mishra9129: In that documentation REMOTE_ERROR= 5 and LOCAL_ERROR= 4

src/jarabe/frame/activitiestray.py
Saumya-Mishra9129: Yeah I agree now

extensions/cpsection/backup/view.py
Saumya-Mishra9129: Yeah you are probably right,But if none of the if statement runs , what value will be assigned to message. We must initialize message to some value before if statements.

src/jarabe/model/filetransfer.py
quozl: Please review https://github.com/sugarlabs/collabwrapper/pull/24

extensions/cpsection/backup/view.py
quozl: Yes, if you choose to change the code to defend against a vulnerability that a tool advises you about (that _operation_ is not one of the tested values), then you are duty bound to think about the outcomes of that change.  What value and type of _message_ will be accepted?  That's what you have to decide.

As for me, I don't bother making this kind of change, because I know the tool is giving advice in the context that is questionable.  There is no call to instantiate OperationPanel that has an _operation_ value other than the values that are tested for.

Is a more appropriate solution to rename OperationPanel as _OperationPanel so that it is treated as a private class?  Whether this is acceptable depends on the tool you are using; and I don't know what that tool is, nor which version.

Saumya-Mishra9129: > Is a more appropriate solution to rename OperationPanel as _OperationPanel so that it is treated as a private class? Whether this is acceptable depends on the tool you are using; and I don't know what that tool is, nor which version.

I am getting any such warning. I am using Pycharm as a tool.

Saumya-Mishra9129:
https://github.com/sugarlabs/sugar/blob/d59b62a5f1db8c1e1ecb7ccbbcc8b85163290908/src/jarabe/desktop/favoritesview.py#L201

The way we have defined _set_layout is not correct way. Setter should not return any value.

extensions/cpsection/backup/view.py
quozl: Okay, then I don't know why you are changing the code in this patch band.

Saumya-Mishra9129: Yeah The warning was probably given by that Pycharm tool and there is not a reason to solve that, I have removed those commits.

Saumya-Mishra9129:
@quozl @pro-panda Please review.

srevinsaju:
@Saumya-Mishra9129 although, not a big issue, there is a typo on https://github.com/sugarlabs/sugar/pull/919/commits/dbddda340d5ab5173936bf8d471b94dd15543627 's commit message

tests/extensions/cpsection/webaccount/web_service.py
```diff
@@ -16,7 +16,7 @@
 
 class WebService():
     def get_icon_name(self):
-        raise "Not implemented"
+        raise Exception("Not implemented")
```
srevinsaju, line 19:
```python
raise NotImplementedError
```

tests/extensions/cpsection/webaccount/web_service.py
```diff
@@ -16,7 +16,7 @@
...
 
     def config_service_cb(self, widget, event, container):
-        raise "Not implemented"
+        raise Exception("Not implemented")
```
srevinsaju, line 22:
```python
raise NotImplementedError
```

extensions/cpsection/webaccount/web_service.py
```diff
@@ -16,7 +16,7 @@
 
 class WebService():
     def get_icon_name(self):
-        raise "Not implemented"
+        raise Exception("Not implemented")
```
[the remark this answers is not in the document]
quozl: @walterbender, is the Web Services control panel section useful?  It doesn't seem to do anything at the moment.  If you click on it, you get "No web services are installed".  The [installer activity](http://activities.sugarlabs.org/en-US/sugar/addon/4723) hasn't been updated since 2013 and isn't on Sugar Labs GitHub but is in @i5o's GitHub https://github.com/i5o/install-extensions

Saumya-Mishra9129:
@quozl Can you please tell me why we used Webkit and Webkit2 support for Sugar.? I am a little confused about it.

quozl:
@Saumya-Mishra9129, sure, you can see in the commit descriptions of the commits that changed the imports.  Some systems on which Sugar is being used did not have WebKit2, so both versions are supported by Sugar.

Saumya-Mishra9129:
> @Saumya-Mishra9129, sure, you can see in the commit descriptions of the commits that changed the imports. Some systems on which Sugar is being used did not have WebKit2, so both versions are supported by Sugar.

Thanks
I am done with changes here @quozl you can review and merge.

PR #922, opened 2020-06-03
Major Fixes 

### LINKED ISSUE #855: Language control panel emits warning in logs
```log
(main.py:6165): Gtk-WARNING **: 17:56:49.381: Failed to set text
'<span foreground="#000000">Antigua & Barbuda</span>' from
markup due to error parsing markup: Error on line 1: Entity did not
end with a semicolon; most likely you used an ampersand character
without intending to start an entity — escape ampersand as &amp;
```

Saumya-Mishra9129:
Fixes #855 . @quozl @srevinsaju Please Review.

### DIFF
--- a/extensions/cpsection/modemconfiguration/model.py
+++ b/extensions/cpsection/modemconfiguration/model.py
@@ -318,8 +318,6 @@ class ServiceProviders(object):
         provider_name = self._settings.get_string(CONF_SP_PROVIDER)
         if not provider_name:
             provider_name = ''
-        else:
-            provider_name = provider_name.decode('utf-8')
 
         plan_idx = self._settings.get_int(CONF_SP_PLAN) or 0
 
--- a/extensions/cpsection/language/view.py
+++ b/extensions/cpsection/language/view.py
@@ -469,6 +469,7 @@ class FilterToolItem(Gtk.ToolItem):
     def set_widget_label(self, label):
         size = 'x-large'
         color = style.COLOR_BLACK.get_html()
+        label = GLib.markup_escape_text(label)
         span = '<span foreground="%s" size="%s">' % (color, size)
         self._label_widget.set_markup(span + label + '</span>')
 
@@ -486,6 +487,7 @@ class BlackLabel(PaletteMenuItem):
         self.set_label(text_label)
 
     def set_label(self, text_label):
+        text_label = GLib.markup_escape_text(text_label)
         text = '<span foreground="%s">' % style.COLOR_BLACK.get_html() + \
             text_label + '</span>'
         self.label.set_markup(text)
--- a/src/jarabe/desktop/favoriteslayout.py
+++ b/src/jarabe/desktop/favoriteslayout.py
@@ -247,7 +247,7 @@ class RandomLayout(SpreadLayout):
                     x = min(x, allocation.width - child_requisition.width)
                     y = min(y, allocation.height - child_requisition.height)
                 elif hasattr(child, 'get_bundle_id'):
-                    name_hash = hashlib.md5(child.get_bundle_id().decode())
+                    name_hash = hashlib.md5(child.get_bundle_id().encode('utf-8'))
                     x = int(name_hash.hexdigest()[:5], 16) % \
                         (allocation.width - child_requisition.width)
                     y = int(name_hash.hexdigest()[-5:], 16) % \
--- a/src/jarabe/desktop/meshbox.py
+++ b/src/jarabe/desktop/meshbox.py
@@ -486,7 +486,7 @@ class MeshBox(ViewContainer):
         # if we have mesh hardware, ignore OLPC mesh networks that appear as
         # normal wifi networks
         if len(self._mesh) > 0 and ap.mode == network.NM_802_11_MODE_ADHOC \
-                and ap.ssid == 'olpc-mesh':
+                and ap.ssid == b'olpc-mesh':
             logging.debug('ignoring OLPC mesh IBSS')
             ap.disconnect()
             return
--- a/src/jarabe/desktop/networkviews.py
+++ b/src/jarabe/desktop/networkviews.py
@@ -454,7 +454,7 @@ class WirelessNetworkView(EventPulsingIcon):
 
     def is_olpc_mesh(self):
         return self._mode == network.NM_802_11_MODE_ADHOC \
-            and self._ssid == 'olpc-mesh'
+            and self._ssid == b'olpc-mesh'
 
     def remove_all_aps(self):
         for ap in list(self._access_points.values()):
--- a/src/jarabe/journal/journalentrybundle.py
+++ b/src/jarabe/journal/journalentrybundle.py
@@ -87,7 +87,7 @@ class JournalEntryBundle(Bundle):
         preview_path = os.path.join(bundle_dir, 'preview', uid)
         if not os.path.exists(preview_path):
             return ''
-        f = open(preview_path, 'r')
+        f = open(preview_path, 'rb')
         try:
             preview_data = f.read()
         finally:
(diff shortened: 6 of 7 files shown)


chimosky:
Haven't reviewed 491305e yet but there was a mention of a similar traceback in the [mailing list](http://lists.sugarlabs.org/archive/sugar-devel/2020-June/058403.html), they're 5 definitions of `get_bundle_id()` in sugar, can you track and make sure they're all strings and not bytes so they wouldn't be decoded. Thanks.

Saumya-Mishra9129:
> Haven't reviewed 491305e yet but there was a mention of a similar traceback in the mailing list, they're 5 definitions of get_bundle_id() in sugar, can you track and make sure they're all strings and not bytes so they wouldn't be decoded. Thanks.

Thanks for suggestion , I am sure Its going to help me in further fixes.

Saumya-Mishra9129:
@chimosky @srevinsaju Please Review.

extensions/cpsection/modemconfiguration/model.py
```diff
@@ -319,7 +319,7 @@ def _get_initial_config(self):
         if not provider_name:
             provider_name = ''
         else:
-            provider_name = provider_name.decode('utf-8')
+            provider_name = provider_name
```
srevinsaju, line 322: I would suggest you to remove the `else` block as it doesn't make sense (if the change you are intending is right)

Saumya-Mishra9129: You are right. Fixed in cfec35e

srevinsaju: :+1:

chimosky:
Reviewed, thanks.

Your commit messages in 7ef6d2e...1e4709a  is misleading as you use `unicode` and `string` interchangeably and they both meant different things in python2.

As an instance, in 7ef6d2e you said;
" Use byte instead of unicode
ByteArray objects must be initialized with bytes objects, not unicodes."

but `bytes` is used instead of a `string` and not `unicode`.

Saumya-Mishra9129:
> Reviewed, thanks.
> 
> Your commit messages in 7ef6d2e...1e4709a is misleading as you use unicode and string interchangeably and they both meant different things in python2.
> 
> As an instance, in 7ef6d2e you said;
> " Use byte instead of unicode
> ByteArray objects must be initialized with bytes objects, not unicodes."
> 
> but bytes is used instead of a string and not unicode.

@chimosky I have done as suggested by you.

quozl:
Given my research as to the cause of https://github.com/sugarlabs/sugar/issues/923 and https://github.com/sugarlabs/sugar/pull/922/commits/efb71f6be3ff98593192aa2c1551373f7df087c4 it
doesn't look like you tested this before writing it (get_bundle_id should always returns a `str` so there's no need to check for type), so it doesn't look like the right solution.  How did you test?  Was there ever an instance where something other than a `str` was returned, and if so how did that happen?

Saumya-Mishra9129:
> Given my research as to the cause of #923 and looking only at efb71f6 it doesn't look like you tested this before writing it (get_bundle_id should always returns a str so there's no need to check for type), so it doesn't look like the right solution. How did you test? Was there ever an instance where something other than a str was returned, and if so how did that happen?

Thanks for reviewing. get_bundle_id always returns a str. I agree with you. The way I have done it is wrong. However md5 hash function accepts sequence of bytes. So rest part is correct.

quozl:
I'll wait for you to change the commit and message on https://github.com/sugarlabs/sugar/commit/efb71f6be3ff98593192aa2c1551373f7df087c4.  It seems clear to me that aa18879 used `decode` instead of `encode` and wasn't tested.

Saumya-Mishra9129:
> I'll wait for you to change the commit and message on efb71f6. It seems clear to me that aa18879 used decode instead of encode and wasn't tested.

Done in d97fff2

Saumya-Mishra9129:
I think this Pull request can merge now. @quozl @chimosky @JuiP @srevinsaju Please review and suggest.

quozl:
Thanks.  Reviewed.

* in https://github.com/sugarlabs/sugar/pull/922/commits/4de8eaf8217731cb39557a20e5a8c4a26b7a567c the warning does not mention _size_, but the code that uses _size_ is changed; is this change needed?  Also, it isn't clear why this has begun to be a problem now.  Also, there are many (30) other calls to set_markup that are not being protected by markup_escape_text; were these considered?  For instance if a journal entry is named `&`.  The commit message should explain.
* in https://github.com/sugarlabs/sugar/pull/922/commits/854b798cf6fa23d12b39289aa7b1cc7d10e6e9bd it isn't clear what problem is being solved.  The commit message should explain.
* in https://github.com/sugarlabs/sugar/pull/922/commits/04de77f913c1b1950fd39275fa81d9876b60053b under what circumstances will ConfigParser return anything that is not `str`, and if never why not apply the `encode` directly to the `config.get`?  The commit message should explain.

The commit messages are not in style for the Sugar repository;
* "Use GLib.markup_escape_text() before set_markup " could be "Fix Gtk-WARNING in language control panel",
* "Fix _read_preview : read in binary mode" would be better explained another way, but I don't yet know what the problem is that is being solved,
* "Use bytes as argument instead of string", "Use byte in ssid instead of string", "String argument without an encoding" and "invalid comparison - can raise to errors" could be "Use bytes for SSIDs",
* " 'str' object has no attribute 'decode'" should be "Fix AttributeError in SCENARIO'" where SCENARIO explains what is done to cause it, e.g. opening modem control panel.

Saumya-Mishra9129:
> in 4de8eaf the warning does not mention size, but the code that uses size is changed; is this change needed? Also, it isn't clear why this has begun to be a problem now. Also, there are many (30) other calls to set_markup that are not being protected by markup_escape_text; were these considered? For instance if a journal entry is named &. The commit message should explain.

@quozl I agree , if there are 30 calls to set_markup, we should protect them by markup_esacpe_text if there is a probability of getting '&', '>' or '<' sign. 

> in 04de77f under what circumstances will ConfigParser return anything that is not str, and if never why not apply the encode directly to the config.get? The commit message should explain.

Thanks Yeah congigparser will return str, You are correct.

Saumya-Mishra9129:
@quozl I have made requested changes , Please review.

quozl:
Thanks.  I've  cherry-picked and merged some of the patches.  Please rebase.

* 3f1fb1877d42f41686d711ee6d7af47a66a72812 was picked, the change to line 472 was removed (as I had pointed out the warning did not include _size_), and merged as ea2cd9c7b,
* 54dfb70, df715e3, f370dfe, 785f42e and 7024fc6 were picked, merged, commit message rewritten, and merged as 9900ccfbe,
* a211893 was picked, commit message rewritten, and merged as e0496b06d,
* 137584d was picked, commit message rewritten, and merged as 118fe73ee,

One of the patches was not cherry-picked;

* [ ] https://github.com/sugarlabs/sugar/pull/922/commits/637bbd84ebe3ee8fcf6ebe635475806c106b58fe wasn't clear to me from the commit message how to reproduce the problem, or what the problem is,

Saumya-Mishra9129:
> 637bbd8 wasn't clear to me from the commit message how to reproduce the problem, or what the problem is,

Thanks , the intuition behind this is that https://github.com/sugarlabs/sugar/blob/118fe73eee4177833dc692e1172451656f580b7f/src/jarabe/journal/journalentrybundle.py#L62
 reads the preview, and passes the output returned by _read_preview to dbus.ByteArray in
https://github.com/sugarlabs/sugar/blob/118fe73eee4177833dc692e1172451656f580b7f/src/jarabe/journal/journalentrybundle.py#L64 ,
 dbus.ByteArray takes bytes data as argument , but in _read_preview https://github.com/sugarlabs/sugar/blob/118fe73eee4177833dc692e1172451656f580b7f/src/jarabe/journal/journalentrybundle.py#L90 
 file is opened as 'r' only.

quozl:
Sorry for the delay, notification was filed as spam.  Thanks, merged as 05b4ca08d.

PR #939, opened 2021-01-27
v4.activities.sugarlabs.org

quozl:


quozl:
Tested.  I had to switch from my non-default updater to the Sugar default with;

```
gsettings set org.sugarlabs.update backend 'aslo.AsloUpdater'
```

It worked, but it was really slow.  Looking at it with _tcpdump_, Sugar is making one connection for each installed activity.  The connection is not cached.  Since the connection now uses https, the encryption setup time is multiplied by the number of installed activities.  With me in Australia, the server in the USA, and 39 activities, it took 42 seconds, We need a way to cache and re-use the connection.

* [ ] cache and re-use the connection,

src/jarabe/model/update/aslo.py
```diff
@@ -82,9 +82,8 @@ def __init__(self):
         self._bundle = None
 
     def check(self, bundle):
-        # ASLO knows only about stable SP releases
         major, minor = config.version.split('.')[0:2]
-        sp_version = '%s.%s' % (major, int(minor) + int(minor) % 2)
+        sp_version = '%s.%s' % (major, int(minor) + int(minor))
```
srevinsaju, line 86: Why is `int(minor)` added to itself?

srevinsaju:
> Tested. I had to switch from my non-default updater to the Sugar default with;
> 
> ```
> gsettings set org.sugarlabs.update backend 'aslo.AsloUpdater'
> ```
> 
> It worked, but it was really slow. Looking at it with _tcpdump_, Sugar is making one connection for each installed activity. The connection is not cached. Since the connection now uses https, the encryption setup time is multiplied by the number of installed activities. With me in Australia, the server in the USA, and 39 activities, it took 42 seconds. We need a way to cache and re-use the connection.
> 
> * [ ]  cache and re-use the connection,

Another possible solution is to use [Cloudflare](https://cloudflare.com) to manage our DNS. This would provide a faster Cache to any other country, including but not limited to Australia, as they have data centers all around the world. Alex and I had been recently talking about this. So, using this would be a huge benefit especially for children in countries with poor internet connection.

src/jarabe/model/update/aslo.py
quozl: Thanks, ec2bbd177.

quozl:
I don't think so; the DNS lookup occurs once and is cached.

srevinsaju:
> I don't think so; the DNS lookup occurs once and is cached.

Migrating our DNS to cloudflare would also help us to proxy certain domains through cloudflare's servers. Data which is transfered through their servers gets cached, and this cache would be served later on. This could possibly improve speed.


image

image

srevinsaju:
@codewiz gave me another idea of using a single http request to get the data of multiple activities. This would however, require client side as well as server side changes. 


**Proposed plan**
POST a json with the bundle id's of required activities to
```
https://v4.activities.sugarlabs.org/services/multipleAsloUpdater/
```

`aslo4-server` should reply with the data required for updating for all activities. The size of response will increase, for reducing the number of connections. What do you think?

quozl:
Sounds complicated.  Can't we just fix our use of Soup.Session in downloader.py and aslo.py so that the connection is not closed unless it needs to be closed?

quozl:
Regarding CDN, this doesn't fix slow last mile links, as experienced by school children sharing a limited connection.  It will be even worse for them than it is for me.

srevinsaju:
> Sounds complicated. Can't we just fix our use of Soup.Session in downloader.py and aslo.py so that the connection is not closed unless it needs to be closed?

Agreed, I was suggesting something to be done along with `Soup.Session`.

Issue #941, opened 2021-02-08
Language control panel fails to open with Sugar 0.118


aperezbios:
On a nightly build of Fedora SoaS, with 0.118 for all packages, the Language control panel does not open, and the shell.log has a message stating "AttributeError: 'NoneType' object has no attribute 'needs_restart'"

aperezbios:
`  File "/usr/lib/python3.9/site-packages/jarabe/controlpanel/gui.py", line 383, in __accept_clicked_cb
    if self._section_view.needs_restart:
AttributeError: 'NoneType' object has no attribute 'needs_restart'`

aperezbios:
This is the Fedora nightly ISO I tested with, which is 1.06GB in size: https://www.dropbox.com/s/kdry595y0s7ae0a/Fedora-SoaS-Live-x86_64-Rawhide-20210208.n.0.iso?dl=0

quozl:
Thanks.  7bbff9f59 ("Deepsource: unnecessary `else` or `elif`") was the most recent change, and was preceded by ea2cd9c7b ("Fix Gtk-WARNING in language control panel").

Each control panel is an extension that is imported with exceptions handled so that Sugar does not terminate when a buggy extension is added.  When regressions occur in extensions, the cause is hidden.  I suggest;

* check shell.log for an earlier messages during loading of the extension,
* temporarily turn off the exception handler in src/jarabe/controlpanel/cmd.py load_modules.

aperezbios:
Thanks, Walter. I've applied this patch manually and confirmed that it fixes the problem. I'm able to successfully switch languages to Spanish, and then back to English.

PR #942, opened 2021-02-09
Possible fix to #941

walterbender:
There was a decode() error:

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 7862: invalid continuation byte

Fixed by adding 'utf-8' and 'ignore' args to decode()

Not 100% sure that this is the only thing causing #941 but it is certainly part of the problem.


aperezbios:
I have manually applied this patch and confirmed that it fixes the problem. I am able to change languages to Spanish, successfully restart, and change language back to English.

PR #944, opened 2021-04-13
notifications: Fix signature of GetServerInformation

The-Compiler:
See https://specifications.freedesktop.org/notification-spec/latest/index.html, specifically the [GetServerInformation signature](https://specifications.freedesktop.org/notification-spec/latest/ar01s09.html).

I didn't test this myself as I only found this by accident while reviewing different notification daemons. You can test it using `dbus-send --session --print-reply --dest=org.freedesktop.Notifications /org/freedesktop/Notifications org.freedesktop.Notifications.GetServerInformation` which will show a `TypeError` without this change.

### DIFF
--- a/src/jarabe/model/notifications.py
+++ b/src/jarabe/model/notifications.py
@@ -104,9 +104,9 @@ class NotificationService(dbus.service.Object):
     def GetCapabilities(self):
         return []
 
-    @dbus.service.method(_DBUS_IFACE, in_signature='', out_signature='sss')
-    def GetServerInformation(self, name, vendor, version):
-        return 'Sugar Shell', 'Sugar', config.version
+    @dbus.service.method(_DBUS_IFACE, in_signature='', out_signature='ssss')
+    def GetServerInformation(self):
+        return 'Sugar Shell', 'Sugar', config.version, '1.2'
 
     @dbus.service.signal(_DBUS_IFACE, signature='uu')
     def NotificationClosed(self, notification_id, reason):


srevinsaju (approved):

src/jarabe/model/notifications.py
```diff
@@ -104,9 +104,9 @@ def CloseNotification(self, notification_id):
     def GetCapabilities(self):
         return []
 
-    @dbus.service.method(_DBUS_IFACE, in_signature='', out_signature='sss')
-    def GetServerInformation(self, name, vendor, version):
-        return 'Sugar Shell', 'Sugar', config.version
+    @dbus.service.method(_DBUS_IFACE, in_signature='', out_signature='ssss')
+    def GetServerInformation(self):
```
chimosky, line 108: Why remove all functions args as looking at the specification it still shows that the function definition has these args with `spec_version` as a new argument?

chimosky:
Don't really know how to test this so I watched the dbus session with `dbus-monitor` and running the above gives this output
```
method call time=1618363367.690996 sender=:1.30 -> destination=org.freedesktop.DBus serial=1 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=Hello
method return time=1618363367.691044 sender=org.freedesktop.DBus -> destination=:1.30 serial=1 reply_serial=1
   string ":1.30"
signal time=1618363367.691067 sender=org.freedesktop.DBus -> destination=(null destination) serial=60 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameOwnerChanged
   string ":1.30"
   string ""
   string ":1.30"
signal time=1618363367.691100 sender=org.freedesktop.DBus -> destination=:1.30 serial=2 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameAcquired
   string ":1.30"
signal time=1618363367.693877 sender=org.freedesktop.DBus -> destination=:1.30 serial=5 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameLost
   string ":1.30"
signal time=1618363367.693924 sender=org.freedesktop.DBus -> destination=(null destination) serial=61 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameOwnerChanged
   string ":1.30"
   string ":1.30"
   string ""
```

@quozl how should this be tested?

src/jarabe/model/notifications.py
The-Compiler: Those are C out arguments (since you can't easily return a tuple in C), which in Python are translated to returning a tuple.

The-Compiler:
@chimosky See my post above for a suggestion on how to test this :slightly_smiling_face:

src/jarabe/model/notifications.py
chimosky: Oh okay cos seeing as other message commands have their C out arguments in the function definition I was wondering.

chimosky:
> @chimosky See my PR description above for a suggestion on how to test this slightly_smiling_face

Oh I did test with that, my reply above was using `dbus-monitor` while testing with that, should've posted the output of the command when I ran it.
```
dbus[1004]: arguments to dbus_message_new_method_call() were incorrect, assertion "_dbus_check_is_valid_path (path)" failed in file ../../../dbus/dbus-message.c line 1366.
This is normally a bug in some application using the D-Bus library.

  D-Bus not built with -rdynamic so unable to print a backtrace
Aborted
```

src/jarabe/model/notifications.py
The-Compiler: They do? Where? I can't see any example of that.

chimosky: https://github.com/sugarlabs/sugar/blob/9b525fcfe13be741b1305b7a5aa2e126d8bc3628/src/jarabe/model/notifications.py#L99-L101 and other `@dbus.service.method` decorators.

The-Compiler: That method (as per the [notification spec](https://specifications.freedesktop.org/notification-spec/latest/ar01s09.html) and the `in_signature` / `out_signature` in the decorator) has a single input argument and no output arguments.

chimosky: Yeah, just checked that. Thanks.

Can you respond to my message below?

The-Compiler:
I've never seen that error message, no idea what it means - but I assume you see that without my changes too?

chimosky:
> I've never seen that error message, no idea what it means - but I assume you see that without my changes too?

Yeah I do.

PR #949, opened 2021-12-06
Fixed issue:Journal Entries can be renamed to blank via DetailView

chimosky:
@srevinsaju @quozl can you test?

This was moved from #663.

The comment made by @i5o here https://github.com/sugarlabs/sugar/pull/663#issuecomment-260526997 isn't a blocker as the `entry` doesn't handle when the `enter` key is pressed and I don't think it needs to, so pressing the `enter` key would have no effect.

### DIFF
--- a/src/jarabe/journal/expandedentry.py
+++ b/src/jarabe/journal/expandedentry.py
@@ -512,7 +512,19 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
 
         old_title = self._metadata.get('title', None)
         new_title = self._title.get_text()
-        if old_title != new_title:
+        if new_title == '' or new_title.isspace():
+            alert = Alert()
+            alert.props.title = _('Empty title')
+            alert.props.msg = _('The title cannot be empty')
+            ok_icon = Icon(icon_name='dialog-ok')
+            alert.add_button(Gtk.ResponseType.OK, _('Ok'), ok_icon)
+            ok_icon.show()
+            alert.connect('response', self._title_alert_response_cb)
+            journalwindow.get_journal_window().add_alert(alert)
+            alert.show()
+            new_title = self._title.set_text(old_title)
+
+        elif old_title != new_title:
             self._icon.palette.props.primary_text = new_title
             self._metadata['title'] = new_title
             self._metadata['title_set_by_user'] = '1'
@@ -539,6 +551,9 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
             self._write_entry()
 
         self._update_title_sid = None
+
+    def _title_alert_response_cb(self, alert, response_id):
+        journalwindow.get_journal_window().remove_alert(alert)
 
     def _write_entry(self):
         if self._metadata.get('mountpoint', '/') == '/':
--- a/src/jarabe/journal/expandedentry.py
+++ b/src/jarabe/journal/expandedentry.py
@@ -28,7 +28,7 @@ from sugar3.graphics import style
 from sugar3.graphics.xocolor import XoColor
 from sugar3.graphics.icon import CanvasIcon, get_icon_file_name
 from sugar3.graphics.icon import Icon, CellRendererIcon
-from sugar3.graphics.alert import Alert
+from sugar3.graphics.alert import Alert, ConfirmationAlert
 from sugar3.util import format_size
 from sugar3.graphics.objectchooser import get_preview_pixbuf
 from sugar3.activity.activity import PREVIEW_SIZE
@@ -255,6 +255,7 @@ class ExpandedEntry(Gtk.EventBox, BaseExpandedEntry):
 
         self._metadata = None
         self._update_title_sid = None
+        self._title_changed = False
 
         self.modify_bg(Gtk.StateType.NORMAL, style.COLOR_WHITE.get_gdk_color())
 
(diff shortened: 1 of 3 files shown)


quozl:
Thanks for asking.  I'm not in favour of the change.  My opinions are in #663, bug 4940 and #752.  I would accept the change if;
* the entry title is not restored to previous value,
* the message *title cannot be empty* is written as *title is usually not left empty*.

srevinsaju:
I restarted bugs.sugarlabs.org back, but I forgot to restore the domain names back. Will review this PR after getting to bug 4940

quozl:
Thanks.  My main comment was that the user experience was inconsistent between journal list and detail view.  There's also the activity toolkit.  Quoting from 4940 ...

> My opinion is the bug seems like censorship; why can't the learner use a blank or empty name for a journal object?  It is an unnecessary restriction on expression.  Why can't Sugar just do what it is told?  The learner can always discover how to rename again.
> So I'd like to see the first patch reverted, and instead patches made to the Help activity explaining the rename to blank feature.

srevinsaju:
Restored bugs.sugarlabs.org, 

The MR looks good overall, but, yea, I am not sure whats the best decision here.

chimosky:
> Thanks for asking. I'm not in favour of the change. My opinions are in #663, bug 4940 and #752. I would accept the change if;
> 
>     * the entry title is not restored to previous value,
> 
>     * the message _title cannot be empty_ is written as _title is usually not left empty_.

I think that the user being able to leave a journal object without any name isn't that bad but at the same time it might be nice to have a placeholder text to indicate there's no title as that might be a bit more helpful, I'll change the alert message.

@aperezbios @walterbender thoughts?

src/jarabe/journal/expandedentry.py
```diff
@@ -512,7 +512,19 @@ def _update_entry(self, needs_update=False):
 
         old_title = self._metadata.get('title', None)
         new_title = self._title.get_text()
-        if old_title != new_title:
+        if new_title == '' or new_title.isspace():
+            alert = Alert()
+            alert.props.title = _('Empty title')
+            alert.props.msg = _('The title cannot be empty')
+            ok_icon = Icon(icon_name='dialog-ok')
+            alert.add_button(Gtk.ResponseType.OK, _('Ok'), ok_icon)
```
srevinsaju, line 520: On a second thought, maybe we can have something in between prohibiting empty names, and allowing them. How about a warning instead of an alert, "The title is set to empty. Are you sure you want to proceed?", perhaps something like this :thinking: , thoughts?

chimosky: That does sound like a good idea but alerts are one major way to give additional info to the user and are usually not interactive, I don't know if they're any interactive implementations of such in sugar. @quozl does such exist?

aperezbios: @chimosky I'm in favor of asking the questtion to the user, as @srevinsaju has suggested. More often than not, I'd imagine a journal entry with no name would be something they would want to correct, and was likely unintentional.

chimosky: > @chimosky I'm in favor of asking the questtion to the user, as @srevinsaju has suggested. More often than not, I'd imagine a journal entry with no name would be something they would want to correct, and was likely unintentional.

Agreed, took a look and found `sugar3.graphics.alert.ConfirmationAlert()` which is perfect for this. 

If anyone else has no objection I'll implement with `ConfirmationAlert()`.

chimosky:
@quozl could you help me test and figure this out?

Currently, renaming and activity instance to blank via the listview uses a `ConfirmationAlert` and works as intended, activity name changes based on user response to the alert but while in detail view it doesn't work as intended, sometimes it changes to a blank when renaming to a blank regardless of the users' response but usually the first few changes don't seem to have any effect.

quozl:
Tested.  If I create an activity instance with an empty title (e.g. with Calculator and set the title empty), then open the Journal list view, start editing the name, but press enter without making a change; the alert appears, and if I press Cancel _the title changes to the last title on the list_.  Switching to Documents then back to Journal restores the title.

On the same empty title journal entry, view details, trying to change the title, the alert appears, but the Cancel button does not work, and the logs contain `TypeError: _title_alert_response_cb() takes 3 positional arguments but 5 were given`.

chimosky:
> Tested [a5689c8](https://github.com/sugarlabs/sugar/commit/a5689c80031e27a7345441e386098889b35cf833).
> 
> If I create an activity instance with an empty title (e.g. with Calculator and set the title empty), then open the Journal list view, start editing the name, but press enter without making a change; the alert appears, and if I press Cancel _the title changes to the last title on the list_. Switching to Documents then back to Journal restores the title.
> 
> On the same empty title journal entry, view details, trying to change the title, the alert appears, but the Cancel button does not work, and the logs contain `TypeError: _title_alert_response_cb() takes 3 positional arguments but 5 were given`.

My bad, I've fixed it in 12a9c44.

quozl:
Tested 12a9c44.

In detail view, pressing <kbd>enter</kbd> on the title does not show the alert, but <kbd>tab</kbd> works.  Also, pressing Cancel or Ok on the alert has to be done twice before the alert is dismissed.

For an instance with non-blank title, I click for detail view, empty the title, then press the back button.  The alert appears on the list view, not the detail view.  Presssing Cancel did not restore the title; it had changed to empty.

Some stack traces appear in shell.log;

```
  File "/usr/lib/python3/dist-packages/jarabe/journal/listview.py", line 374, in __buddies_set_data_cb
    buddy = tree_model.do_get_value(tree_iter, cell._model_column_index)
  File "/usr/lib/python3/dist-packages/jarabe/journal/listmodel.py", line 160, in do_get_value
    return self._cached_row[column]
IndexError: list index out of range
  File "/usr/lib/python3/dist-packages/jarabe/journal/listmodel.py", line 160, in do_get_value
    return self._cached_row[column]
IndexError: list index out of range
```

quozl:
Changing an entry in list view from non-empty to non-empty is ineffective; the title remains as it was.

chimosky:
> Tested [12a9c44](https://github.com/sugarlabs/sugar/commit/12a9c444f56cf5795b9bb77799aaa71eaf2d6779).
> 
> In detail view, pressing enter on the title does not show the alert, but tab works. Also, pressing Cancel or Ok on the alert has to be done twice before the alert is dismissed.

Yes pressing enter doesn't work as there's no handler for that event.

> For an instance with non-blank title, I click for detail view, empty the title, then press the back button. The alert appears on the list view, not the detail view. Presssing Cancel did not restore the title; it had changed to empty.

Yes it does as before now the title only gets saved when back to list view as that's the current behaviour at master, the issue I'm having is the title not changing despite Cancel being the response and so far I don't see why that's happening, added logging warnings to see what's happening and sometimes it seems like the code that sets the title runs before the `_title_alert_response_cb` as the warnings added there get printed after the warnings added in `if self._title_changed` of `_update_entry`

> Some stack traces appear in shell.log;
> 
> ```
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/jarabe/journal/listview.py", line 374, in __buddies_set_data_cb
>     buddy = tree_model.do_get_value(tree_iter, cell._model_column_index)
>   File "/usr/lib/python3/dist-packages/jarabe/journal/listmodel.py", line 160, in do_get_value
>     return self._cached_row[column]
> IndexError: list index out of range
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/jarabe/journal/listmodel.py", line 160, in do_get_value
>     return self._cached_row[column]
> IndexError: list index out of range
> ```

First time seeing this, can you remember what you'd done before this appeared?

chimosky:
@quozl @srevinsaju @aperezbios kindly review.

srevinsaju (approved):
Reviewed, looks good to me!

PR #951, opened 2022-01-22
Minimal bringup under GNOME Wayland

quozl:
Not for merging.

Demonstration of what is required to start the Sugar Home View from
within a GNOME Terminal on a Wayland session.

* fix a silent fail to start with a premature exit status 1 by removing
  all calls to the Wnck library; this breaks window switching to
  Neighbourhood View, Group View, or Journal, and detection of activity
  startup completion,

* fix a segmentation fault in XQueryExtension by removing the
  CursorTracker initialisation; this breaks The Frame,

* fix a segmentation fault in libxklavier by removing keyboard
  configuration,

* fix failure to redraw home view by removing GStreamer, both the thread
  initialisation and the toolkit SpeechManager; this breaks speech in
  the shell only.

Other regresssions;

- mouse pointer is "busy",

- not fullscreen, has title bar,

- view switching keys don't work,

- right-click and mouse-over menus are incorrectly positioned,

- My Settings, Keyboard does segmentation fault in libxklavier,

- Journal is not shown.

Things that work;

- new user dialog,

- home view, favourites view, list view,

- my settings; everything except Keyboard,

- the Log and Calculate activities.

### DIFF
--- a/src/jarabe/frame/eventarea.py
+++ b/src/jarabe/frame/eventarea.py
@@ -17,7 +17,6 @@ from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import GObject
 from gi.repository import GLib
-from gi.repository import Wnck
 
 from sugar3.graphics import style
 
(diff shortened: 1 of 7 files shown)


chimosky:
Thanks for looking at this, interesting changes so far and it seems like you can't grab an active window or Wayland automatically displays on a particular window. 

Can't really remember the details when I'd taken a look at this but I think it's one problem with Wayland.

PR #960, opened 2022-08-03

chimosky:
@quozl @srevinsaju kindly test and review.

Will fix object recopying to the clipboard after it's dragged and saved to the journal in another PR.

> It seems to have some weird conditions with drags from browse. If I drag an image from browse to the journal, it fails [1]. But if I go browse->clipboard then clipboard->browse it works.

This is still the case and I don't think it should change.

Seeing as this is a feature, would one commit be preferred?

### DIFF
--- a/src/jarabe/frame/activitiestray.py
+++ b/src/jarabe/frame/activitiestray.py
@@ -24,7 +24,6 @@ from gi.repository import GObject
 from gi.repository import Gio
 from gi.repository import Gtk
 from gi.repository import Gdk
-from gi.repository import GLib
 
 from sugar3.graphics import style
 from sugar3.graphics.tray import HTray
@@ -53,7 +52,7 @@ from jarabe.frame.notification import NotificationButton
 from jarabe.frame.notification import NotificationPulsingIcon
 from jarabe.frame.clipboardmenu import ClipboardMenu
 from jarabe.frame.clipboardobject import ClipboardObject
-from jarabe.frame.clipboardtray import _ContextMap
+from jarabe.frame.clipboardtray import ContextMap
 from jarabe.frame.clipboardobject import Format
 import jarabe.frame
 from jarabe.frame import clipboard
@@ -70,7 +69,7 @@ class ActivityButton(RadioToolButton):
         self._home_activity = home_activity
         self._notify_launch_hid = None
 
-        self._context_map = _ContextMap()
+        self._context_map = ContextMap()
         self._cb_object = None
         self._icon = NotificationPulsingIcon()
         self._icon.props.base_color = home_activity.get_icon_color()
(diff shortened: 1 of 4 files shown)


src/jarabe/frame/activitiestray.py
```diff
@@ -90,6 +109,44 @@ def __init__(self, home_activity, group):
...
+        return True
+
+    def __journal_drag_data_received_cb(self, widget, context, x, y, selection,
+                                        targetType, time):
+        data = None
+        if selection.get_pixbuf() is not None:
+            data = selection.get_pixbuf()
+        if len(selection.get_uris()) != 0:
+            data = selection.get_uris()[0].encode()
+            GLib.strfreev(data)
```
srevinsaju, line 134: What does L133-134 do?

src/jarabe/frame/clipboardmenu.py
```diff
@@ -176,13 +176,19 @@ def _journal_item_activate_cb(self, menu_item):
...
-            os.write(f, data)
+            if not is_pixbuf:
+                os.write(f, data)
+            else:
+                _type = self._cb_object.get_mime_type().split('/').pop()
+                data.savev(file_path, _type, [], [])
         finally:
-            os.close(f)
+            if not is_pixbuf:
+                os.close(f)
```
srevinsaju, line 191: I am not really sure, shouldn't we close the file even if we are not writing to; since `tempfile.mkstemp` retrurns an open file handle? Or is it fine? Also, if we have to close it, should we close the file before the `data.savev` function is called?

src/jarabe/frame/activitiestray.py
chimosky: It basically gets the first item from a list of URIs and encodes that data for `os.write`, [`GLib.strfreev`](https://lazka.github.io/pgi-docs/index.html#Gtk-3.0/classes/SelectionData.html#Gtk.SelectionData.get_uris) is called if the result is non-None and in this case it's expected that it's non-None.

src/jarabe/frame/clipboardmenu.py
chimosky: Agreed, I'd gotten an error earlier as I didn't see a reason not to close the file too and I can't remember the error and it seems it no longer happens, seeing as the `close` call is in a finally block, all operations happen before it.

quozl:
Minor review comments.

Perhaps _ContextMap from cliboardtray should not have underscore
prefix now that it is used elsewhere.

I'm surprised GLib.strfreev is needed, as usually the Python API
does all the memory management when objects are deleted.  And then to
use "data" shortly after it is freed doesn't make sense to me.

chimosky:
> Minor review comments. Perhaps _ContextMap from cliboardtray should not have underscore prefix now that it is used elsewhere. 

Agreed.

> I'm surprised GLib.strfreev is needed, as usually the Python API does all the memory management when objects are deleted. And then to use "data" shortly after it is freed doesn't make sense to me.

Yeah agreed, I'll change it. I'd added it as a measure as I didn't know if memory management for objects were done automatically, it was also something that wasn't used elsewhere in the API so it was difficult to add it. I'll remove it. Thanks!

chimosky:
Changes made, @quozl kindly review.

PR #964, opened 2023-03-25
Added Validation check for screen Issue in wayland

bhavyabansal9068:
Fix for Issue #965

### DIFF
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -198,13 +198,16 @@ def _restart_window_manager():
 def _start_window_manager():
     global _cursor_theme_settings, _cursor_theme
 
+    screen = Wnck.Screen.get_default()
+    if screen is None:
+        return sys.exit()
+
     _cursor_theme_settings = Gio.Settings.new('org.gnome.desktop.interface')
     _cursor_theme = _cursor_theme_settings.get_string('cursor-theme')
     _cursor_theme_settings.set_string('cursor-theme', 'sugar')
 
     _restart_window_manager()
 
-    screen = Wnck.Screen.get_default()
     screen.connect('window-manager-changed', __window_manager_changed_cb)
 
     _check_for_window_manager(screen)
@@ -340,21 +343,9 @@ def _check_group_label():
     return intro.check_group_label()
 
 
-def _validation_check():
-
-    screen = Wnck.Screen.get_default()
-    if screen is not None:
-        return True
-    else:
-        return False
-
-
 def main():
     Gst.init(sys.argv)
 
-    if not _validation_check():
-        sys.exit()
-        
     cleanup_temporary_files()
 
     _start_window_manager()
--- a/src/jarabe/main.py
+++ b/src/jarabe/main.py
@@ -350,11 +350,11 @@ def _validation_check():
 
 
 def main():
+    Gst.init(sys.argv)
+
     if not _validation_check():
         sys.exit()
         
-    Gst.init(sys.argv)
-
     cleanup_temporary_files()
 
     _start_window_manager()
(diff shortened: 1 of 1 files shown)


chimosky:
> Thanks for guiding @quozl . I have attached the error logs that we are getting after running Sugar from the GNOME terminal.
> 
> ```
> (main.py:3327): Wnck-WARNING **: 08:54:39.331: libwnck is designed to work in X11 only, no valid display found
> 
> (metacity:3334): metacity-WARNING **: 08:54:39.359: Option “--no-composite” is deprecated, use the “--compositor” instead.
> Traceback (most recent call last):
>   File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
>     exec(code, run_globals)
>   File "/usr/lib/python3/dist-packages/jarabe/main.py", line 380, in <module>
>     main()
>   File "/usr/lib/python3/dist-packages/jarabe/main.py", line 347, in main
>     _start_window_manager()
>   File "/usr/lib/python3/dist-packages/jarabe/main.py", line 208, in _start_window_manager
>     screen.connect('window-manager-changed', __window_manager_changed_cb)
> AttributeError: 'NoneType' object has no attribute 'connect'
> 
> (metacity:3334): metacity-WARNING **: 08:54:40.261: Screen 0 on display ":0" already has a window manager; try using the --replace option to replace the current window manager.
> ```
> 
> So, what I am thinking is, if we could add a validation check at the beginning of our main code to determine whether the screen is a window manager or not, it might help. Right? This way, before executing any process or job, our code will stop its execution."

The error you posted has a workaround which is to pass `GDK_BACKEND=x11` when running sugar from the terminal as we don't have Wayland support yet which is why you're getting the `NoneType` error.

> G'day, and thanks for your input. The problem described is that the process does not exit. The process does continue running. If you think that changing the cursor theme has anything to do with it, then please do test and let us know your result, either here or as a pull request. As for me, I don't see how the cursor theme can cause a process not to exit.

I also agree with Quozl.


Haven't tested your change yet but you should check your processes if Sugar still runs after your change as you didn't indicate that in your commit messages.

quozl:
Perhaps it would be better to check if `screen` is None in `_start_window_manager` and return if so.  Also, I don't understand why you moved the `Gst.init`, your commit message does not explain.

quozl:
You said this is a fix for issue 952, but I don't think it is.  Issue 952 is that Sugar does not exit.  These commits are to do with handling when _Wnck_ does not work under Wayland, which is because the library explicitly refuses to, in turn because Wayland does not have the same window manager concept as X11.

quozl:
And this makes me suddenly think; why are you trying to test issue 952 under Wayland?  :grin:  I've edited the issue 952 opening comment to make it clear that I was reporting the problem happens under X11.

bhavyabansal9068:
Thank you very much, @quozl  and @chimosky , for clarifying and guiding me well 😄..
I realize that I misunderstood some things. I will be creating a new issue for it and raising an updated PR under that.

quozl:
@bhavyabansal9068 please tell me if you have reviewed https://github.com/sugarlabs/sugar/pull/951?

bhavyabansal9068:
@quozl  I hadn't reviewed before PR but those tasks seems quite interesting would definitely explore them..
Thank you

quozl:
Oh, good.  I wondered why all this seemed familiar; I had started on it last year.  ;-)

bhavyabansal9068:
Yeah @quozl  i will continue on it and will raise new PR..

Issue #856, opened 2019-08-14
Text clipping cannot be removed from Frame


quozl:
Reproducer: start Terminal activity, select an area of text, use ctrl+shift+c shortcut to copy text, open the Frame, right click on the clipping, select Remove.

Expected result: text clipping is removed.

Observed result: text clipping remains.

shell.log contains;

```log
1565770157.805620 WARNING root: no data for selection target MULTIPLE.
Traceback (most recent call last):
  File "/usr/lib/python3.7/dist-packages/jarabe/frame/clipboardtray.py", line 117, in _object_added_cb
    objects_to_delete = self.get_children()[:-self.MAX_ITEMS]
TypeError: slice indices must be integers or None or have an __index__ method
Traceback (most recent call last):
  File "/usr/lib/python3.7/dist-packages/jarabe/frame/clipboardicon.py", line 155, in _object_state_changed_cb
    if self._current_percent < 100 and cb_object.get_percent() == 100:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
Traceback (most recent call last):
  File "/usr/lib/python3.7/dist-packages/jarabe/frame/clipboardicon.py", line 155, in _object_state_changed_cb
    if self._current_percent < 100 and cb_object.get_percent() == 100:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
1565770157.814280 WARNING root: no data for selection target NULL.
```

quozl:
Fixing the above tracebacks;

```patch
From 5f21a0772492be2f8cc138c2fd1b82768008bd91 Mon Sep 17 00:00:00 2001
From: James Cameron <quozl@laptop.org>
Date: Wed, 4 Sep 2019 16:29:03 +1000
Subject: [PATCH] Work in progress - Fix text clipping cannot be removed from Frame

---
 src/jarabe/frame/clipboardicon.py | 7 ++++---
 src/jarabe/frame/clipboardtray.py | 2 +-
 2 files changed, 5 insertions(+), 4 deletions(-)

diff --git a/src/jarabe/frame/clipboardicon.py b/src/jarabe/frame/clipboardicon.py
index 6a4c3275e..dfdc6d419 100644
--- a/src/jarabe/frame/clipboardicon.py
+++ b/src/jarabe/frame/clipboardicon.py
@@ -48,7 +48,7 @@ class ClipboardIcon(RadioToolButton):
         self.props.sensitive = False
         self.props.active = False
         self._notif_icon = None
[...]

PR #966, opened 2023-03-26
Changed the datatype for the parameter of the emit function to int64 during object deletion signal.

bhavyabansal9068:
Fix for issue #856
I have tested it locally and its working fine.


quozl:
Thanks, but please remove the unrelated commits, and for bd42aeb rewrite the commit message to explain the problem and how it is fixed.  Given the information available in the commit message right now, I was unable to comprehend the reason for the change.

bhavyabansal9068:
Sure @quozl

PR #971, opened 2023-08-25
Update live build link to recent image

chimosky:
@quozl kindly review.

### DIFF
--- a/src/jarabe/journal/model.py
+++ b/src/jarabe/journal/model.py
@@ -941,6 +941,7 @@ def get_documents_path():
         return _documents_path
 
     try:
+        subprocess.run(['xdg-user-dirs-update'])
         pipe = subprocess.Popen(['xdg-user-dir', 'DOCUMENTS'],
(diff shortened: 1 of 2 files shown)


quozl:
The link is updated fine.  Looking randomly at the packages list, rsync is from Debian 10 (Buster), now marked as oldoldstable in packages.debian.org, but still supported according to a quick glance at Wikipedia's Debian version history page.  I'm not sure how useful a live build would be for that vintage.  Debian 12 (Bookworm) would be a more appropriate target release.

chimosky:
> The link is updated fine. Looking randomly at the packages list, rsync is from Debian 10 (Buster), now marked as oldoldstable in packages.debian.org, but still supported according to a quick glance at Wikipedia's Debian version history page. I'm not sure how useful a live build would be for that vintage. Debian 12 (Bookworm) would be a more appropriate target release.

Agreed, I'll take care of that and upload a new image.

PR #979, opened 2024-02-18
Introducing continuous integration...

zthurman:
which builds sugar on both fedora and ubuntu.
There is also a tiny patch to po/LINGUAS to
address a file that was removed upstream in
[ca6c6d2](https://github.com/sugarlabs/sugar/commit/ca6c6d27c0c323ee0e7f1ac4e8d4cf900afb110a).


chimosky:
Looked at your changes, it doesn't run any tests, it only builds Sugar and we have tests that are part of sugar, also we don't really need CI/CD as it's just a maintenance overhead as not a lot of things happen in this repo.

Thanks for the changes though, looking forward to more contributions.

PR #984, opened 2024-04-04
Add weblate sync workflow

chimosky:
This will automatically open a PR whenever weblate pushes changes to weblate-sync.

### DIFF
--- a/.github/workflows/weblate-sync.yaml
+++ b/.github/workflows/weblate-sync.yaml
@@ -1,0 +1,14 @@
+name: Weblate - Create PR when weblate pushes changes
+on:
+  push:
+    branches:
+      - weblate-sync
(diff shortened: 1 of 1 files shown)


quozl:
File name has syc; does it need it?

chimosky:
Thought I'd fixed it earlier, thanks for pointing it out.

chimosky:
@quozl you're find with merging this right?

quozl:
I've no real objection.  I don't like the idea of more notifications, when the step involved doesn't need to happen until release it could just as easily be part of release process, and for the previous releases that's how I handled it.

chimosky:
Seeing as this removes the need to manually do it at release, I think it's a good thing.

PR #986, opened 2024-05-31
translate to malay

joshuatanxx:
I have updated some of the translations for SUGAR


chimosky:
Thanks for the changes, I'll merge this but for future translations you can use [weblate](translate.sugarlabs.org).

PR #988, opened 2024-07-29
Fix incomplete port to python3

chimosky:
Call xdg-user-dirs-update to update xdg directories before looking for any user directory.

@kshitijdshah99 kindly test and review, you can test with Pippy activity, you'll need to build sugar
with this branch to test.

Before this change `ActivityToolbarButton -> Create a Sugar activity Bundle` threw an error, this fixes that.

### DIFF
--- a/src/jarabe/journal/model.py
+++ b/src/jarabe/journal/model.py
@@ -941,6 +941,7 @@ def get_documents_path():
         return _documents_path
 
     try:
+        subprocess.run(['xdg-user-dirs-update'])
         pipe = subprocess.Popen(['xdg-user-dir', 'DOCUMENTS'],
                                 stdout=subprocess.PIPE)
         documents_path = os.path.normpath(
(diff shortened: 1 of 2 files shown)


quozl:
Reviewed.

chimosky:
> Reviewed.

Thank you!

kshitijdshah99:
To confirm that I have perfectly understood this issue
1. Firstly I have to build live Sugar
2. Modify the files which you have changed
3. and then test it
Have I addressed the issue correctly?

chimosky:
> To confirm that I have perfectly understood this issue
> 
>     1. Firstly I have to build live Sugar
> 
>     2. Modify the files which you have changed
> 
>     3. and then test it
>        Have I addressed the issue correctly?

The first step is correct, you don't have to modify the files which I've changed, you can just pull my changes, and build sugar with my changes, then restart Sugar.

[L38-L44](https://github.com/sugarlabs/sugar-live-build/blob/master/src/config/hooks/normal/0900-sugar.hook.chroot#L38) shows you how to build Sugar, you'll have to move the files to your machine directories that exist and is part of `sys.path`.

kshitijdshah99:
Ok I think now I get the point, I'll try with this and let you know if any issues👍

PR #989, opened 2024-08-30
Translated using Weblate (Georgian)

chimosky:
Weblate synchronisation.


quozl:
Just push to master, no need for a pull request.  Waste of notifications.

chimosky:
Agreed, I'll change it to just push to master.

PR #994, opened 2025-02-09
Update Makefile.am

bhavesh572:
✅ Performance Boost:

Parallel execution support using .NOTPARALLEL: where necessary. Dynamic wildcard expansion to reduce hardcoded lists, allowing automation and reducing maintenance overhead. ✅ Better Maintainability:

Separates common variables like PYTHON, MIME_DIR, and BUILD_DIR. Uses wildcards instead of manually listing files.
✅ Modernization:

Reduces reliance on outdated GTK resource generation; can be expanded for GTK 4. Uses dynamic MIME updates instead of hardcoded paths. ✅ Better Error Handling & Logging:

Echos logs when updating MIME.
Uses [ -z "$$DESTDIR" ] && ... || true for error resilience during install/uninstall hooks.

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -1,64 +1,55 @@
 SUBDIRS = icons
 
-sugar-72.gtkrc: gtkrc.em
-	$(PYTHON) -m em -D scaling=\'72\' $(srcdir)/gtkrc.em > \
-		$(top_builddir)/data/sugar-72.gtkrc
+# Define common variables for maintainability
+PYTHON ?= python3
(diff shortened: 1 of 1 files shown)


chimosky:
The changes you've made look like they were gotten from an AI, and poorly thought out, also unreadable for some of the tasks.

Why do we need to change the current makefile?
Why do we need parallel execution support?

Your commit message does very little to answer these questions.

@quozl thoughts?

quozl:
Agreed.

PR #995, opened 2025-03-04
floss fund estabhlishment

MostlyKIGuess:
In order for this to validate our schema, we would need to have this structure .
https://github.com/sugarlabs/www/pull/679
As present in the guidelines:
https://floss.fund/funding-manifest/


MostlyKIGuess:
@pikurasa

Issue #1001, opened 2025-05-18
Outdated installation command for speak on fedora


mebinthattil:
In the website : https://wiki.sugarlabs.org/go/Installation ; under the **fedora** section, the following code is given as part of installation steps: 
```
sudo dnf groupinstall sugar-desktop
```
But this throws an error. `groupinstall` seems to be an alias for `group install`.
But the more modern dnf versions don’t seem to know this alias and just throw an error saying it's not recognized.

So the corrected installation command should look like this:
```
sudo dnf group install sugar-desktop
```

---
P.S: I'm not sure where to go to be able to edit the contents of wiki.sugarlabs.org, if someone could tell me how to edit wiki pages I could go ahead and replace it myself as well. Thanks

chimosky:
Closing this as there's a PR to address this, in the future please just open a PR if you intend to fix an issue.

Opening both an issue and a PR is just noise for us.

mebinthattil:
I created the issue, since there is still one part which is not yet fixed - updating the wiki site.

quozl:
Our advice to use dnf was us being nice to Fedora users, as an
implementation of advocacy.

Fedora also have a duty to explain how to use packages.

When we explain it, it could be in our repository, on our web site, or
on our Wiki.  I don't mind either way, but as you've seen, the wider
the distribution of advice the greater the probability of the
information aging, becoming bad, and allowing a negative impression to
form.

mebinthattil:
We can only improve what’s in our control. 

Considering this is the first step to experiencing and downloading sugar, a roadblock here can be a big turn off. I'm looking at this from the perspective of a person who is not very tech-savvy and is trying to install sugar DE. 
The fix #1002 is fairly simple, but it’s only for the GH docs and not for the wiki. I suggest we update and merge it, since it's also backwards compatible with older versions of dnf.

Sure Fedora could have done a better job, but maybe we can fix and make the experience better for our users.

Let me know your thoughts. Thanks!

PR #1002, opened 2025-05-18
Update deprecated `groupinstall` command for sugar fedora installation

mebinthattil:
For newer version of dnf (dnf 5+) the `groupinstall` command no longer works. This is because the `groupinstall` was actually an alias to `group install` which was last supported until dnf4. So for newer versions of fedora, running the old command would throw an error saying `groupinstall` was not recognized as a command.

I have updated the following:
- `README.md`
- `docs/fedora.md`
---
This PR partially solves the issue #1001, since the wiki is still yet to be updated.

### DIFF
--- a/docs/fedora.md
+++ b/docs/fedora.md
@@ -10,7 +10,7 @@ Using Sugar as a Desktop Environment
 
 Install Fedora. Then, in a Terminal, type:
 
-    sudo dnf groupinstall sugar-desktop
+    sudo dnf group install sugar-desktop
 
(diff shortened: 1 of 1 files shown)


chimosky:
Could you preserve the existing command and clearly specify what version of dnf it works for, and what version the new command is for.

mebinthattil:
I don't see the need for preserving the old version -`groupinstall`. 

In the older versions of dnf, the `groupinstall` was just an alias to `group install` which means that using `group install` should work just fine for both the older and newer versions of dnf.

But if it you still want me to segregate them, let me know. I'll split it up. 
Do let me know if anything else needs to be done, thanks!

chimosky:
I've never had to use that so I had no idea it was an alias, thanks for the change.

mebinthattil:
Kindly ensure that the wiki is updated too, thanks

PR #1011, opened 2025-12-18
Fix Alt-click handling on Home view

### LINKED ISSUE #969: Alt-click not working on home view
Pressing <kbd>alt</kbd> correctly removes colour from home view icons, but while the key is held down clicking on the icon doesn't work.  It is supposed to do a _Start new_ action.

Sugar 0.118 on Ubuntu 20.04

nishtha-agarwal-211:
Fix Alt-click handling on Home view.

Alt + left click is now handled explicitly on BUTTON_PRESS events
and shows the activity palette immediately. Normal click behavior
remains unchanged.

Fixes #969

### DIFF
--- a/src/jarabe/desktop/favoritesview.py
+++ b/src/jarabe/desktop/favoritesview.py
@@ -214,11 +214,15 @@ class FavoritesView(ViewContainer):
         return False
 
     def __button_press_cb(self, widget, event):
-        if event.button == 1 and event.type == Gdk.EventType.BUTTON_PRESS:
+        if event.type != Gdk.EventType.BUTTON_PRESS:
+            return False
+
+        if event.button == 1 and event.state & Gdk.ModifierType.MOD1_MASK:
             self._last_clicked_icon = widget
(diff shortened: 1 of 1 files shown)


chimosky:
The issue says **"It is supposed to do a _Start new_ action"** and from your opening comment and change, it doesn't look like that's what it does.

Also, please write better commit messages, your opening comment should've been part of your commit message, see [making commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).

nishtha-agarwal-211:
Thanks for the review.

You’re right — the issue refers to Alt-click being used to “Start new”.
My intention with this change was to restore proper Alt-click handling
by ensuring the palette appears immediately, since “Start new” is
exposed through the activity palette.

If the expected behavior is to trigger “Start new” directly without
showing the palette, please let me know and I’ll be happy to adjust
the implementation accordingly.

Regarding the commit message — agreed. I will amend it to follow the
project’s commit message guidelines.

Issue #1017, opened 2026-01-13
Warn users if existing DSA key is unsupported in OpenSSH ≥10


vanshjohri09-collab:
----------Problem
- Existing Sugar profiles may have DSA keys.
- OpenSSH ≥10 removed support for DSA keys.
- Sugar currently assumes "key present = valid", without checking key type.
- As a result, collaboration features (Chat, shared activities) fail silently on modern systems.

--------Proposed Solution
- Detect if the existing key is DSA.
- Check OpenSSH version (≥10).
- If DSA and unsupported, show a GTK3 warning to the user:
  > "Your SSH key type (DSA) is unsupported by your system. Please generate RSA or ED25519 key."
- Do not modify or delete existing keys.
- Maintain backward compatibility with older systems.

-------References
- Related old issue: #1004
- Key check in code: `src/jarabe/intro/window.py`, `get_pubkey()`
- Observation confirmed by searching repo and testing key presence logic.

----- Additional Notes
- This is mainly a **user-awareness improvement**; no activity or collaboration code changes needed.
- Provides a minimal but effective fix, ensures users know why their collaboration fails.

PR #1018, opened 2026-01-13
Draft: Warn users when DSA SSH key is detected

vanshjohri09-collab:
Draft PR for issue #1017

This PR demonstrates the detection logic and user-facing warning message
for unsupported DSA SSH keys on systems using OpenSSH ≥10.

Notes:
- This is a draft to validate approach and flow.
- GTK warning integration is not finalized yet due to environment limits.
- No keys are modified or deleted.
- Backward compatibility is preserved.

Requesting feedback before wiring this into:
src/jarabe/intro/window.py

### DIFF
--- a/src/test_dsa_warning.py
+++ b/src/test_dsa_warning.py
@@ -1,0 +1,10 @@
+def test_dsa_warning():
+    key_type = "ssh-dss"
+    if key_type == "ssh-dss":
+        print(
+            "WARNING: Your SSH key type (DSA) is unsupported. "
+            "Please generate RSA or ED25519 key."
+        )
+
+if __name__ == "__main__":
+    test_dsa_warning()


chimosky:
> Thanks for the clarification, that makes sense.
> 
> You're right, this draft only demonstrated the detection logic and does not use Sugar's alert system, so it wouldn't properly notify users.
> 
> I'll rework this to use Sugar alerts (e.g. NotifyAlert) and wire it directly into src/jarabe/intro/window.py, then open a new PR with the correct UX.

I'm saying alerting the user isn't a good way to handle this, because the user isn't involved in the process that led to the error. It makes sense to user alerts for user triggered actions, this isn't triggered by the user as it happens without their knowledge.

PR #1019, opened 2026-01-21
Gtk4 port

chimosky:
I'm pushing because I want to take a break from looking at this, anyone is welcome to build on this.

I've also used [casilda](https://gitlab.gnome.org/jpu/casilda/-/tree/main?ref_type=heads) as a wayland compositor, wanted to test it with Sugar as it's a lightweight compositor we can use.

If you intend to continue this, you should have the [sugar-live-build](https://github.com/sugarlabs/sugar/blob/master/docs/development-environment.md#sugar-live-build) setup as that's how you can test changes to Sugar and its core APIs.

[Sugar Ext](https://github.com/sugarlabs/sugar-ext) is also needed to test this.


vyagh:
Hey @chimosky, I'd like to help continue this port
I have noticed there are still GTK3 patterns remaining, like `Gtk.VBox`/`Gtk.HBox` and `Gdk.Screen` usages.
I've started converting expandedentry.py locally (the Gtk.VBox/Gtk.HBox -> Gtk.Box conversions)

Should I submit this as a commit to your branch? or would you prefer I open a separate PR that you can select from?
Let me know the best way to contribute to this

D-I-R-M:
Hey @chimosky 
i have set up the live build 
can you tell how can i install sugar toolkit gtk4 in the live build and how can i make changes to the file?
Debian 12 x - VMware Workstation 17 Player (Non-commercial use only) 3_2_2026 3_35_19 PM

chimosky:
[sugar-toolkit-gtk4](https://github.com/sugarlabs/sugar-toolkit-gtk4) contains information on how to install it, for future reference, it'll be great if you shared text as text and not a screenshot.

I don't know what you're trying to achieve from the screenshot, but `sugar-activity3` wasn't meant to be used directly but can be used that way, you'll have to be in an activity directory to get it to work as expected.

You can make changes to the file using a text editor, seeing as it's a vm, you can use your preferred editor and share your changes to the vm via ssh, or you can use vim in the vm, that just depends on your preference.

D-I-R-M:
Ok Thanks https://github.com/sugarlabs/sugar-toolkit-gtk4 but
I was trying to install sugar toolkit gtk4 so after the second step when i try to do pip install -e
-e option requires an argument shows up
and 
Do i need to install sugar-toolkit-gtk4 ? in order to contribute in this migration or do i need to work on the sugar toolkit 4 https://github.com/sugarlabs/sugar-toolkit-gtk4/issues


First i set up the live build then i was installing sugar-toolkit-gtk4 (above error occurs) then as mentioned by you to run build commands for the core sugar API and the git -pull unshallow  (done) 
Cannot understand what you were referring through casilda readme

How can you and i know if i am even working in the correct direction?(that is why i thought screenshot might help me to communicate to you)

chimosky:
> Do i need to install sugar-toolkit-gtk4 ? in order to contribute in this migration or do i need to work on the sugar toolkit 4 https://github.com/sugarlabs/sugar-toolkit-gtk4/issues

Yes, this port uses it.

> First i set up the live build then i was installing sugar-toolkit-gtk4 (above error occurs) then as mentioned by you to run build commands for the core sugar API and the git -pull unshallow (done)
Cannot understand what you were referring through casilda readme

I don't know the error you're referring to, it's difficult to tell which is an error from your message.

Casilda is a lightweight Wayland compositor, this change uses it too.

> How can you and i know if i am even working in the correct direction?(that is why i thought screenshot might help me to communicate to you)

The first step would be to install the dependencies, most are in my opening comment.
You can build and go from there, fix build issues as they come up, and continue the port to Gtk4 as that's the point of this PR.

Seeing as you're trying to understand why some of these components exist, I'll suggest you take on issues with lesser difficulty instead of this one, when you understand some things then this would be easier to tackle.

I'd love to help you, but I can't hold your hand, guide you and show you exactly what needs to be done.

Dakshsharma1108:
Hey @chimosky, can ```GObject.GObject.__init__(self)``` be replaced with ```super().__init__()``` ?

quozl:
> Hey @chimosky, can `GObject.GObject.__init__(self)` be replaced with `super().__init__()` ?

We no longer need to retain Python 2 compatibility, and this is a Python 3 language optimisation.

We may have bugs or unintended features reliant on the older code.

As long as you have a test for the changed code, I'm unconcerned.

Dakshsharma1108:
Hi @chimosky , I had a quick question — can we replace direct calls like ```Gtk.Box.__init__(self, orientation=Gtk.Orientation.VERTICAL)``` (or ```Gtk.EventBox.__init__```) with ```super().__init__(...)``` in GTK4? Just wanted to confirm the preferred approach.

chimosky:
> Hi @chimosky , I had a quick question — can we replace direct calls like `Gtk.Box.__init__(self, orientation=Gtk.Orientation.VERTICAL)` (or `Gtk.EventBox.__init__`) with `super().__init__(...)` in GTK4? Just wanted to confirm the preferred approach.

I assume you're talking about replacing it with `super().__init__` calls, I agree with that.

Dakshsharma1108:
Great, thanks for confirming! I’ll proceed with replacing those direct ```Gtk.*.__init__``` calls with ```super().__init__(...)```

Dakshsharma1108:
Hi @chimosky , I had a quick question about this commit :- https://github.com/sugarlabs/sugar/commit/6205ab81ddd6b39e8bfee0b2f1eedf6e63179bc8

Since ```Gtk.ToolItem``` was replaced with ```Gtk.Box```, should we follow the same approach for similar widgets like ```ToolbarBox```and ```ToolButton```, or are there preferred GTK4 replacements for those?

chimosky:
> Since `Gtk.ToolItem` was replaced with `Gtk.Box`, should we follow the same approach for similar widgets like `ToolbarBox` and `ToolButton`, or are there preferred GTK4 replacements for those?

I imagine that should be the case, and you'll also have to use your intuition.

You can look at how Sugar currently looks and try to replicate that when porting, you should also look at sugar-toolkit-gtk4 as some of this has probably been replicated there too.

Dakshsharma1108:
> 
> I imagine that should be the case, and you'll also have to use your intuition.
> 
> You can look at how Sugar currently looks and try to replicate that when porting, you should also look at sugar-toolkit-gtk4 as some of this has probably been replicated there too.

Got it, that makes sense. I checked sugar-toolkit-gtk4 and saw that ToolbarBox is built using Gtk.Box while ToolButton is reimplemented using Gtk.Button. I’ll follow a similar pattern and the current UI while porting. Thanks!

PR #1030, opened 2026-02-03
Handle datastore restart after DBus disconnect

### LINKED ISSUE #870: Feature - handle datastore restart
When the datastore process is terminated by hand or through error, the Sugar session does not recover, and the datastore is not restarted.  This may be because Sugar does not;

* detect the consequent error on next transaction; _dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name :1.20 was not provided by any .service files_,
* open a new interface over D-Bus.

Dev10-sys:
What breaks
When the sugar-datastore process is terminated or crashes, Sugar keeps a
stale D-Bus proxy. Subsequent datastore operations fail with
org.freedesktop.DBus.Error.ServiceUnknown and the Sugar session does not
recover.

Why it breaks
The datastore D-Bus interface is cached and not recreated when the service
disappears and is later reactivated via D-Bus activation.

How recovery works
Datastore D-Bus calls are wrapped to catch ServiceUnknown, NoReply, and
Disconnected errors. On such failure the cached datastore proxy is cleared,
a fresh D-Bus interface is created, and the operation is retried once. If
the retry also fails, the original exception is raised.

What is intentionally NOT handled
This change does not add UI notifications, signal-based monitoring, or
infinite retries. Recovery is silent and limited to a single retry to avoid
activation loops and unintended behavior changes.

Fixes #870

### DIFF
--- a/src/jarabe/journal/model.py
+++ b/src/jarabe/journal/model.py
@@ -222,8 +222,8 @@ class DatastoreResultSet(BaseResultSet):
         BaseResultSet.__init__(self, query, page_size)
 
     def find(self, query):
-        entries, total_count = _get_datastore().find(query, PROPERTIES,
-                                                     byte_arrays=True)
+        entries, total_count = _call_datastore('find', query, PROPERTIES,
+                                               byte_arrays=True)
 
         for entry in entries:
             entry['mountpoint'] = '/'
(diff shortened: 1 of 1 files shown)


chimosky:
I haven't tested this yet, have you tested this? You haven't mentioned testing.

Your opening comment should be your commit message, see [making commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).

PR #1032, opened 2026-02-05
frame: fix icon redraw when opening frame palettes

### LINKED ISSUE #819: Frame icons vanish when frame palettes open
Reproducer: open the frame, hover mouse or touch the battery or speaker icon.

Expected result: palette for icon is drawn, no other changes.

Observed result: palette for icon is drawn, but other icons may vanish.  Is most likely on first use of palette for an icon; repeated uses don't cause other icons to vanish.

Occurs on Ubuntu 18.04

Dev10-sys:
Fixes #819

When opening palettes for frame icons, adjacent icons could vanish due to
missing redraws after pointer enter/leave events.

The frame hover state was updated, but no redraw was triggered, leaving
parts of the frame invalidated after palette windows overlapped it.

This change queues a redraw on hover enter and leave, ensuring the frame
is fully repainted and icons remain visible.

Manual testing:
- Open frame
- Hover over frame icons to open palettes
- Move between adjacent icons
- Verified icons no longer disappear

### DIFF
--- a/src/jarabe/frame/framewindow.py
+++ b/src/jarabe/frame/framewindow.py
@@ -148,10 +148,12 @@ class FrameWindow(Gtk.Window):
     def _enter_notify_cb(self, window, event):
         if event.detail != Gdk.NotifyType.INFERIOR:
             self.hover = True
+            self.queue_draw()
 
     def _leave_notify_cb(self, window, event):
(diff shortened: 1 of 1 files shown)


chimosky:
Tested, issue still persists.

Nothing in the logs, did you test this? If yes, how'd you test?

PR #1036, opened 2026-02-06
Fix paste failure when multiple clipboard items exist

### LINKED ISSUE #783: Paste fails when frame contains multiple copies
Reproducer:

* start Sugar 0.112,
* start Terminal,
* display some text,
* select and copy some text using `ctrl+shift+c`; the frame will show a text clipping,
* select and copy some other text using `ctrl+shift+c`; the frame will show two text clippings,
* paste the text using `ctrl+shift+v`,

Expected result:

* most recent text clipping is pasted,

Observed result:

* nothing happens.

Logs: nothing unusual in Terminal log, but shell.log contains these two lines for each `ctrl+shift+c` key event,

```
1517518078.319098 WARNING root: no data for selection target MULTIPLE.
1517518078.328716 WARNING root: no data for selection target NULL.
```

Reported by @Menzador.

Dev10-sys:
This change fixes a paste failure that occurs when multiple clipboard
objects exist in the frame.

The drag-and-drop handler was attempting to process unsupported selection
targets, which caused paste to fail. These targets are now skipped so
valid clipboard data is handled correctly.

Reproducer:
- Copy multiple items so several clipboard entries exist
- Attempt to paste
- Paste now correctly inserts the most recent clipboard item

Fixes #783

### DIFF
--- a/src/jarabe/frame/clipboardtray.py
+++ b/src/jarabe/frame/clipboardtray.py
@@ -163,8 +163,9 @@ class ClipboardTray(tray.VTray):
         self._context_map.add_context(context, object_id, len(context_targets))
 
         for target in context_targets:
-            if str(target) not in ('TIMESTAMP', 'TARGETS', 'MULTIPLE'):
-                widget.drag_get_data(context, target, time)
+            if str(target) in ('TIMESTAMP', 'TARGETS'):
+                continue
+            widget.drag_get_data(context, target, time)
 
         cb_service.set_object_percent(object_id, percent=100)
 


chimosky:
Tested, this shows in the log;

```python3
  File "/usr/lib/python3/dist-packages/jarabe/frame/clipboardicon.py", line 125, in _clipboard_data_get_cb
    selection.set(selection_target, 8, data)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Must be number, not str
```

Paste doesn't happen as expected, I copied journal objects.

PR #1047, opened 2026-02-10
neighborhood: guard against None active activity in buddy update

SaaiAravindhRaja:
`_update_buddy_activities` calls `get_activity_id()` on the result of `get_active_activity()` without a None check. `get_active_activity()` returns `None` when no activity is active (e.g. on the home screen).

When a Telepathy activity callback fires with no active activity, this crashes with `AttributeError: 'NoneType' object has no attribute 'get_activity_id'`.

Added a None guard before the method call.


quozl:
How is this reproduced?

PR #1051, opened 2026-02-19
journal: Remove obsolete FAT hidden attribute logic

Akshay-krish:
This PR removes the legacy logic in the Journal model that attempted to set FAT hidden attributes on metadata directories.

Analysis:
As noted in the source code comments, this was kept for "backward compatibility." However, on modern systems, this call is redundant and can lead to unnecessary error logging. Removing this call allows for the complete removal of the sugar-fatattr infrastructure in the toolkit.

Changes:

Removed the SugarExt.fat_set_hidden_attrib call block in src/jarabe/journal/model.py.
Removed the SugarExt import from gi.repository as it is no longer required in this module.
Cleaned up the associated error logging block.

Dependencies:
Note to Mentors: This PR should only be merged in coordination with the toolkit changes here: 
[f5c5398](https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/504)

### DIFF
--- a/src/jarabe/journal/model.py
+++ b/src/jarabe/journal/model.py
@@ -30,8 +30,6 @@ import dbus
 from gi.repository import Gio
 from gi.repository import GLib
 
-from gi.repository import SugarExt
-
 from sugar3 import dispatch
 from sugar3 import mime
 from sugar3 import util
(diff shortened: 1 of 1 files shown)


quozl:
Not obsolete.

PR #1067, opened 2026-02-26
Enable NumLock at startup and support KP_Enter input

### LINKED ISSUE #1053: Sugar starts in wrong numlock mode
Problem: Sugar home view search entry and the Calculate activity do not respond to keypad.

Not a problem with Calculate.  Sugar does not support a keyboard with numlock, because the key was removed for children's laptops.  However, on a system with a numlock key, and a numeric keypad, Sugar starts the wrong way, and the numeric keypad keys are ineffective.

Aniketsy:
fixes #1053


Aniketsy:
I’ve run the tests locally and they’re passing. please let me know if there’s another recommended way to validate this fix in practice. I’d be happy to test further if needed.

PR #1068, opened 2026-03-01
Fix: Journal list updates breaking when datastore write fails

Ady0333:
## Summary

Fix an issue where the Journal list view could stop receiving datastore update notifications if a datastore write operation failed.

---

## Problem

In `src/jarabe/journal/model.py`, the `error_handler` used by `write()` logged a static error message and returned without invoking `ready_callback`.

However, callers such as `ListModel.set_value()` temporarily disconnect the `model.updated` signal before calling `write()` and rely on `ready_callback` to reconnect it after the operation completes.

If the datastore write fails, `ready_callback` was never called, leaving the signal handler disconnected. As a result, the Journal list view would stop updating for the remainder of the session — new entries, deletions, or metadata changes would not appear until Sugar was restarted.

---

## Fix

Two small changes were made in `write()`:

1. Include the actual error object in the log message so datastore failures are diagnosable:

   ```
   logging.error('Could not create/update datastore entry: %s', error)
   ```

2. Ensure `ready_callback` is invoked even when the write operation fails, allowing callers to restore their state (e.g., reconnect the `model.updated` signal).

---

## Impact

* Prevents the Journal list view from becoming disconnected from datastore updates after a write failure.
* Ensures callers relying on `ready_callback` always restore their state.
* Improves debugging by logging the underlying datastore error.

These changes keep the Journal UI responsive even when datastore operations fail due to conditions such as disk pressure or temporary D-Bus errors.

### DIFF
--- a/src/jarabe/journal/model.py
+++ b/src/jarabe/journal/model.py
@@ -728,7 +728,9 @@ def write(metadata, file_path='', update_mtime=True, transfer_ownership=True,
             ready_callback(metadata, file_path, metadata['uid'])
 
     def error_handler(error):
-        logging.error('Could not create/update datastore entry')
+        logging.error('Could not create/update datastore entry: %s', error)
+        if ready_callback:
+            ready_callback(metadata, file_path, metadata.get('uid', ''))
 
     logging.debug('model.write %r %r %r', metadata.get('uid', ''), file_path,
                   update_mtime)


Ady0333:
@walterbender 
Please review this pr.

quozl:
Has this been observed to occur?

Ady0333:
Hello @quozl thankyou for looking into this PR.

Yes, I came across it while i was tracing the Journal update flow in the codebase. The path from `ListModel.set_value()` to `model.write()` caught my attention because the signal is disconnected before the async write and depends on `ready_callback` to restore it.

quozl:
No, I think you may have misunderstood my question; has this been observed to occur in a running instance of Sugar?  If so, please provide the datastore logs.

Ady0333:
> No, I think you may have misunderstood my question; has this been observed to occur in a running instance of Sugar? If so, please provide the datastore logs.

I haven’t personally encountered this during normal use of Sugar yet. I noticed it while tracing the Journal update flow in the codebase. `ListModel.set_value()` disconnects the `model.updated` signal and relies on `ready_callback` from `write()` to reconnect it. In `write()`, the success handlers invoke the callback, but the error handler only logs and returns, so if the datastore call fails the reconnect step would never run. The change simply ensures the callback runs in both paths so the model state is restored.

quozl:
How would the datastore call fail?  Why should we pay attention to this scenario?

Ady0333:
The datastore call goes through D-Bus in `model.write()`, so it depends on the availability of the datastore service and the underlying storage. In practice the call can fail if the datastore service restarts, is temporarily unavailable, or if the operation fails due to storage conditions (for example disk pressure or an I/O error). In those cases the async request returns through the error handler.

In `ListModel.set_value()` the `model.updated` signal is explicitly disconnected before calling `model.write()` and the reconnect happens inside the `ready_callback` (`__reconnect_updates_cb`). The success handlers in `write()` invoke this callback, but the error handler only logs and returns. If the datastore call fails, the callback that restores the signal connection is never executed.

The change ensures that the callback is invoked in both success and failure paths so callers that temporarily modify their state around the async call can reliably restore it.

PR #1069, opened 2026-03-01
Port: Replace activityfactory with standard uuid and subprocess

Inuth0603:
Removes dependency on `sugar4.activity.activityfactory` by replacing:

* `activityfactory.create_activity_id()` → `str(uuid.uuid4())`
* `activityfactory.create(bundle, handle)` → new `launch_activity()` function using `subprocess.Popen`

`sugar-launch` still uses `activityfactory.get_command()` and `activityfactory.get_environment()` which are outside the scope of this change.

### DIFF
--- a/src/jarabe/desktop/activitychooser.py
+++ b/src/jarabe/desktop/activitychooser.py
@@ -15,6 +15,7 @@
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 import logging
+import uuid
 
 from gi.repository import GLib
 from gettext import gettext as _
@@ -22,7 +23,6 @@ from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import GObject
 
-from sugar4.activity import activityfactory
 from sugar4.graphics import iconentry
 from sugar4.graphics import style
 from sugar4.graphics.toolbutton import ToolButton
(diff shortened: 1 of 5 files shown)


chimosky:
activityfactory isn't deprecated, you're trying to backport a change and I don't see the reason why.

Your commit message also doesn't say why.

Inuth0603:
> activityfactory isn't deprecated, you're trying to backport a change and I don't see the reason why.
> 
> Your commit message also doesn't say why.

`activityfactory` wasn't ported to sugar-toolkit-gtk4 — it doesn't exist in the `sugar4.activity` package. Importing it raises `ImportError`. This PR replaces it with `uuid` and `subprocess` so the gtk4-port branch can actually run.

Updated the PR title and commit message to remove "deprecated"    you're right, it's not deprecated, it was just never ported to gtk4.

chimosky:
> activityfactory wasn't ported to sugar-toolkit-gtk4 — it doesn't exist in the sugar4.activity package. Importing it raises ImportError. This PR replaces it with uuid and subprocess so the gtk4-port branch can actually run.

There's a reason why it was never ported, @MostlyKIGuess can talk more about why but it will be added.

I still don't see the need to change anything here, you haven't talked about why the change is necessary.

Inuth0603:
Got it, wasn't aware it's planned to be ported. Happy to close this if activityfactory will be added to sugar-toolkit-gtk4.

PR #1070, opened 2026-03-02
Fix updator refresh timing

Ravi-Poddar26:
Tested on Ubuntu 24.04 (X11) in a VM.

It does resolve the Software Update UI issue. The section still requires multiple clicks to open, although no traceback is shown.

The problem appears to be caused by _refresh() executing during initialization and interfering with GTK view selection.

Deferring the call using:

    GLib.idle_add(self._refresh)

resolves the issue. Software Update now opens with a single click and no UI freeze.

issue #1055 

PR submitted with this fix.

### DIFF
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -5,6 +5,17 @@ Using Sugar on Ubuntu
 
 In relation to Sugar, Ubuntu is a downstream distribution project that can be used to run Sugar.
 
+Ubuntu 24.04 (Noble Numbat)
+-------------------
+
+Sugar 0.121 can be installed with the following commands:
+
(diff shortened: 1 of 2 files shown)


chimosky:
Reviewed, not tested.

Your PR contains a commit that has nothing to do with the title.

Ravi-Poddar26 (commented):
Thank you for the review.
I try to remove the unrelated documentation commit and cleaned the PR and  contains only the updater refresh timing fix as soon as possible.

Ravi-Poddar26:
I have created a new PR that contains **only the refresh timing fix and related changes**. Could you please review that one? I would appreciate your feedback.

PR #1075, opened 2026-03-04
Fix updator refresh clean

Ravi-Poddar26:
## Summary

This PR fixes the Software Update section requiring multiple clicks to open on Ubuntu 24.04 (X11).

The issue was caused by `_refresh()` being executed during initialization, which interfered with GTK view selection and caused a temporary UI freeze.

## Fix

The `_refresh()` call is deferred using:

GLib.idle_add(self._refresh)

This ensures the refresh runs after the UI is fully initialized.

## Testing

Tested on Ubuntu 24.04 (X11) in a VM.

Results:
- Software Update now opens with a single click
- No UI freeze
- No traceback in logs

Fixes #1055

### DIFF
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -5,6 +5,17 @@ Using Sugar on Ubuntu
 
 In relation to Sugar, Ubuntu is a downstream distribution project that can be used to run Sugar.
 
+Ubuntu 24.04 (Noble Numbat)
+-------------------
+
+Sugar 0.121 can be installed with the following commands:
+
+    sudo apt update
+    sudo apt install sucrose
+
+-   log out,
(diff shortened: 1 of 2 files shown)


quozl:
Opening and then closing pull requests generates multiple notifications to maintainers and other contributors.  Please be mindful of this.  The more noise you make, the more we would expect you to self-manage your workflow.

Ravi-Poddar26:
Apologies for the extra notifications and the previous cleanup attempts. I realized I had included two commits in the earlier PR—one documentation change and the actual updater refresh fix—which was my mistake. I have now created a clean PR containing only the relevant change.

Thank you for your patience and review. I will be more careful with commit structure and PR scope going forward.

chimosky:
In the future, you can drop commits you don't need by doing an interactive rebase.

PR #1077, opened 2026-03-07
Journal: fix Select All button deselecting entries instead of selecting

### LINKED ISSUE #945: Select all button in Journal has no effect.
**Steps to reproduce**
* Search for existing files.
* Select one then use the `Select all` button above items.
* Selections are deselected instead of being selected.

moksha-hub:
Fixes #945

The "Select All" button in the Journal did not work correctly. Instead of selecting all entries, it deselected them. This happened because `select_all()` relied on `ListModel._all_ids` which is populated via the datastore's `find_ids()` D-Bus method. The returned UIDs could have type mismatches (bytes vs strings) compared to UIDs used by the rest of the selection logic, or could be empty when the datastore index is being rebuilt.

## Changes

- Modified `select_all()` in `BaseListView` to iterate the visible TreeModel entries by path and select each entry using the same `COLUMN_UID` source that individual checkbox toggle uses.
- This ensures consistent UID types and correctly selects only entries in the current filtered/searched view.

## Testing

1. Open Sugar Journal
2. Search for existing files
3. Select one entry using the checkbox
4. Click the "Select All" button
5. All visible entries should now be selected (previously they were deselected)

### DIFF
--- a/src/jarabe/journal/listview.py
+++ b/src/jarabe/journal/listview.py
@@ -696,9 +696,19 @@ class BaseListView(Gtk.Bin):
         return self._model
 
     def select_all(self):
-        self.get_model().select_all()
+        model = self.get_model()
+        model.select_none()
+
+        tree_model = self.tree_view.get_model()
+        if tree_model is not None:
+            for row in range(len(tree_model)):
+                path = Gtk.TreePath(row)
+                uid = tree_model[path][ListModel.COLUMN_UID]
+                if uid is not None:
+                    model.set_selected(uid, True)
+
         self.tree_view.queue_draw()
-        self.emit('selection-changed', len(self._model.get_selected_items()))
+        self.emit('selection-changed', len(model.get_selected_items()))
 
     def select_none(self):
         self.get_model().select_none()


moksha-hub:
@omsuneri please have a look.

chimosky:
Tested, works as expected. Thank you!

While I was looking into this, I was more concerned about the datastore not returning any UIDs for selection, if the returned UIDs were a type mismatch then that's an easy fix.

This works and we can merge this, although if there's an underlying problem with the datastore then that's another thing else entirely.

@quozl thoughts?

quozl:
Reviewed.  However, no idea, perhaps there's an underlying problem that needs further testing, and we are only using a band-aid in this.

chimosky:
I agree, I'll hold off on merging and see what I find.

quozl:
Perhaps write some datastore tests, and use them to find out when and why the returned data is incorrect.

chimosky:
> Perhaps write some datastore tests, and use them to find out when and why the returned data is incorrect.

Will do, for some reason, logging to the datastore logger doesn't work as expected, there's no output in the logs.

PR #1081, opened 2026-03-13
Fix AttributeError: replace get_pixbuf with render_pixbuf for Rsvg 2.61+

BALLOON005:
Hi! I'm Saurav, a student interested in GSoC 2026.
I noticed issue #1058 where Sugar shows a black screen on systems with Rsvg 2.61+. Looking at the logs, the error was AttributeError: 'Handle' object has no attribute 'get_pixbuf' which happens because Rsvg 2.61+ removed the get_pixbuf() method.
I fixed this by replacing get_pixbuf() with render_pixbuf() in activitiestray.py as suggested by @quozl in the issue thread.
Please let me know if there's anything I need to change!

### DIFF
--- a/src/jarabe/frame/activitiestray.py
+++ b/src/jarabe/frame/activitiestray.py
@@ -126,8 +126,8 @@ class ActivityButton(RadioToolButton):
     def __journal_drag_data_received_cb(self, widget, context, x, y, selection,
                                         targetType, time):
         data = None
-        if selection.get_pixbuf() is not None:
-            data = selection.get_pixbuf()
+        if selection.render_pixbuf() is not None:
+            data = selection.render_pixbuf()
         if len(selection.get_uris()) != 0:
             data = selection.get_uris()[0].encode()
         if selection.get_text() is not None:


chimosky:
Have you tested this?

Testing might involve building a fedora live image yourself and applying this patch to the [spec file](https://src.fedoraproject.org/rpms/sugar/blob/rawhide/f/sugar.spec), testing this might be a bit demanding and would require you to build the image using [kiwi](https://osinside.github.io/kiwi/quickstart.html), the [soas kiwi definition](https://pagure.io/fedora-kiwi-descriptions/blob/rawhide/f/components/desktops/soas.xml) might come in handy.

I haven't used kiwi yet, so you'll have to figure out how to build the image.

@Meolsei would you be willing to test this?

BALLOON005:
Hi @chimosky ,I haven't been able to test this locally as I don't have a system with Rsvg 2.61+ available. The fix is based on the analysis in issue #1058 where @quozl  suggested replacing get_pixbuf() with render_pixbuf(). I'm hoping @Meolsei who reported the issue can verify the fix. Please let me know if there's anything else I can do!"

chimosky:
You can test this yourself, you don't need to have a system with RSVG >=2.61, the live iso for soas can be installed in a vm.

It's great when someone proposing a fix has been able to test their fix, it's easier to fix any issues that may arise rather than wait for a reviewer.

quozl:
And you can upgrade or downgrade Rsvg as required to develop, you don't have to confine yourself to systems where specific releases of Rsvg are in use.

BALLOON005:
@quozl  @chimosky  i will let u guys knows once i have tested the fix

BALLOON005:
Okay @chimosky  @quozl I tested this on Fedora SOAS 43 in a VM and confirmed the bug. I found the issue is deeper than just activitiestray.py — it also affects sugar3/graphics/icon.py and sugar3/graphics/iconentry.py. The Rsvg 2.61+ API requires render_cairo instead of render_pixbuf. I'm trying to work for a complete fix. :)

BALLOON005:
Hi @chimosky I spent a lot of time testing this on a Fedora SOAS 43 VM to properly verify the fix. After going through multiple attempts I discovered the bug is actually deeper than just activitiestray.py.
The issue also exists in sugar-toolkit-gtk3 specifically in sugar3/graphics/icon.py and sugar3/graphics/iconentry.py where handle.get_pixbuf() is also being called. I tried patching these files in the VM but the fix for icon.py is more complex since it involves restructuring the drawing logic with Cairo contexts.
I've forked sugar-toolkit-gtk3 as well and have been looking into the right approach. Would really appreciate some guidance on how icon.py should be fixed for Rsvg 2.61+ compatibility. Happy to work on a follow-up PR once I understand the right approach.

quozl:
Okay, but I'm surprised I've not seen you show your understanding of the actual changes in the Rsvg git and API.  How long can we continue to use the previous API?  Where are the release notes for Rsvg release that tell us what to change?  Is it Fedora 43 dropping a previous API that is our trigger for action?

BALLOON005:
@quozl  I did some research on this. Looking at the Rsvg changelog, get_pixbuf() was deprecated back in 2.51.0 when they moved towards Cairo-based rendering. In 2.61.0, the pixbuf loader was disabled by default since GNOME started using glycin loaders instead. Since Fedora 43 ships with Rsvg 2.61, that's what's triggering this for us now.
So the previous API has actually been on borrowed time for a while — Fedora 43 is just the point where it finally breaks. The fix would need to replace get_pixbuf() with a Cairo rendering approach using render_document() in sugar-toolkit-gtk3.
I'll try my best putting together a proper fix for icon.py. Would it be okay if I submitted a PR to sugar-toolkit-gtk3 for this?"

quozl:
Sure, just don't hide it in another pull request.  It would be good to extend your change to sugar-toolkit-gtk3 and all the activities.  I see quite a few matches in a search.

BALLOON005:
Sure @quozl will be doing that

BALLOON005:
Hi @quozl  As suggested, I've extended the fix to sugar-toolkit-gtk3 replacing handle.get_pixbuf() with render_pixbuf() in icon.py and iconentry.py. 

PR here: https://github.com/sugarlabs/sugar-toolkit-gtk3/pull/518.
 Would love your feedback!

PR #1083, opened 2026-03-16
gtk4: remove  show_all() calls from detailview.py

BALLOON005:
Part of the ongoing GTK4 migration effort.
show_all() is deprecated and removed in GTK4. Widgets are visible by default in GTK4 so these calls are unnecessary.Removed show_all() calls from src/jarabe/journal/detailview.py as part of preparing the Sugar Shell for GTK4 migration.

Related to the GTK4 transition work tracked in #1019.


chimosky:
There's a `gtk4-port` branch, any gtk4 changes you make should go there instead of master.

PR #1089, opened 2026-03-22
Replace GObject.GObject() with super().__init__()

Dakshsharma1108:
# Replace `GObject.GObject()` with `super().__init__()`

## Summary
Modernize GObject initialization in 30 instances across 27 files by replacing `GObject.GObject.__init__(self)` with `super().__init__()`.

## Changes
- Replaces old-style GObject initialization with Python 3 standard `super()` pattern
- Covers both `super().__init__()` and `super().__init__(**kwargs)` patterns
- No functional changes; fully backward compatible


Dakshsharma1108:
Please review @quozl  @chimosky :)

chimosky:
Having tests for this is overkill, we know what to expect with the change.

In the future, view your changes using `git diff` so you can avoid issues like no new line at EOF in the test file.

Dakshsharma1108:
Got it, I’ll remove the test file and keep this in mind for future changes.

Dakshsharma1108:
you can review it now :)

Issue #840, opened 2019-06-12
Neighbourhood view not updated on shared activity stop


quozl:
When an invited user stops a shared activity, the neighbourhood view may be incorrectly updated.

Reproducer;
* on instance one, start and share the CollabWrapper test activity,
* on instance two, join the shared activity from the neighbourhood view, then stop it

Observed behaviour;
* on instance one, two buddy icons are shown on neighbourhood view for instance two,
* on instance two, the buddy icon for instance one may disappear and reappear,

Expected behaviour;

* on instance one, one buddy icon should be shown for instance two,
* on instance two, no change to neighbourhood view,

Instance one shell.log;
```log
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/jarabe/desktop/meshbox.py", line 409, in _buddy_removed_cb
    self._remove_buddy(buddy_model)
  File "/usr/local/lib/python2.7/dist-packages/jarabe/desktop/meshbox.py", line 435, in _remove_buddy
    icon = self._buddies[buddy_model.props.key]
KeyError: dbus.ByteArray('AAAAB...Y85WlA==', variant_level=1)
```

Instance two shell.log;
```log
1560310543.375053 ERROR root: set_active() failed: org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
```

_org.freedesktop.DBus.Error.NoReply_ is caused by premature termination of the _telepathy-salut_ process before it replies to a message.

On instance two at the time of the event, the process ID of the _telepathy-salut_ process changes, and on Ubuntu 19.04 and an apport crash file is added to `/var/crash` on behalf of the process.

The problem is always reproducible and repeatable.

Theory; a message sent to Telepathy Salut causes it to crash.  It should not crash.  This is a bug in Telepathy Salut, possibly combined with a bug in Sugar usage of the API.

aperezbios:
Was this against a released version of Sugar, top-of-trunk, or a branch?

quozl:
Yes.

aperezbios:
All of the above?

quozl:
Sure, that's another way of looking at it.  At least released versions 0.112, 0.113, 0.114, _master_, and _python3_ branches are affected, as are the branches in development for TelepathyGLib porting.

Aniket21mathur:
My question might be off-topic.
[...]

PR #1090, opened 2026-03-23
Fix KeyError in _remove_buddy by safely handling missing buddy keys

AYUSH676-JARVIS:
Fix KeyError in _remove_buddy in meshbox.py.

The crash occurs when buddy_model.props.key is not present in the _buddies dictionary. This PR adds a safe check and logging to handle the situation gracefully and prevent crashes during inconsistent Telepathy states.

### DIFF
--- a/src/jarabe/desktop/meshbox.py
+++ b/src/jarabe/desktop/meshbox.py
@@ -429,12 +429,18 @@ class MeshBox(ViewContainer):
             icon.set_filter(self._query)
 
         self._buddies[buddy_model.props.key] = icon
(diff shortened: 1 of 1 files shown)


AYUSH676-JARVIS:
Fixes #840
I investigated the issue and found that the crash is caused by a KeyError in _remove_buddy when the buddy key is missing. I have submitted a fix to handle this safely.

AYUSH676-JARVIS:
Hi, I have submitted a PR fixing the KeyError and improving buddy state consistency. Would appreciate feedback. Thanks!

PR #1093, opened 2026-03-24
gtk4: migrate GTK3 container and layout APIs to GTK4 equivalents

Dev10-sys:
- Replace Gtk.VBox and Gtk.HBox with Gtk.Box
- Replace Gtk.Alignment with halign/valign and margins
- Replace pack_start/pack_end with append/prepend
- Replace set_border_width with margin properties
- Replace Gtk.EventBox with Gtk.Box
- Replace Gtk.Toolbar and Gtk.ToolItem with Gtk.Box
- Replace Gdk.Screen with Gdk.Display monitor geometry
- Remove deprecated show_all() calls

No logic or functionality changed. Only deprecated GTK3 container APIs migrated.

### DIFF
--- a/extensions/cpsection/aboutme/view.py
+++ b/extensions/cpsection/aboutme/view.py
@@ -274,12 +274,11 @@ class AboutMe(SectionView):
             self._color_alert.props.msg = self.restart_msg
             self._color_alert.show()
 
-        center_in_panel = Gtk.Alignment.new(0.5, 0, 0, 0)
-        center_in_panel.add(grid)
+        grid.set_halign(Gtk.Align.CENTER)
+        grid.set_valign(Gtk.Align.START)
         grid.show()
 
-        self.pack_start(center_in_panel, False, False, 0)
-        center_in_panel.show()
+        self.append(grid)
 
     def _setup_gender(self):
         self._saved_gender = load_gender()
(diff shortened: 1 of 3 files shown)


extensions/cpsection/aboutme/view.py
```diff
@@ -213,18 +213,16 @@ def _setup_nick(self):
...
 
-        center_in_panel = Gtk.Alignment.new(0.5, 0, 0, 0)
-        center_in_panel.add(grid)
+        grid.set_halign(Gtk.Align.CENTER)
+        grid.set_valign(Gtk.Align.START)
         grid.show()
-
-        center_alert = Gtk.Alignment.new(0.5, 0, 0, 0)
-        center_alert.add(alert_grid)
+
```
chimosky, line 219: No need for this space.

src/jarabe/controlpanel/toolbar.py
```diff
@@ -47,8 +47,8 @@ def __init__(self):
 
         self._add_separator()
 
-        tool_item = Gtk.ToolItem()
-        self.insert(tool_item, -1)
+        tool_item = Gtk.Box()
```
chimosky: Would be great if you used the right kind of box here, you can look at a running sugar instance to figure out which is the right one.

Dev10-sys:
Thanks for the review.

I have updated the alignment usage in aboutme/view.py to avoid the extra container and applied alignment directly to the child.

I have also updated the toolbar container replacement to use the appropriate Gtk.Box and append() instead of insert().

Please let me know if this looks correct.

PR #1099, opened 2026-03-25
fix(journal): replace bare except with OSError in delete logic

### LINKED ISSUE #1097: Replace bare except with OSError in journal delete logic
While reviewing `src/jarabe/journal/model.py`, I noticed a bare `except:` block in the `delete()` function:

```python
        try:
            os.rmdir(metadata_path)
        except:
            # if can't remove is because there are other metadata
            pass
```
Bare except: clauses catch all exceptions, including KeyboardInterrupt and SystemExit, which is discouraged.

Since os.rmdir() raises OSError for expected filesystem cases (e.g., directory not empty), this can be safely scoped to except OSError:.

I can submit a PR for this if it aligns with the intended behavior.

Micha062301:
### PR Category
- [x] Bug fix
- [ ] Feature
- [ ] Performance
- [ ] Tests
- [ ] Documentation

### Summary
Replaces a bare `except:` block in `model.py` with a scoped `except OSError:`.

### Why
Bare `except:` clauses catch all exceptions, including system-level interrupts. Since `os.rmdir()` raises `OSError` for expected filesystem cases (e.g., directory not empty), this can be scoped more safely.

### What changed
- Updated exception handling in `delete()` to catch `OSError` instead of all exceptions

### Notes
- No change in intended behavior
- Improves error handling clarity

Fixes #1097

### DIFF
--- a/src/jarabe/journal/model.py
+++ b/src/jarabe/journal/model.py
@@ -691,7 +691,7 @@ def delete(object_id):
                                   'for file=%s', old_file, filename)
         try:
             os.rmdir(metadata_path)
-        except:
+        except OSError:
             # if can't remove is because there are other metadata
             pass
         deleted.send(None, object_id=object_id)


quozl:
This is obviously a change in intended behaviour; previously any exception is ignored, now only OSError is ignored.  How have you _verified_ that the `pass` is not executed in normal operation of Sugar?

Micha062301:
Thanks for the detailed feedback — that makes sense.

You're right that my reasoning was too generic and didn’t take into account how Sugar runs without a controlling terminal.

I also understand the concern about changing behavior — I haven’t verified whether non-OSError exceptions are intentionally being suppressed here.

I’ll take a step back and re-evaluate this part of the code before proposing any change.

Also noted on avoiding opening an issue alongside a small PR — I’ll follow that going forward.

PR #1101, opened 2026-03-31
Improve README clarity and add usage example

Vinubr:
This PR improves the documentation by adding clearer usage instructions and improving the README's readability.

These changes help new users understand how to use the project more easily.


quozl:
Technically incorrect.  It is not how Sugar is designed to be run.

PR #1103, opened 2026-04-06
docs(journal): add docstrings to bundle evaluation helpers in misc.py

Micha062301:
### Summary
Adds missing docstrings to the bundle evaluation helper functions in `misc.py`.

### Why
Functions like `is_activity_bundle`, `is_content_bundle`, `is_journal_bundle`, and `is_bundle` are frequently used throughout the Journal logic to identify bundle types, but they lacked docstrings. Adding them improves readability for developers navigating the metadata handling logic.

### What changed
- Added descriptive, one-line docstrings to 4 bundle helper functions.

### Impact
- Zero logic changes.
- Improves codebase documentation and maintainability.

### DIFF
--- a/src/jarabe/journal/misc.py
+++ b/src/jarabe/journal/misc.py
@@ -332,19 +332,23 @@ def _downgrade_alert_response_cb(alert, response_id, metadata):
 
 
 def is_activity_bundle(metadata):
+    """Return True if metadata represents an activity bundle."""
     mime_type = metadata.get('mime_type', '')
     return mime_type == ActivityBundle.MIME_TYPE
 
 
 def is_content_bundle(metadata):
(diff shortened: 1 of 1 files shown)


chimosky:
The return statement in these functions are clear that they return a bool, I don't see how the docstring improves that.

For future reference, your commit message can be better, some part of your opening comment can be used.
See [making commits](https://github.com/sugarlabs/docs/blob/master/src/contributing.md#making-commits).

Micha062301:
Thanks for the feedback, that’s a fair point.

I see that the docstrings were just repeating what’s already clear from the code. I’ll focus on adding documentation only where it provides real context or clarity.

Also noted on improving commit messages - I’ll make sure they better reflect the intent of the change.

Appreciate the guidance.

PR #1104, opened 2026-04-08
journal: add docstrings to model.py public interfaces

Micha062301:
Context

model.py provides access to journal data from both the datastore
and filesystem-backed entries. Some public functions have minimal
docstrings that do not explain this behavior.

This can make it difficult for contributors to understand how
operations differ depending on the backend.

What this PR does

Adds docstrings to key functions (find, get, write, delete, copy)
to clarify datastore vs filesystem behavior.

No functional changes.

### DIFF
--- a/src/jarabe/journal/model.py
+++ b/src/jarabe/journal/model.py
@@ -12,7 +12,10 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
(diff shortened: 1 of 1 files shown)


quozl:
I don't think the changes added much.  While these methods form the public API for the module, the module is internal to Sugar, activities (the largest code base) should not use them, and most don't.

chimosky:
Agreed.

Micha062301:
Thanks for the clarification, that makes sense. I’ll focus on contributions that add more practical value going forward.

PR #1108, opened 2026-07-20
optimize: neighborhood buddy and activity lookups to O(1)

sainimal1ba-hue:
This PR replaces the slow O(N) linear scans for buddies and activities with O(1) dictionary lookups to improve overall scaling and performance.

To maintain strict backward compatibility (specifically for modules like friends.py that might query the owner's key), the local owner is now explicitly seeded into the lookup index during initialization. This matches the exact behavior of the old linear scan.

I also did a codebase-wide check for mutability. Since .handle and .props.key are only set at construction or within the update callbacks tracked here, the new lookup dictionaries are safe from going out of sync.

### DIFF
--- a/src/jarabe/model/neighborhood.py
+++ b/src/jarabe/model/neighborhood.py
@@ -756,10 +756,14 @@ class Neighborhood(GObject.GObject):
     def __init__(self):
         GObject.GObject.__init__(self)
 
-        self._buddies = {None: get_owner_instance()}
-        self._buddies_by_key={}
-        self._buddies_by_handle={}
+        owner = get_owner_instance()
+        self._buddies = {None: owner}
+        self._buddies_by_key = {}
+        if owner.props.key is not None:
+            self._buddies_by_key[owner.props.key] = owner
+        self._buddies_by_handle = {}
         self._activities = {}
+        self._activities_by_room = {}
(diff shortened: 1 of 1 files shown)


sainimal1ba-hue:
sorry i think my auto correct changed pep8 to flask8 🙏

chimosky:
I see the idea behind this, but this isn't a problem we have so I don't see why this needs to change.

sainimal1ba-hue:
That's fair, thanks for looking at it. It's not fixing anything broken today, buddy counts are small enough that the linear scan never showed up as a real problem. I put it up mainly because it was a quick change with no behavior difference for normal use. Happy to close this if you'd rather keep the code as-is, no worries either way.

chimosky:
I'd love to keep this as is, a "quick change with no behaviour difference for normal use" isn't a great reason for this change, considering it's just more lines of code. Thank you!

quozl:
I agree with @chimosky.

sainimal1ba-hue:
I thought this will be useful in the future so I wanted to implement it now rather than wait for later. But sure I do agree that it's more line of codes and will be pretty hectic for new devs to understand the codebase. 

Thank you for taking your time once again:- @chimosky @quozl

PR #1109, opened 2026-07-21
fix: Harden proxy ping to avoid shell injection

sainimal1ba-hue:
This replaces os.system() with subprocess.run() in the proxy configuration verification flow.

Previously, the proxy hostname was concatenated into a shell command before invoking ping. Using subprocess.run() with an argument list avoids invoking a shell, eliminating the possibility of shell interpretation of the hostname while preserving the existing behavior for valid hostnames.

I also pass -- before the hostname so that values beginning with - cannot be interpreted as command-line options by ping.

Took inspiration from https://github.com/sugarlabs/sugar-ai/pull/136

### DIFF
--- a/extensions/cpsection/network/view.py
+++ b/extensions/cpsection/network/view.py
@@ -642,7 +642,7 @@ class Network(SectionView):
                     if hostname:
                         non_blank_host_name_counter += 1
                         try:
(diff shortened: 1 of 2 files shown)


sainimal1ba-hue:
wtf wait a sec 3 more PR's joined somehow

PR #1123, opened 2026-08-17
journal: ai server configuration panel on the settings menu

vyagh:
## The online half:
- the journal can ask a configured AI server for Jo's next question, and a new AI section in the control panel holds the three settings:
  1. an enable switch
  2. the server address
  3. the key.

- The switch ships off; with nothing configured.
- The journal's built-in question bank stays the floor: when the server is slow (30 seconds), unreachable, or answers badly, the next question comes from the bank instead, mid-conversation included.

Builds on #1121.

## Notes:
- there is no public server to point this at yet; the address is meant for a school's own machine or a hosted deployment.
- **what goes out:** 
  - title, description, activity id, the conversation turns, and now the entry's context too (preview, moment snapshots, tags and time spent, added in the two newest commits).
  - Answers to Jo's "who was with you" questions stay on the machine, and if one was starred into the description, the description is not sent.
- the request and response use the reflection engine's wire format directly; the server half is sugarlabs/sugar-ai#163.
- the key is stored in a file only the user can read, and the panel only reports Connected once both address and key are set.
- most of the diff is tests: about 1,050 of 2,000 lines.
- the panel itself has no automated tests (GTK); it was checked in a running Sugar shell.

## Testing:
324 journal tests pass with unittest discover; flake8 clean on the new files.

Series notes in #1111.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,6 +40,7 @@ data/icons/Makefile
 data/Makefile
 extensions/cpsection/aboutcomputer/Makefile
 extensions/cpsection/aboutme/Makefile
+extensions/cpsection/ai/Makefile
 extensions/cpsection/background/Makefile
 extensions/cpsection/backup/Makefile
 extensions/cpsection/backup/backends/Makefile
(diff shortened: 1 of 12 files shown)


vyagh:
**Two recent commits since opening:**

The Journal now speaks the engine's wire format directly, and the request packs the entry's context, which is:
preview, moment snapshots, tags, time spent, behind the existing toggle

**Server half**: sugarlabs/sugar-ai#163.

